PR #62, opened 2019-04-26
Gist900

mkumba:
gist 9.0.0 (finally)


uscholdm:
This seems to have worked. I did some comparisons of new and old verisions.

We might want to change the command to serialize to turtle rather than RDF/XML.  We’d have to do some experiments to see whether compare with turtle were easier or harder.  With RDF/XML, the file differences often do not reflect changes in the ontology, but only changes in ordering where one thing is move to somewhere else in the file.

Michael

uscholdm:
I was able to compare the files, so the git commit seems to have worked.  In the future, perhaps we can use Turtle?
---

Why were the restrictions requiring a UnitOfMeasure to have a standard unit and conversion removed?
---

There are no release notes updates for 9.0.0 in the ReleaseNotes.txt file.
The change log in the Visio file is very incomplete.  To get a decent grasp of the changes will require manually and painfully comparing the text files one by one from version to version.


I also added 3 smallish issues to github.

PR #83, opened 2019-08-01
Fix convertToBase value of gist:_minute

rjyounes:
Base unit is gist:_second, conversion factor was 1.0, should be 60.0.


rjyounes:
PR withdrawn, issue to be fixed in a later commit.

rjyounes:
@uscholdm Michael, now I'm confused about why the change needs to be made in e6, since you told me that not everyone is using e6. How are they making edits?

uscholdm:
Not everyone uses e6 for client projects (I might be the only one, now that Dave does very little ontology development).

But the only way to evolve gist for now is using e6.

Issue #20, opened 2018-01-04
Update all labels according to a standard convention


uscholdm:
A number of gist resources do not have labels.  They probably should be there, across the board.

rjyounes:
Need to agree on conventions before this is addressed.

rjyounes:
I will make a proposal for the next meeting, but not committing to the implementation.

rjyounes:
Proposal:

My proposal involves mapping from local names to labels, so I begin with my proposed _orthographic_ (as  opposed to structure/content) conventions for local names:

Classes: camelcase, initial uppercase. Acronyms also camel-cased so that word boundaries are unambiguous - e.g., AmaGuideline not AMAGuideline. Alphanumeric characters only - e.g., Isbn10, not Isbn-10 or ISBN-10.

Properties: camelcase, initial lowercase. 

Both: non-standard abbreviations should not be used: e.g., hasUoM should be hasUnitOfMeaure. 

Labels:
Classes: Initial uppercase, remainder lowercase, normalized to natural language standards - I.e., hyphens inserted, acronyms in all caps, etc. Examples: NSDAP officer, ISBN-10. I prefer only initial word capitalized because sentence case, while technically well-defined, can introduce inconsistencies when actually implemented by different people.

Properties: same as classes, with initial lowercase. Hypothetical examples: has unit of measure, has SSN. Initial lowercase follows the convention applied to local names.

These conventions facilitate automated mapping between local names and labels. E.g., "hasUoM" and label "Has Unit of Measure" cannot be automatically mapped to one another. Labels with acronyms and  hyphens can be automatically mapped to local names but not the reverse, but this is a general problem.

rjyounes:
Michael: Separate end of acronym from beginning of word with underscore.

Boris: Likes entire proposal.

Dan: Uppercase label for property names, though lowercase in local names. Capitalization in column names is better. Reference to MS.

Boris: Should in this case argue for alternative labels of some sort. Could create other annotations like hasOfficialName for special-purpose labels. URI and prefLabel should not be driven by display or other concerns. 

Michael: Different clients can decide to do things differently. 

Consensus: adopt this proposal.

Jamie-SA:
A comment regarding automated mapping of terms to labels, it seems that the purpose of a label is the ability to provide something different than what is contained in the name. If it is required that it be possible to automate it then you wouldn't need labels (at least for the English language).

rjyounes:
[...]

PR #156, opened 2020-01-26
Fix Issue #153; add missing labels

### LINKED ISSUE #153: Add missing labels
Was originally part of #20, now split out as separate issue. For this issue, just add labels, based on local names, to terms with no labels.

Mark has a script to automate this.

marksem:
Fixes #153 

Had to do 1 re-serialize commit first, so that the "real" changes would be easy to see. **See latter 3 commits for the actual changes.**  Note that the fix to labeling property `goal `was to fix the fact that it was a typo in the `Artifact `restriction.


marksem:
Yes @sa-bpelakh .  I guess when we know which commits/PRs make the cut for the next release, and we have a release branch for it, we'll need to edit the OntologyFiles/ReleaseNotes.txt file on that branch with the appropriate notes.  @rjyounes , perhaps that needs a place in some release checklist you may be developing for SA?

OntologyFiles/gistIoT.owl
```diff
@@ -61,6 +61,8 @@
 	</owl:Class>
 	
 	<owl:Class rdf:about="&gist;ControllerType">
+		<rdfs:subClassOf rdf:resource="&gist;Category"/>
+		<rdfs:label rdf:datatype="&xsd;string">Controller Type</rdfs:label>
```
rjyounes: We agreed on sentence case rather than title case for all labels - thus "Controller type" rather than "Controller Type." See discussion on issue #20 - sorry, I forgot to copy over our decisions when I split up the issues.

marksem: Changed it to sentence case.

DanCarey404:
I definitely do not remember agreeing to sentence case for labels.  And a quick search of the notes in #20 show nothing like that.

I, for one, do not like the idea of sentence case labels.  We should resolve this issue quickly.

rjyounes:
@DanCarey404 I am copying your comment to #153, which is where the original discussion took place. 

Let's move the discussion to that issue.

OntologyFiles/gistIoT.owl
```diff
@@ -4,98 +4,108 @@
...
+						<owl:someValuesFrom rdf:resource="&gist;Sensor"/>
+					</owl:Restriction>
+				</owl:intersectionOf>
+			</owl:Class>
+		</owl:equivalentClass>
+	</owl:Class>
+	
+	<owl:Class rdf:about="&gist;ControllerType">
+		<rdfs:subClassOf rdf:resource="&gist;Category"/>
+		<rdfs:label rdf:datatype="&xsd;string">Controller type</rdfs:label>
```
uscholdm: This change is a bit confusing, but on careful inspection, on each commit, its fine.

OntologyFiles/gistIoT.owl
```diff
@@ -4,98 +4,108 @@
...
 	</owl:Class>
 	
 	<owl:Class rdf:about="&gist;Equipment">
 	</owl:Class>
 	
 	<owl:Class rdf:about="&gist;MessageDefinition">
 		<rdfs:subClassOf rdf:resource="&gist;SchemaMetaData"/>
-		<rdfs:label rdf:datatype="&xs;string">Message Definition</rdfs:label>
-		<rdfs:comment rdf:datatype="&xs;string">Each pulse from a Sensor is reflected in a message, as well as each instruction to an Actuator</rdfs:comment>
+		<rdfs:label rdf:datatype="&xsd;string">Message Definition</rdfs:label>
```
uscholdm: gist has always used xs not xsd. Making this change should not be done here. It should  be first put up as an issue to discuss, and if approved, make the change across all files.

rjyounes:
I find this a strange point. The use of a prefix is file-based and has no formal significance. It could be called xyz as long as it points to the right URI. The prefix itself has no consequence and no semantic value. xsd seems to me to be more commonly used.

sa-bpelakh:
It's primarily convention, right? If I am going to edit the file, I don't want to have to scroll back to the prefix declarations for commonly used namespaces. I could name `rdf:` and `owl:` as `foo:` and `bar:`, respectively, but it would result in confusion. We should pick one for `xs` vs `xsd` and be consistent with it. I have always used `xsd` (and `platts-ontology` uses it throughout), but I am OK either way. The actual W3C standard (https://www.w3.org/TR/xmlschema11-1/#xsd-namespace) uses `xs:`.

rjyounes:
I've always used xsd as well. Interesting that the standard uses the prefix xs when the uri itself has "xsd-namespace."

sa-bpelakh:
¯\\\_(ツ)\_/¯

uscholdm:
I also prefer xsd.

rjyounes:
@sa-bpelakh At first I thought that was a pushme-pullyou (are you familiar with Dr Doolittle?), which would have been equally applicable.

rjyounes:
Anyway, I agree with @uscholdm that we should raise this as a separate issue and have created one: #158. @marksem should change the prefixes back.

rjyounes:
@marksem Re release notes: Boris is responsible for these as he is our release facilitator for the January release. We probably want to come up with a template for these at some point.

OntologyFiles/gistIoT.owl
```diff
@@ -4,98 +4,108 @@
...
-		<rdfs:comment rdf:datatype="&xs;string">The things that a sensor can sense, such as light, heat, current, moisture et</rdfs:comment>
+		<rdfs:label rdf:datatype="&xsd;string">Phenomena Type</rdfs:label>
+		<rdfs:comment rdf:datatype="&xsd;string">The things that a sensor can sense, such as light, heat, current, moisture et</rdfs:comment>
 	</owl:Class>
 	
 	<owl:Class rdf:about="&gist;PhysicalActionType">
 		<rdfs:subClassOf rdf:resource="&gist;Category"/>
-		<rdfs:label rdf:datatype="&xs;string">Phenomena Type</rdfs:label>
-		<rdfs:comment rdf:datatype="&xs;string">The effects to be realized in the real world, such as lifting a garage door, turning off a valve, dropping cadmium rods etc</rdfs:comment>
+		<rdfs:label rdf:datatype="&xsd;string">Phenomena Type</rdfs:label>
```
rjyounes: This looks like the wrong label.

rjyounes: Never mind, this was just the serialization commit.

rjyounes:
I'd like to propose the following so we can merge this PR:

1. Use Mark's current labels and resolve this matter in the discussion of #20.
2. Restore xs: prefix and defer to issue #158.

If rdf-toolkit changed the prefix, it looks like it will be a problem to stick with xs.

uscholdm:
I'm happy with @rjyounes 's suggestion.
I doint think the serializer changes the prefix. We use .xsd for PLEO.

rjyounes:
The reason I suspected the serializer was that the prefix changes occurred in the commit that was only reformatting with the serializer. PLEO files are ttl; I wonder if the serializer behavior is different between the file types. It seems unlikely but I don't know how else to explain it.

OntologyFiles/gistIoT.owl
marksem: The serializer changed it.  Or protege.  I fought with it for a while, but then decided to do the format change as 1 commit, and the real fixes next, as otherwise I was at an impasse.  The reformat should be a 1 time thing if we use the same serializer going forward.

marksem:
> @marksem should change the prefixes back.

For pragmatic reasons, I disagree.  It is now in the form that serializer wants.  And that form uses xsd, our preferred ns.  To put it back means more maintenance next time.   We need to be avoiding changes that add maintenance time.

rjyounes:
Technically we have enough approval to merge, but I don't want to act by fiat. We should be able to merge this and continue the discussion of label formats in issue #20.

OntologyFiles/gistTop.owl
```diff
@@ -28,67 +29,79 @@
...
-			</rdf:Description>
-			<owl:Restriction>
-				<owl:onProperty rdf:resource="&gist;hasUoM"/>
-				<owl:someValuesFrom rdf:resource="&gist;AreaUnit"/>
-			</owl:Restriction>
-		</owl:intersectionOf>
+		<rdfs:label rdf:datatype="&xsd;string">Area</rdfs:label>
+		<rdfs:comment rdf:datatype="&xsd;string">A measurement of two-dimensional space.</rdfs:comment>
+		<owl:equivalentClass>
+			<owl:Class>
```
sa-bpelakh: So the transition I am seeing in several places here from 
```
gist:SomeClass owl:intersectionOf ( FirstClass, SecondClass ) .
```
to
```
gist:SomeClass owl:equivalentClass [
  a owl:Class; 
  owl:intersectionOf ( FirstClass, SecondClass ) .
]
```
Is it because the original declaration was never compliant and we are fixing it? Or are both forms valid and we are just choosing a preferred form?

rjyounes:
Fixed merge conflicts. Will address formatting of label in #20.

OntologyFiles/gistTop.owl
Jamie-SA: IIRC there are a bunch of these owl:intersectionOf constructs in gist that the old E6 tool wrote out that way but that Protege will rewrite into owl:equivalentClass (as above).
We (devs) tried to decide if the OWL spec requires it or not. I'd be interested in a more official answer from anyone that wants to go interpret the OWL spec.

rjyounes:
@Jamie-SA @sa-bpelakh I've had Protégé play this trick on me as well. It seems like the simpler version should be valid OWL and that the other is more verbose than necessary. Another reason not to like Protégé.

uscholdm:
If possible, just do what the serializer does. What Protégé or E6 do should not matter.

Issue #123, opened 2019-11-13
Remove gistWiki 


Jamie-SA:
In the updates to gist 9.0.0, gistWiki didn't get updated to use SSL. Switch to https on at least these lines (there might be others...):

Line  3: <!ENTITY gist "http://ontologies.semanticarts.com/gist/">
Line 10: xmlns:gist="http://ontologies.semanticarts.com/gist/"
Line 16: <owl:Ontology rdf:about="http://ontologies.semanticarts.com/o/gistWiki">
Line 19: <owl:imports rdf:resource="http://ontologies.semanticarts.com/o/gistTop9.0.0"/>
Line 20: <owl:versionIRI rdf:resource="http://ontologies.semanticarts.com/o/gistWiki9.0.0"/>

uscholdm:
I thought we were getting rid of gistWiki entirely?

M.

Jamie-SA:
Dave confirmed that gistWiki should be removed.

rjyounes:
Closing issue.

rjyounes:
Added #177 to remove gistWiki.

rjyounes:
Don't need #177 since this is already includes the possibility of removing gistWiki. Does this constitute a major change? Can we do it for February?

rjyounes:
Group decision: gistWiki is only there due to error. Removing it can be a patch.

PR #173, opened 2020-02-13
Correct namespace to https for gistWiki

sa-bpelakh:
Found this while gathering ontology metrics.


Jamie-SA:
See #123, gistWiki is supposed to be removed.
And there was already a issue for this problem, don't forget to search existing issues before creating a new one.

sa-bpelakh:
@Jamie-SA Ah, missed that entirely, will remove it entirely.

PR #192, opened 2020-02-28
Release 9.1.0

sa-bpelakh:
I removed gistWiki, as discussed, and reformatted ReleaseNotes as Markdown, adding a 9.1.0 section.


sa-bpelakh:
This will also close https://github.com/semanticarts/gist/issues/123

uscholdm:
I just took a good look at all the changes from 9.0.0 and noticed a large number of changes with disjoints. I do not recall any discussions about that.  The affected file is gistTop. They call came in with commits by Rebecca on Jan 29 relating to Issue 153. But that issues does not mention disjoints.  I searched through all the issues around then and none mention disjoints.   

A lot of the changes are dubious. Here are a few:
- A geopoint can now be an intention
- a magnitude can now be a language
- an organization can also now be a language
- a PhysicalIdentifiableItem can now be an Organization

Michael

uscholdm:
I looked into this a bit further. The file compare seems to say that a geopoint used to be but no longer is disjoint with an intention.  Similar for the other examples. However, when I load the ontology, the disjoints seem correct.   So the question becomes: 
- when I look at the file comparison, why does it seem like disjoints have moved all over the place?

It rather defeats the purpose of looking at file differences.

sa-bpelakh:
@uscholdm Michael, I think one of the changes that was made during this release is a more consistent reformat of the files. It should improve things going forward, but may have resulted in some weird diffs for this release. As long as it is sound when loaded, we should be good for now.

rjyounes:
@sa-bpelakh I haven't looked at the files, but why would reformatting cause changes to disjointness axioms? And does this mean that if we use ontology-toolkit instead of the old serializer in our projects we will initially get a large amount of reformatting?

sa-bpelakh:
@rjyounes I don't think the serializer changed the axioms, it just made the diffs less obvious. We can examine the specific axioms Michael is referencing to validate that they have not changed. 

The commit hooks on the repo should use the serializer - if you see non-clean diffs in a pull request, that is usually the cause. The `bundle` functionality in ontology-toolkit that I use to package the release re-runs the standard serializer on each OWL file, just to be sure.

rjyounes:
@uscholdm @sa-bpelakh OK, who is going to examine the files?

uscholdm:
> The commit hooks on the repo should use the serializer
We need to ensure that everyone has it set up, and is using the exact same .jar file.

uscholdm:
Another thing that would be nice to have in our ontology toolkit is a script that mushes all the gist files into one ontology. Then we could use Protégé to do an ontology compare.  Dean Allemang did this for his work. Neat trick.

uscholdm:
> I haven't looked at the files, but why would reformatting cause changes to disjointness axioms?

I don't get it either.  I also saw one of the commits that did a fresh serialize, and I did my compare from there to 9.1 hoping the disjoint differences would go away. Did not work, they were still there.

rjyounes:
@uscholdm @Jamie-SA @sa-bpelakh I am going to contact Jamie and ask him not to post the new files until we sort this out.

rjyounes:
@sa-bpelakh As the release manager, it would be up to  you to figure this problem out. Do you have the bandwidth? I know you are traveling. We can postpone the release until you are back home.

Jamie-SA:
I had done most of the release. I will undo the easy parts.

PR #234, opened 2020-04-23
Improve the definition of SimpleUnitOfMeasure (stylistic change only). Fixes #84.

### LINKED ISSUE #84: Fix definition of gist:SimpleUnitOfMeasure
The definition currently reads:

```
Each simple unit has a base unit and a conversion factor to the base. The bases are from the System International (SI). This is the number by which one multiplies a Unit by to get to base, or divides by to get from base.  So the convertToBase for inch is 0.0254 to get you to the base (meter).
```
1. There is a minor typo in sentence 3: an extra "by."

2. If "this is the number" refers to the convertToBase value, that should be explicit. Currently it's a bit confusing.

Proposed rewrite:

```
Each simple unit has a base unit and a conversion factor to the base. The bases are from the System International (SI). The conversion factor is the number that one multiplies by to get to base, or divides by to get from base.  So the convertToBase for inch is 0.0254 to get to the base (meter).
```

rjyounes:
@uscholdm I have not added to the release notes for this item. It seems to me we don't want to put such a trivial item in the release notes -

### DIFF
--- a/OntologyFiles/gistTop.owl
+++ b/OntologyFiles/gistTop.owl
@@ -575,7 +575,7 @@
 	
 	<owl:Class rdf:about="&gist;SimpleUnitOfMeasure">
 		<rdfs:label rdf:datatype="&xsd;string">Simple Unit Of Measure</rdfs:label>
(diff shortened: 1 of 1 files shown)


sa-bpelakh:
@uscholdm Sorry, Michael, I reviewed and merged, since it was trivial, just trying to move the process along 😸

Issue #243, opened 2020-04-30
gist:memberOf incorrect rdfs:domain


marksem:
Currently (gist 9.1.0) has

gist:memberOf  rdfs:domain   gist:Collection.

This means that if X memberOf Y, then X is a collection.
But clearly this is incorrect.  Could say that if X memberOf Y, then Y a collection, but not X.
E.g.,  _PersonX memberOf _OrgY does not make _PersonX a collection. 

Note that gist:hasMember (the inverse of gist:memberOf) already has   rdfs:domain  of  gist:Collection.

Suggest just remove the rdfs:domain restriction from gist:memberOf.
Alternatively change it to an rdfs:range restriction.

uscholdm:
Cleary an unintended bug. I would go further and recommend to get rid of the domain/range constraint.  If you want someone to be a member of an organization, then the organization is forced to be a collection. Bad idea, forces introducing extra properties that only differ in domain or range.

rjyounes:
@uscholdm I agree.

PR #247, opened 2020-05-01
Remove redundant disjoints on UnitOfMeasure

marksem:
As discussed on Teams regarding my work on issue #243 there are redundant disjoints that could be removed.

Suggestion:  Remove redundant disjoints in gistTop, i.e. state the distjoint on just one side.  Ususally the one that occurs 1st alphabetically.  

Here's the teams discussion transcript:

> [9:18 AM] Mark Wallace
>     sa-gist, I have an issue where when I run the serializer, it removes a bunch of disjointness axioms, even though I didn't touch anything there.  Is anyone else seeing this?  
> ​[9:43 AM] Dalia Dahleh
>     I believe that's protege, not the serializer. When you save an ontology in protege it only writes one side of a symmetric property.
> ​[9:48 AM] Mark Wallace
>     Ah, ok  Wonder how many of us are using Protege?  More importantly, is it OK to "agree with " protege and let these come out (under a different atomic commit), or to keep having to "unstage that hunk" after I serialize from Protege? 
> ​[10:19 AM] Dalia Dahleh
>     Good question. **I'm thinking it might be better to remove them.** Not only to make it easier to save with protege, but also to make it more consistent. Most disjoints are only stated on one class in our gist files (even UnitOfMeasure has a few more that are only stated on the other class)
> ​[10:21 AM] Mark Wallace
>     ok, sounds goodl.  I'll take them out as another atomic commit
> Edited​[10:46 AM] Boris Pelakh
>     Is Protege consistent about which half of the symmetrical disjoint it drops? Should there be a standard? 
> ​[10:48 AM] Dalia Dahleh
>     It looks like it keeps the first, sorted alphabetically
> (2 liked)

### DIFF
--- a/OntologyFiles/gistTop.owl
+++ b/OntologyFiles/gistTop.owl
@@ -663,15 +663,6 @@
 	<owl:Class rdf:about="&gist;UnitOfMeasure">
 		<rdfs:label rdf:datatype="&xsd;string">Unit of Measure</rdfs:label>
 		<rdfs:comment rdf:datatype="&xsd;string">Standard unit by which we measure things</rdfs:comment>
-		<owl:disjointWith rdf:resource="&gist;Content"/>
-		<owl:disjointWith rdf:resource="&gist;GeoPoint"/>
-		<owl:disjointWith rdf:resource="&gist;GeoRegion"/>
-		<owl:disjointWith rdf:resource="&gist;IntellectualProperty"/>
-		<owl:disjointWith rdf:resource="&gist;Language"/>
-		<owl:disjointWith rdf:resource="&gist;Organization"/>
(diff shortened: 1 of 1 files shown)


rjyounes:
You missed my contribution to the discussion:

Protégé does the same with inverses, btw (which is a symmetric property). The problem is when you read the ontology without Protégé, e.g., in a text file, you cannot see the relationships on half the terms involved.

rjyounes:
@marksem OK to close this PR without merging and create an issue? I think the issue should be a general one, as I mentioned in my comment above, though it can certainly refer to this example.

marksem:
Sure.

rjyounes:
Creating an issue for triage before making the revision.

PR #251, opened 2020-05-03
Remove weird characters in first line of release notes, and delinted markdown. Fixes #248.

### LINKED ISSUE #248: Remove weird chars from ReleaseNotes.md
ReleaseNotes has 3 non-printable leading chars, can be seen by doing `od -a -x` on the 1st line.

rjyounes:


rjyounes:
@sa-bpelakh Yes, I did a full de-linting of the file and that was one of the changes. 

@Jamie-SA I'm following Markdown rules in adding the angled brackets. First, this turns it into a link - see https://www.markdownguide.org/basic-syntax/. Second, there are many other contexts in which URLs are enclosed in angle brackets for this reason. Third, according to the markdown lint rules I followed at https://github.com/DavidAnson/markdownlint, URLs should be enclosed in angle brackets. David Anson has created a markdown lint tool that operates according to these rules, so I decided to follow them. It is possible to configure the plugin to ignore some of the rules, but I figured it would be easiest to follow all of them so that it's consistent across editors.

Jamie-SA:
@rjyounes sorry, I realized after that maybe it was a Markdown feature I was not familiar with and should have checked before opening my mouth :)

rjyounes:
@Jamie-SA No problem. Just think of how many times I open my mouth without checking ;)

PR #264, opened 2020-05-15
Add definedBy action to bundle

sa-bpelakh:
Broke it into 3 steps, as we discussed. There is no issue that asks for this, should we have something in the release notes?


rjyounes:
Yes, let's add it to the release notes. I think it's a good idea to have an issue to point to, so I'll create one.

PR #304, opened 2020-05-28
orderedMemberOf is Functional, not Inverse Functional

### LINKED ISSUE #262: gist:orderedMemberOf should not be Inverse Functional
In gist 9.2.0, gist:orderedMemberOf is defined as Inverse Functional.  

This means that the inverse property, if it exists, is functional. [1] The inverse property does in fact exist, and it is gist:hasOrderedMember, which should clearly **not** be functional.  If it were functional, each OrderedCollection could only have 1 member. 

Suggestion: 
Define gist:orderedMemberOf as Functional instead of Inverse Functional

Refs.
[1] http://protegeproject.github.io/protege/views/object-property-characteristics/

marksem:
In the current model of each Ordered Member being
in at most 1 ordered list, the orderedMemberOf
property should be functional. Fixes #262

### DIFF
--- a/OntologyFiles/gistMeasure.owl
+++ b/OntologyFiles/gistMeasure.owl
@@ -134,7 +134,7 @@
 	</owl:AnnotationProperty>
 	
 	<rdf:Description rdf:about="&gist;orderedMemberOf">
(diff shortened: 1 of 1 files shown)


rjyounes:
Patch fix because it corrects an error.

Issue #93, opened 2019-08-14
Remove range assertion gist:Address from gist:hasCommunicationAddress


DougalW:
The Range of the gist:hasCommunicationAddress object property is Address. This causes unintended reasoning side effects when constructing an ontology that follows the pattern described in the 'Approaching Addresses in GIST' video (https://www.youtube.com/watch?v=8uFWX-xQU-U&feature=youtu.be) - see around 17:05 in the video.
E.g. if I create a Role class that is a subclass of Collection, and give it the following class expression:
'Communication Address Of' some 'Communication Preference'
and 'Communication Preference is a subclass of Category with the following class expression:
'Has Communication Address' some Address
When I run the reasoner, it classifies Role as a SubClass Of Address.

Would be nice to remove the range:Address for gist:hasCommunicationAddress

rjyounes:
@ungricht This issue is still in review. If it was assigned to you, the task is to consider it and bring back options/proposals to the group.

uscholdm:
@DougalW What is an example of something having a communication address that is not an address? It is a contradiction. Either the range assertion stays, or the meaning of and terms for these concepts needs to be more clear.

The problem might be modeling a Role as a Collection. That does not seem correct to me. What is a role a collection of? Surely not every Role is a collection.

rjyounes:
Closing issue.

PR #308, opened 2020-06-11
Removed range assertion: gist:Address from gist:hasCommunicationAddress

ungricht:
Looks like this got re-serialized?
Signed-off-by: Meika Ungricht <meika.ungricht@semanticarts.com>


uscholdm:
What is an example of something having a communication address that is not an address? It is a contradiction.  Either the range assertion stays, or the meaning of and terms for these concepts needs to be more clear.

rjyounes:
@uscholdm The discussion is on the issue, #93.

Issue #292, opened 2020-05-18
Do away with modularity in gist


rjyounes:
Put everything into one file. See #291 for rationale - there may be other issues raising the same point.

I'm starting to come around to the idea that modularity doesn't work in gist - while still supporting it in theory. I believe there are many examples like this one: gist:categorizedBy and gist:Category are not defined in gistCategory, but rather gistTop, because they are referenced either in gistTop or files that import gistTop. Rather than trying to untangle this web of references, perhaps it is best to dump everything into the same file and forget about modularity. gist:hasSubCategory and gist:hasSuperCategory are defined in gistCategory as you would expect.

rjyounes:
We say this on our website: "Gist is extremely modular. There are several modules that can be used collectively or in subsets if you don't need all the concepts in gist." Needs to be removed if we make this change. In any case it's an overstatement.

uscholdm:
>  In any case it's an overstatement.

Another way to put it is to say it was aspirational – IMHO, it was only ever a little true some of the time, and it got less so over time as more stuff migrated to gistTop. Hence the learning.

rjyounes:
Relates to #260 and #288; all three can be addressed together.

rjyounes:
Adding to June 2020 project; let's discuss whether we can get this done for the June release.

marksem:
AC)

1. No more catalog-v001.xml will be needed.
2. Just gistDeprecated.owl and gistCore.owl needed (top level)
3. Update bundling/packaging to still inject proper version number and create good folder shape.

rjyounes:
Other tickets need to be closed out, Mark will implement this and notify others to continue working on other issues.

Meika will finish #93, the PR will be merged, then all other modifications will be postponed till Mark's work is finished. Whoever merges the PR, please send out notification to the group to stop work.

Issue #223, opened 2020-04-06
Consider using Turtle for editing gist


uscholdm:
Turtle designers met their goal of making a much friendlier syntax than rdf/xml. I often make changes in the raw files, and rdf/xml slows me down quite a lot. Can we conver to using Turtle instead, and then include both in releases?

Related to Issue https://github.com/semanticarts/gist/issues/129

JonathonGist:
As a newcomer to ontologies and the Semantic web and semantic disciplines, with a few decades of database design, development and data tier architecture, I have found Turtle easy, intuitive and wonderfully concise. I far prefer Turtle to rdf/xml. I would find it much easier to work with gist in Turtle natively, than continue with rdf/xml. My two cents.

rjyounes:
Do we need to include both in the release package? Can we just make ttl our official serialization and let users convert to RDF/XML or another serialization they prefer? Even if both are included in the package, the URIs have to point to one of them when a request header is not explicitly included.

JonathonGist:
From my perspective, having only ttl as the official serialization is fine. I like that idea; users who need RDF/XML can convert it. My two cents.

Jamie-SA:
There may be some issues with returning a ttl file when the URL has no file extension as many parsers use the file extension to determine the file type. It wouldn't surprise me if many pieces of software will assume that a Ontology without an extension is RDF/XML.

uscholdm:
I think we should include any/all formats that people are likelyi to want. I disagree with the suggestion that it is easy to go into/out of Protege. It a big fiddle, especially if you want to look at the text files.

JonathonGist:
Protoge isn't an easy tool to learn, from my perspective as a newbie to the semantic technology stack. I love the idea of the easy-to-access ttl being available. But I see there are complicating factors to providing both. I'm not equipped to provide much useful feedback on this topic.

rjyounes:
@uscholdm What comment about Protégé are you referring to?

uscholdm:
@rjyounes Somewhere, (maybe in a separate chat on Teams) someone said not to bother with multiple formats because it is easy enough to import into Protege and export whatever format you like.

rjyounes:
I see. Better to just run rdf-toolkit.jar.

uscholdm:
For someone who wants to have a look at ontology files and is not a developer, having all the formats available is the answer.

Jamie-SA:
[...]

PR #312, opened 2020-06-16
Issue292 Demodularize gist

marksem:


marksem:
@sa-bpelakh , I don't think it invalidates module imports.  I think if someone imports an un-versioned module other than gistCore,, e.g. gistAddress, that they'll get 9.2.0, and its imports closure (all 9.2.0).  Which *is* (and ever will be) the most recent version of gistAddress (and its imported modules).  Of course, this is why I think it is better to import a specific version of an ontology (use versionIRI), so you know exactly what concepts you are committing to.

rjyounes:
@marksem I agree that imports should always use versioned IRIs; it doesn't make sense to import something when you don't know what you're getting. But (to state the obvious) what makes sense is not necessarily what happens in practice.

If someone imports unversioned gistCore and (redundantly) unversioned gistAddress, there is the potential for unresolvable clashes, since they'll get version 9.3.0 (or whatever) of gistCore, and 9.2.0 of gistAddress, and there may be conflicts between term definitions in the two versions. (See https://www.w3.org/2007/OWL/wiki/Imports#Clashing_versions). Note that you could say the same thing if they import, say, gistCore9.2.0 and gistAddress9.1.0, but this is specifically prohibited by the [OWL spec](https://www.w3.org/TR/owl2-syntax/#Imports): "The import closure of O should not contain ontologies O1 and O2 such that O1 and O2 are different ontology versions from the same ontology series."

marksem:
Wonder if this would be a good time to switch to Turtle format. :)

rjyounes:
@marksem Sorry, when you said "maybe now's the time" I thought you meant this release, not this PR. There is a separate issue #223 for this, which hasn't been decided on, and in any case I don't think this change should be included in this PR in order to keep things atomic. What do you think?

rjyounes:
We've determined this will be a minor rather than major change, based on the fact that we encourage users to import versioned rather than unversioned IRIs.

uscholdm:
@rjyounes said: 
> There is a separate issue #223 for [Turtle], which hasn't been decided on, and in any case I don't think this change should be included in this PR in order to keep things atomic. What do you think?

I agree. Good practice.

marksem:
Fixes #292

Issue #334, opened 2020-07-10
Domain and range of gist:geoContains, if correct, should be simply gist:Place


rjyounes:
Currently:

```
	<rdf:Description rdf:about="&gist;geoContains">
		<rdf:type rdf:resource="&owl;ObjectProperty"/>
		<rdf:type rdf:resource="&owl;TransitiveProperty"/>
		<rdfs:label rdf:datatype="&xsd;string">Geo Contains</rdfs:label>
		<rdfs:comment rdf:datatype="&xsd;string">Transitive version of geoDirectlyContains</rdfs:comment>
		<rdfs:domain>
			<owl:Class>
				<owl:unionOf rdf:parseType="Collection">
					<rdf:Description rdf:about="&gist;GeoPoint">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoRegion">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoRoute">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoSegment">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoVolume">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;Landmark">
				</owl:unionOf>
			</owl:Class>
		</rdfs:domain>
		<rdfs:range>
			<owl:Class>
				<owl:unionOf rdf:parseType="Collection">
					<rdf:Description rdf:about="&gist;GeoPoint">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoRegion">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoRoute">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoSegment">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoVolume">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;Landmark">
					</rdf:Description>
				</owl:unionOf>
			</owl:Class>
		</rdfs:range>
	</rdf:Description>
```

and gist:Place is: 
```
	<owl:Class rdf:about="&gist;Place">
		<rdfs:label rdf:datatype="&xsd;string">Place</rdfs:label>
		<rdfs:comment rdf:datatype="&xsd;string">Union of all the geo classes</rdfs:comment>
		<owl:equivalentClass>
			<owl:Class>
				<owl:unionOf rdf:parseType="Collection">
[...]

PR #335, opened 2020-07-10
Deprecate gist:Room. 

### LINKED ISSUE #102: Problems with gist:Room
The term specification does not define a "room" in the ordinary sense: e.g., elevators, staircases, hallways, and closets are direct parts of buildings but not rooms in the ordinary sense; and they can all have IDs. Also, a room in the ordinary sense can be a direct part of a wing, where the wing is a direct part of a building. In this case the room is a part of the building but not a direct part. According to the gist definition of Room, the wing is a Room and the room within a wing is not a Room. We could rename the class, but the usefulness of a class that includes elevators, staircases, hallways, wings, and some rooms but not others, is dubious; this is confirmed by the difficulty of finding an accurate name for the class - i.e., it's not a real concept. I suggest modifying the definition of gist:Room to be a semantically and coherent reflection of the concept of "room."

rjyounes:
Fixes #102 

Note that gist:Room also had to be removed from three union classes (which are actually one union class repeated three times; see #334)

### DIFF
--- a/gistCore.owl
+++ b/gistCore.owl
@@ -1901,8 +1901,6 @@
 					</rdf:Description>
 					<rdf:Description rdf:about="&gist;Landmark">
 					</rdf:Description>
-					<rdf:Description rdf:about="&gist;Room">
-					</rdf:Description>
 				</owl:unionOf>
 			</owl:Class>
 		</owl:equivalentClass>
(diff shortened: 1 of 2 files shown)


sa-bpelakh:
In order to keep this a minor change, deprecate the class _only_, leaving it in the other axioms. When we move to the next major release, deprecated classes will be removed entirely, and the axioms will be trimmed at that time.

rjyounes:
@sa-bpelakh IMO non-deprecated terms should not reference deprecated terms. Note that gistCore correctly does _not_ import gistDeprecated, while gistDeprecated correctly does import gistCore. Therefore, if one imports only gistCore, the axioms are rendered meaningless anyway. 

I would prefer to withdraw this change until the next major release.

rjyounes:
We can add a skos:editorialNote or rdfs:comment to ourselves to record this intention.

rjyounes:
Closing the PR; will make the change when we are ready for the next major release.

rjyounes:
Deferring issue till next major release.

PR #350, opened 2020-08-27
Generate RDF/XML from Turtle. Fixes #319

### LINKED ISSUE #319: Support multiple formats in gist bundling
When we bundle gist as a ZIP file to put on our website, should we only provide the Turtle serialization (#223), or should we also provide other formats?

Propose: we provide 3 formats, Turtle, JSON-LD, and RDF/XML.  

This may somewhat overlap with the issue about Widoco (#170)

sa-bpelakh:


marksem:
Fixes #319

Issue #343, opened 2020-08-17
Cannot create subclasses of Place since it is a Unioned class


mdfeickert:
When a class is defined as the union of a set of other classes, it is not possible to define additional subclasses that are not also subclasses of one of the members of the union. Is this by intention? It makes extension difficult.

Specific example for Place:

	<owl:Class rdf:about="&gist;Place">
		<rdfs:label rdf:datatype="&xs;string">Place</rdfs:label>
		<rdfs:comment rdf:datatype="&xs;string">Union of all the geo classes</rdfs:comment>
		<rdfs:isDefinedBy rdf:resource="https://ontologies.semanticarts.com/o/gistCore"/>
		<owl:equivalentClass>
			<owl:Class>
				<owl:unionOf rdf:parseType="Collection">
					<rdf:Description rdf:about="&gist;GeoPoint">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoRegion">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoRoute">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoSegment">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;GeoVolume">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;Landmark">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;Room">
					</rdf:Description>
				</owl:unionOf>
			</owl:Class>
		</owl:equivalentClass>
	</owl:Class>
	
Because this is a closed list, an ontology importing gist cannot define another direct subclass of Place.

uscholdm:
This is true for all practical purposes.  For the pedants in our midst, strictly speaking, you can create new subclasses, e.g. LandMarkOrRoom .  But it cannot bring any new instances to the table - hence the original objection.

PR #356, opened 2020-08-28
Removed union class equivalence assertion from gist:Artifact

### LINKED ISSUE #110: Remove equivalence to a union class from Artifact definition
When a class is defined as the union of a set of other classes, it is not possible to define additional subclasses. Is this by intention? It makes extension difficult. 

For example:

```
	<owl:Class rdf:about="&gist;Artifact">
		<rdfs:label rdf:datatype="&xs;string">Artifact</rdfs:label>
		<rdfs:comment rdf:datatype="&xs;string">An intentional person made thing, could be physical or content</rdfs:comment>
		<owl:intersectionOf rdf:parseType="Collection">
			<owl:Class>
				<owl:unionOf rdf:parseType="Collection">
					<rdf:Description rdf:about="&gist;Content">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;IntellectualProperty">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;Equipment">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;Building">
					</rdf:Description>
					<rdf:Description rdf:about="&gist;PhysicalSubstance">
					</rdf:Description>
				</owl:unionOf>
			</owl:Class>
			<owl:Restriction>
				<owl:onProperty rdf:resource="&gist;goal"/>
				<owl:someValuesFrom rdf:resource="&gist;Function"/>
			</owl:Restriction>
		</owl:intersectionOf>
	</owl:Class>
```
Because this is a closed list, an ontology importing gist cannot define another direct subclass of Artifact.

In fact, we see already this problem within gist itself. System is a subclass of Artifact, but it is not asserted to subclass any of the unioned classes. Do we really want to say that System is a direct subclass of one of these, but we don't know which? Seems odd. And none of the unioned classes seem like appropriate superclasses of System.

rjyounes:
Fixes #110.

@marksem Added a release note to encompass both this issue and #343 for Place.

### DIFF
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -180,12 +180,7 @@ gist:Behavior
 
 gist:Building
 	a owl:Class ;
-	rdfs:subClassOf [
-		owl:intersectionOf (
-			gist:Landmark
-			gist:Artifact
-		) ;
-	] ;
+	rdfs:subClassOf gist:Landmark, gist:Artifact ;
 	rdfs:label "Building"^^xsd:string ;
 	rdfs:comment "A man-made structure for dwelling or working."^^xsd:string ;
 	.
@@ -881,6 +876,7 @@ gist:Equipment
 	rdfs:label "Equipment"^^xsd:string ;
 	rdfs:comment "Tangible property other than land or buildings.  Any kind of equipment, could be machine, router, car etc."^^xsd:string ;
 	owl:equivalentClass [
+		a owl:Class ;
 		owl:intersectionOf (
 			gist:Artifact
 			gist:PhysicalIdentifiableItem
@@ -1949,7 +1945,6 @@ gist:PhysicalSubstance
 	a owl:Class ;
 	rdfs:subClassOf [
 		owl:intersectionOf (
-			gist:Artifact
 			[
 				a owl:Restriction ;
 				owl:onProperty gist:hasMagnitude ;
@@ -3993,4 +3988,3 @@ gist:viableRange
 		gist:_second
 	) ;
 	.
-
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -116,13 +116,13 @@ gist:AreaUnit
 
 gist:Artifact
 	a owl:Class ;
-	rdfs:label "Artifact"^^xsd:string ;
-	rdfs:comment "An intentional, person-made thing, which could be physical or content"^^xsd:string ;
-	rdfs:subClassOf [ 
+	rdfs:subClassOf [
 		a owl:Restriction ;
 		owl:onProperty gist:hasGoal ;
-		owl:someValuesFrom gist:Function ;			
+		owl:someValuesFrom gist:Function ;
 	] ;
+	rdfs:label "Artifact"^^xsd:string ;
+	rdfs:comment "An intentional, person-made thing, which could be physical or content"^^xsd:string ;
 	.
 
 gist:Aspect
@@ -184,8 +184,8 @@ gist:Building
 		owl:intersectionOf (
 			gist:Landmark
 			gist:Artifact
-		); 
-	];
+		) ;
+	] ;
 	rdfs:label "Building"^^xsd:string ;
 	rdfs:comment "A man-made structure for dwelling or working."^^xsd:string ;
 	.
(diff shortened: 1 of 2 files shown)


gistCore.ttl
```diff
@@ -195,7 +180,12 @@ gist:Behavior
 
 gist:Building
 	a owl:Class ;
-	rdfs:subClassOf gist:Landmark ;
+	rdfs:subClassOf [
+		owl:intersectionOf (
+			gist:Landmark
+			gist:Artifact
+		) ;
+	] ;
```
uscholdm, line 188: I would have expected to see:
```
gist:Building
	a owl:Class ;		
	rdfs:subClassOf gist:Landmark , gist:Artifact .
```
I think the meaning is different, but am not 100% sure.

gistCore.ttl
```diff
@@ -890,8 +881,8 @@ gist:Equipment
 	rdfs:label "Equipment"^^xsd:string ;
 	rdfs:comment "Tangible property other than land or buildings.  Any kind of equipment, could be machine, router, car etc."^^xsd:string ;
 	owl:equivalentClass [
-		a owl:Class ;
```
uscholdm, line 893: Did the serializer remove this?

gistCore.ttl
```diff
@@ -1946,18 +1938,21 @@ gist:PhysicalIdentifiableItem
...
-		] ,
-		[
-			a owl:Restriction ;
-			owl:onProperty gist:hasMagnitude ;
-			owl:someValuesFrom gist:Volume ;
-		]
-		;
+	rdfs:subClassOf [
+		owl:intersectionOf (
+			gist:Artifact
```
uscholdm, line 1943: This is a mistake that was already in gist.  The lump of dirt I have in my hand is not an Artifact.  It should be removed.

gistCore.ttl
rjyounes: I removed it - note that it's redundant, given the range of owl:equivalentClass. However, as above I'll add it back so we can push this through.

rjyounes:
I wrote one release note for both, so the language will be the same.

gistCore.ttl
rjyounes: The meaning is the same. Your way says everything that's a Building is both a Landmark and an Artifact. My way says that everything that's a Building lies at the intersection of Landmark and Artifact. These are logically identical. That said, I can do it either way, so I'll switch it to get this through more quickly.

rjyounes: Yours is stylistically simpler and probably more comprehensible, as well.

uscholdm: Your description made the equivalence clear - my brain must have been muddled!

uscholdm:
The definition of GeoRoute has OrderedCollection twice. Suggest to remove it from the subclass predicate. 
```
gist:GeoRoute
	a owl:Class ;
	rdfs:subClassOf
		gist:OrderedCollection ,   # THIS IS REDUNDANT
		gist:Place
		;
	rdfs:label "Geo Route"^^xsd:string ;
	rdfs:comment "An ordered set of GeoPoints that defines a path from starting point to ending point."^^xsd:string ;
	owl:equivalentClass [
		a owl:Class ;
		owl:intersectionOf (
			gist:OrderedCollection
			[
				a owl:Restriction ;
				owl:onProperty gist:hasDirectPart ;
				owl:someValuesFrom gist:GeoSegment ;
			]
		) ;
	] ;
	.
```

rjyounes:
@uscholdm Is the GeoRoute issue relevant here? If not, I suggest creating a new issue. I think it's more straightforward if we remove the subclass assertion altogether:
```
gist:GeoRoute
	a owl:Class ;
         # REMOVE ENTIRE SUBCLASS STATEMENT
	#rdfs:subClassOf
	#	gist:OrderedCollection ,   # THIS IS REDUNDANT
	#	gist:Place
	#	;
	rdfs:label "Geo Route"^^xsd:string ;
	rdfs:comment "An ordered set of GeoPoints that defines a path from starting point to ending point."^^xsd:string ;
	owl:equivalentClass [
		a owl:Class ;
		owl:intersectionOf (
			gist:OrderedCollection

			[
				a owl:Restriction ;
				owl:onProperty gist:hasDirectPart ;
				owl:someValuesFrom gist:GeoSegment ;
			]
		) ;
	] ;
	.
```

uscholdm:
>  I think it's more straightforward if we remove the subclass assertion altogether:

@rjyounes Yes, that is a perfectly and logically equivalent alternative go for it. 

Why is this closed? The work was pretty much done!!

rjyounes:
I'm not sure how it got closed. I've reopened now and will merge.

rjyounes:
Created new issue #361 for the GeoRoute fix.

PR #357, opened 2020-08-28
Deprecate gist:_unitedNations

### LINKED ISSUE #207: Deprecate _unitedNations
Doesn't belong in an upper ontology, and if someone needs to reference the United Nations, an external URI can be used.

I'm calling this a patch because we could consider its inclusion an error, in a sense. It's a debatable point, however.

rjyounes:
Fixes #207

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -44,6 +44,13 @@ Release X.x.x
 - Motivation/Rationale: Standardize release note format for readability.
 - Issue: [#338](https://github.com/semanticarts/gist/issues/338).
 
+*Deprecated `gist:_unitedNations`*
+
+- Deprecated `gist:_unitedNations` and moved from `gistCore` to `gistDeprecated`. Replaced the restriction on `gist:CountryGovernment` that it must be recognized by the United Nations with a restriction that it must be recognized by some Organization.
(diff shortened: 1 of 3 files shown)


docs/ReleaseNotes.md
```diff
@@ -44,6 +44,13 @@ Release X.x.x
 - Motivation/Rationale: Standardize release note format for readability.
 - Issue: [#338](https://github.com/semanticarts/gist/issues/338).
 
+*Deprecated `gist:_unitedNations`*
+
+- Deprecated `gist:_unitedNations` and moved from `gistCore` to `gistDeprecated`. Replaced the restriction on `gist:CountryGovernment` that it must be recognized by the United Nations with a restriction that it must be recognized by some Organization.
```
uscholdm, line 49: This explanation is correct, but a bit wordy. I wonder if a shorter style would be better. E.g.

Changed the restriction on CountryGovernment from
- `gist:recognizedBy value gist:_unitedNations` to
- `gist:recognizedBy some gist:Organization`

sa-bpelakh: That's a very OWL-y explanation, I think we should keep release notes to simple English (even if it is not quite as terse).

PR #360, opened 2020-08-31
Substitute version in release notes. Fixes #359

### LINKED ISSUE #359: Replace X.x.x with version number in release notes, if they exist
This will save the release manager from remembering this step in doing the release.

Config file should specify the name and location of the release notes file, if there is one.

Note that X.x.x can occur in multiple places, so it should be a full substitution.

sa-bpelakh:
Will still have to edit it permanently when we create the next version template after the release, and there is the pesky reminder text there as well, so not sure if it's worth it.

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -103,6 +103,9 @@ actions:
 - action: "copy"
   source: "{input}/docs"
   target: "{output}/Documentation"
+  replace:
+    from: "X.x.x"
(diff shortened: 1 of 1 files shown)


rjyounes:
Good point. If you're OK with dropping this, I think that's best.

sa-bpelakh:
Closing as per discussion.

PR #363, opened 2020-09-02
Trivial bug fixes to axioms in some geo terms

### LINKED ISSUE #64: Fix range and typo in definition of geoOccupies
The range of geoOccupies is the union of a single class (gist:Place). Should just be that class, directly. 
Interestingly, this is not the case for the file gistPlace9.0.0 in the internally released gist folder.  It has the range as the union of GeoRegion and GeoVolume.  This may or may not be intended, but in any event, the gist on git is different from what is in the internally released gist folder for 9.0.0

rjyounes:
Fixes #64, #334. Also a couple of tweaks to release notes.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -33,16 +33,23 @@ Release X.x.x
 
 *Made `gist:hasPhysicalLocation` transitive*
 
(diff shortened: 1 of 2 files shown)


rjyounes:
Merging without review since these are trivial fixes that have been pre-approved.

PR #368, opened 2020-09-09
Update handles new files

Jamie-SA:
Updates the release script after new file types were added.


sa-bpelakh:
@rjyounes Could you possibly test this script, Windows `bash` doesn't handle soft links well.

Jamie-SA:
In order to be successful you to unpack the archive, rename the *.json files to *.jsonld, and run it as `release-gist.bash 9.4.0`

sa-bpelakh (approved):
👍

PR #380, opened 2020-09-29
Generate legacy annotations. Fixes #379

### LINKED ISSUE #379: Add rdfs:comment and rdfs:labels into a separate file in case anyone wants them.
After Issue https://github.com/semanticarts/gist/issues/351 is implemented, I think we agreed to have the file around as an optiona import. Not sure if it should go into the deprecated file, or a new file called, say deprecatedAnnotations.

This way, we can keep the conversion to skos as a change of lesser significance. 

Michael

sa-bpelakh:
This will only `legacyAnnotations.ttl` via CONSTRUCT query to restore the old `rdfs:label`/`rdfs:comment` annotations. I don't know the right release notes verbiage or whether we should also have a JSON-LD and RDF/XML versions of this file. 

PS This requires onto_tool v0.10.0 or later (including the fix in https://github.com/semanticarts/ontology-toolkit/pull/47, which is not yet merged at this time).

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -99,7 +99,7 @@ actions:
     - "*.ttl"
 - action: "sparql"
   source: "{output}"
-  target: "{output}/legacyAnnotations.ttl"
+  target: "{output}/rdfsAnnotations.ttl"
   format: "turtle"
   includes:
     - "*.ttl"
@@ -114,15 +114,21 @@ actions:
     WHERE {{
       {{ ?entity skos:prefLabel ?label }}
       UNION
-      {{ ?entity skos:altLabel|
-                 skos:definition|
-                 skos:example|
-                 skos:scopeNote ?comment }}
+      {{
+        VALUES (?prop ?prefix ) {{
+          ( skos:altLabel "ALT: " )
+          ( skos:definition "DEFINITION: " )
+          ( skos:example "EXAMPLE: " )
+          ( skos:scopeNote "NOTE: " )
+        }}
+        ?entity ?prop ?text .
+        BIND(CONCAT(?prefix, ?text) as ?comment)
+      }}
     }}
 - action: "transform"
   tool: "serializer"
-  source: "{output}/legacyAnnotations.ttl"
-  target: "{output}/legacyAnnotations.ttl"
+  source: "{output}/rdfsAnnotations.ttl"
+  target: "{output}/rdfsAnnotations.ttl"
 - action: "copy"
   source: "{input}/LICENSE.txt"
   target: "{output}"
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -97,6 +97,32 @@ actions:
     to: "\\g<1>.jsonld"
   includes:
     - "*.ttl"
+- action: "sparql"
+  source: "{output}"
+  target: "{output}/legacyAnnotations.ttl"
+  format: "turtle"
+  includes:
+    - "*.ttl"
+  query: >
+    prefix gist: <https://ontologies.semanticarts.com/gist/>
+    prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+    prefix skos: <http://www.w3.org/2004/02/skos/core#>
+    CONSTRUCT {{
+      ?entity rdfs:label ?label ;
+              rdfs:comment ?comment .
+    }}
+    WHERE {{
+      {{ ?entity skos:prefLabel ?label }}
+      UNION
+      {{ ?entity skos:altLabel|
+                 skos:definition|
+                 skos:example|
+                 skos:scopeNote ?comment }}
+    }}
+- action: "transform"
+  tool: "serializer"
+  source: "{output}/legacyAnnotations.ttl"
+  target: "{output}/legacyAnnotations.ttl"
 - action: "copy"
   source: "{input}/LICENSE.txt"
   target: "{output}"


bundle.yaml
```diff
@@ -97,6 +97,32 @@ actions:
...
+      UNION
+      {{ ?entity skos:altLabel|
+                 skos:definition|
+                 skos:example|
+                 skos:scopeNote ?comment }}
+    }}
+- action: "transform"
+  tool: "serializer"
+  source: "{output}/legacyAnnotations.ttl"
+  target: "{output}/legacyAnnotations.ttl"
```
uscholdm, line 125: How can the source and target be the same file? 
As a sanity check, can I see a sample output of a run of this program?

sa-bpelakh: The serializer doesn't seem to mind if you use the same file for input and output. I tested it, and I got the following output (just the first part included, the whole thing is over 1500 lines).
```
@prefix gist: <https://ontologies.semanticarts.com/gist/> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

gist:Account
        rdfs:label "Account"^^xsd:string ;
        rdfs:comment "An agreement having a balance, as in a bank account, or credit card account, or Accounts Receivable account."^^xsd:string ;
        .

gist:Actuator
        rdfs:label "Actuator"^^xsd:string ;
        rdfs:comment "A device that can affect the real world via a message interface"^^xsd:string ;
        .

gist:Address
        rdfs:label "Address"^^xsd:string ;
        rdfs:comment "A reference to a place (real or virtual) that can be located by some routing algorithm, and where messages or things can be sent to or retrieved from. E.
g. PO Box or URL to a PDF file."^^xsd:string ;
        .

gist:Agreement
        rdfs:label "Agreement"^^xsd:string ;
        rdfs:comment "Something which two or more People or Organizations mutually commit to do."^^xsd:string ;
        .

gist:Area
        rdfs:label "Area"^^xsd:string ;
        rdfs:comment "A measurement of two-dimensional space."^^xsd:string ;
        .

gist:AreaUnit
        rdfs:label "Area Unit"^^xsd:string ;
        rdfs:comment "A unit of two-dimensional area, such as square inches or hectares."^^xsd:string ;
        .
```

uscholdm: OK - I'd like to have a quick peak at the whole output file so I can check for the more interesting cases.

uscholdm:
This loses a modest amount of information. For example:

```
gist:CoherentUnit
	rdfs:label "Coherent Unit"^^xsd:string ;
	rdfs:comment
		"A unit that is expressed in units that have no conversions.  It may be a simple unit.  It may also be a product or ratio unit that bottoms out in simple units."^^xsd:string ,
		"a simple unit: kilogram"^^xsd:string ,
		"coherent unit is the physics term for this, informally you might think of it as the standard unit for a given dimension."^^xsd:string ,
		"in principle, the CoherentUnit for a ProductUnit or RatioUnit can be inferred by recursively decomposing the products and ratios into their respective CoherentUnits, bottoming out in SimpleUnits"^^xsd:string ,
		"the standard unit for acceleration is meters per square second (feet per square second  requires a conversion)"^^xsd:string ,
		"CU"^^xsd:string 
		;
	.
```
This will be pretty confusing in Protege, say compared to:

```
gist:CoherentUnit
	rdfs:label "Coherent Unit"^^xsd:string ;
	rdfs:comment
		"DEFINITION: A unit that is expressed in units that have no conversions.  It may be a simple unit.  It may also be a product or ratio unit that bottoms out in simple units."^^xsd:string ,
		"EXAMPLE: a simple unit: kilogram"^^xsd:string,
		"EXAMPLE: the standard unit for acceleration is meters per square second (feet per square second  requires a conversion)"^^xsd:string ,
		"NOTE: coherent unit is the physics term for this, informally you might think of it as the standard unit for a given dimension."^^xsd:string ,
		"NOTE: in principle, the CoherentUnit for a ProductUnit or RatioUnit can be inferred by recursively decomposing the products and ratios into their respective CoherentUnits, bottoming out in SimpleUnits"^^xsd:string ,
                "ALTLABEL: CU"^^xsd:string ;
		;
	.
```

But frankly, it might not be worth the trouble. Users can always look at the original OWL.

uscholdm:
If you want to merge this, go ahead. We can always add enhancements later if we wish.

rjyounes (commented):
I agree with Michael's point that it would be best to have a faithful rendering of the original rdfs:comment. It should be an easy matter to do individual union clauses and some string transforms in the query based on which SKOS annotation it came from. If you agree that this can be done pretty quickly, I think it would be preferable to add it. Or at a minimum, uppercase the first letter of each string.

uscholdm:
Also: since these are not legacy annotatoins, they will always be up to date annotations, perhaps the file should be called rdfsAnnotations

sa-bpelakh:
@uscholdm I renamed the file as per your suggestion.
@rjyounes I have updated the query to retain the information, as follows:
```
gist:Category
        rdfs:label "Category"^^xsd:string ;
        rdfs:comment
                "DEFINITION: A concept or label used to categorize other instances informally. Things that can be thought of as types are usually Categories."^^xsd:string ,
                "EXAMPLE: Tags used in folksonomies; formal definitions from other systems."^^xsd:string ,
                "NOTE: Often a 'bucket' can be modeled either as an owl:Class or as a gist:Category. Use the latter if you don't care much about the formal structure of the different types, or if there is a whole hierarchy of types that are going to be managed by a group separate from the ontology developers. The formal structure may be defined elsewhere and linked to, if necessary."^^xsd:string
                ;
        .
```
@uscholdm was OK with it.

uscholdm:
Nicely done.

rjyounes:
Ah, I didn't think of doing it that way.

PR #390, opened 2020-10-22
Add standard pre-commit hook and instructions. Fixes #228.

### LINKED ISSUE #228: Add pre-commit hook to run serializer
Do we recommend the EDM Council hook or the one on the shared drive? If the latter, we need to add it to the repo. In either case we need to make sure the documentation is correct. The relevant documentation is gistStyleGuide.md.

sa-bpelakh:
I run all my commits from Git Bash, and this works for me, but it should be tested on Mac and maybe some Git UI to make sure it works. We may need to create some Powershell-specific version of this.


marksem (approved):
I approve of this as a correct implementation of git pre-commit hook to serialize. 

However, I note that with this pre-commit hook vs. manual serializing.... if Protege adds stuff (as it is wont to do), and people don't manually serialize and check diffs, they won't catch it pre-commit.

rjyounes (changes requested):
The script needs to be made executable. I think we can make it executable in the repo, can't we, so that this doesn't have to be changed by the user. 

An addition to the release notes is required.

It works fine on Mac.

rjyounes (approved):

PR #393, opened 2020-10-23
deprecate geoDirectly* properties (Fixes #328)

### LINKED ISSUE #328: remove gist:geoDirectlyContains and gist:geoDirectlyContainedIn
gist:geoDirectlyContains and gist:geoDirectlyContainedIn should be removed, because "direct containment" in the geographic area is problemmatic.  See #115 which discusses the geo direct containment issue at length.

marksem:
Ultimately, want to remove gist:geoDirectlyContains and
gist:geoDirectlyContainedIn because "direct containment" in
the geographic realm is problemmatic.
For now deprecate them to be removed in next major version.


rjyounes (changes requested):
Our current practice when we deprecate a term is to remove it from gistCore entirely and move it to gistDeprecated, along with the owl:deprecated annotation. Now I'm not sure why we don't use just owl:deprecated and keep it in place. Is it sort of an "out of sight out of mind" approach?

marksem:
Just noticed this:  Class `gist:GeoVolume` has this definition:
`(geoDirectlyContains some GeoPoint) and (hasMagnitude some Volume)`

So if we deprecate `geoDirectlyContains`, we should probably change this to use `geoContains`.

marksem:
@rjyounes  please re-review.  Note my comment above... I changed definition of GeoVolume per that comment.   If I did not, then I could not move the items to gistDeprecated.ttl (as gistCore does not import gistDeprecated).

rjyounes (approved):

rjyounes:
Looks good to me.

PR #399, opened 2020-11-02
Remove assertion of type DurationUnit from _second. Fixes #92.

### LINKED ISSUE #92: BaseUnit individuals should have type assertions indicating what dimension they measure
All gist:BaseUnit individuals (gist:_ampere, gist:_second, gist:_kilogram, etc.) are typed gist:BaseUnit, but only gist:_second has an additional type gist:DurationUnit to indicate what it is used to measure. Shouldn't we similarly assert that gist:_USDollar is a gist:CurrencyUnit, gist:_kilogram is a gist:MassUnit, etc.? There is nothing else in the ontology that I can see that further constrains them. I see that gist:MassUnit gist:hasBaseUnit gist:_kilogram, but from that one cannot infer that gist:_kilogram is a gist:MassUnit.  If I'm missing something, please let me know!

rjyounes:


rjyounes:
Patch change because the assertion can still be inferred.

sa-bpelakh (approved):
Looks good.

Issue #188, opened 2020-02-24
Define and implement conventions for local names


rjyounes:
This is a spin-off of #20, which we decided to limit to labels. Conventions articulated there for composing local names are repeated below. The gist Style Guide (forthcoming) also records these conventions, but only as proposals since there remains significant controversy.

```
Classes: camelcase, initial uppercase. Acronyms also camel-cased so that word boundaries are unambiguous - e.g., AmaGuideline not AMAGuideline, UriScheme not URIScheme. Alphanumeric characters only - e.g., Isbn10, not Isbn-10 or ISBN-10. ID is an exception: Merriam-Webster accepts this as a word in all-caps.

Properties: camelcase, initial lowercase, acronyms as above.

Both: non-standard abbreviations should not be used: e.g., hasUoM should be hasUnitOfMeaure.
```

rjyounes:
This is a minor change based on semver criteria.

rjyounes:
The issue will be raised at the gist council meeting on March 5 to canvas user reactions to the proposed URI changes (or indeed any URI changes).

For SA ontologists, Dan's proposed changes are now on the Semantic Arts One Drive for commenting. A link will be sent in an email.

rjyounes:
Comments from SA ontologists:

Dave McComb, March 1:
Does this automatically save?  Can others see my comments?
 
My general thoughts are two fold:
Are these suggestions an improvement?
Are they worth the hassle of migration
 
Obviously we have to answer them in that order.  It is possible they are an improvement but not worth the trouble.  However if they are not an improvement, then we don’t need to worry about the migration.
 
To me, most of them are not an improvement, and many are a deprovment (if that is a word, which apparently my spell check doesn’t think so).  But that’s just one ontologists opinion.

Michael Uschold, March 2:
Things evolved organically over many years, and now several different conventions are used.  Most individual changes are not improvements. The main idea was a collective improvement: to have consistent predictable naming conventions across the board.  One view shared by several of us is that this kind of consistency gives gist a cleaner and more professional look.

Dave McComb, March 2:
Some of this, like the desire to have verbs in the names, stems from a valid desire to telegraph the directionality of the predicate.  The skos:broader failed that, as half the population thinks it is pointing at the boarder term and half seem to think it means broaderThan.  So we definitely want to stomp out ambiguity (and we had a couple like that in gist, not sure if I recall any now)
 
[...]

PR #404, opened 2020-11-19
Refactored giver, getter and hasParty. Fixes #133

### LINKED ISSUE #133: Expand definition of giver and getter
The definition reads: "The active party, the one with the obligation or the one initiating the transfer."

I am using these predicates to point from an Employment resource to the employer and employee, which doesn't fit this definition. Either it's the wrong use of the predicates, or the definition needs to be expanded. Personally, it strikes me as a good use of these predicates. Can be generally useful to apply to complex relationship resources.

uscholdm:
- New property: `hasParticipant` replaces `hasParty`
  - Range: `Person` or `Organization`.
  - Domain: includes `Event`, `Agreement`, `Obligation`.
  - Has subproperties: `toAgent` and `fromAgent`

- Use `hasParticipant` wherever `hasParty` was used
- Use `toAgent` wherever `getter` was used
- Use `fromAgent` wherever `giver` was used
- Deprecate: `hasParty`, `getter` & `giver`.


sa-bpelakh (commented):
Looks OK, but do we have an issue with consistency of property naming, i.e. `fromAgent`/`toAgent` vs `hasParticipant` (as opposed to `hasFromAgent`/`hasToAgent`). When do we prefix properties with `has`/`is`, and when do we not?

uscholdm:
> we have an issue with consistency of property naming, i.e. fromAgent/toAgent vs hasParticipant (as opposed to hasFromAgent/hasToAgent). When do we prefix properties with has/is, and when do we not?

This is true. There are a lot of inconsistencies like this. That is a separate issue (https://github.com/semanticarts/gist/issues/188). I think its best to not too too many things at once.

marksem:
I feel like hasParticipant -- as a general property -- can have a broader range.  Would not restrict domain to Person or Org.

uscholdm:
@marksem said:

> I feel like hasParticipant -- as a general property -- can have a broader range. Would not restrict domain to Person or Org.

Another discussion you missed. I agree with you on this, but the majority felt that  'participate' implied some active intent. So inanimate objects need not apply for this role.  There might still be time to file a law suit and overturn the vote result.

marksem:
Regarding active "intent", if I say a process participates in a larger process, I am not implying agency on the process.   A chemical participates in a chain reaction.  Assuming participatesIn would be the logical inverse of hasParticipant, I again feel that more than Persons and Orgs participate in things.        Your honor, I move that Person/Org restrictions be stricken from the record!

uscholdm:
I could not agree mere!!! Pity you were not there to make your case when it came up. @rjyounes I suggest we revisit this on a future Thursday session.

rjyounes:
OK, let's put this on hold. @uschold Do you remember the issue number so I can change the label to "in review"? 

Maybe @marksem's disagreement is only with the wording - that is, I at least still think that there is value to distinguishing the kind of participation agents do as opposed to things, but perhaps we don't have the right term.

sa-bpelakh:
> OK, let's put this on hold and move the issue back to in review.
> 
> Maybe @marksem's disagreement is only with the wording - that is, I at least still think that there is value to distinguishing the kind of participation agents do as opposed to things, but perhaps we don't have the right term.

@rjyounes @marksem @uscholdm The way I see it , `hasParticipant` is the generic property that does not imply a range, whereas `hasParty` is the property that implies agency (and thus is the super-property of `toAgent`/`fromAgent`). How would you link a non-agent entity into an `Agreement` - e.g. the property into a loan?

sa-bpelakh (approved):
Changes look clean, assuming we have resolved the issue of the range on `hasParticipant`.

marksem (changes requested):
@uscholdm , I don't see where you addressed my concerns about the skos:definitions of toAgent and fromAgent.  Since we have enlarged their scope (now used in place of getter/giver), I think those textual definitions need to change.  You appear to have changed definitions on other properties instead?

gistCore.ttl
```diff
@@ -3776,13 +3782,15 @@ gist:timeZoneStandardUsed
...
+	rdfs:subPropertyOf gist:hasParty ;
 	rdfs:range [
 		a owl:Class ;
 		owl:unionOf (
 			gist:Organization
 			gist:Person
 		) ;
 	] ;
+	skos:altLabel "Getter"^^xsd:string ;
 	skos:definition "Indicates to whom (e.g. a message or shipment) is destined for."^^xsd:string ;
```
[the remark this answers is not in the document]
uscholdm: This is a good suggestion. On a separate note, Dave wants to use something for giver/getter that is NOT a Person or Org -- e.g. an account.  A variation being considered is this:

```
hasParticipant
  giver (no range)
    fromAgent
  getter (no range)
    toAgent
  hasParty (range: Person or Organization)
    fromAgent
    toAgent
```
This does two things:
1. Allows giver and getter to be used with more than just Person or Organization
2. Creates an abstract super-property to collect similar properties in one place

It does these things with minimal disruption:
- Keeps hasParty as originally conceived with range of Person or Org
- Keeps toAgent and fromAgent as originally conceived with range of Person or Org
- Keeps giver and getter 

Proposed release notes:
**Refactored `hasParty`, `giver` and `getter`**
- New property: `hasParticipant`
  - Domain: includes `Event`, `Agreement`, `Obligation`.
  - Range: none.
  - Has subproperties: `hasParty`, `giver` and `getter`
- `giver` and `getter` 
  - are no longer subproperties of `hasParty`
  - are direct subproperties of `hasParticipant`
- `toAgent` and `fromAgent` are now subproperties of `hasParty`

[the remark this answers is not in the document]
marksem: Discussion results:

hasParticipant (no range)  (new)
  fromAgent  (range Person or Org)   DROP (Dep)
  toAgent  range Person or Org       DROP (Dep)
  giver (no range)   hasGiver (dep giver)
  getter (no range)  hasGetter  (dep getter)

hasParty (range: Person or Organization)
  pro: Useful for agreements

Update all refs to from/toAgent to be hasGiver/hasGetter

uscholdm:
LATEST CHANGE NOTES:
**Refactored `hasParty`, `giver` and `getter`**
- 'giver' and 'getter'
  - Rename to `hasGiver` and `hasGetter`
  - Deprecate
  - No longer subproperties of `hasParty`
- New property: `hasParticipant`
  - No domain or range
  - Has subproperties: `hasGiver`, `hasGetter`, `hasParty`, `fromAgent` and `toAgent`
- Added annotations on `fromAgent` and `hasParty`

mkumba:
Didn’t we decide to deprecate fromAgent and toAgent and redo message to be hasGiver and hasGetter?


On Jan 15, 2021, at 5:04 PM, Michael Uschold <notifications@github.com<mailto:notifications@github.com>> wrote:



LATEST CHANGE NOTES:
Refactored hasParty, giver and getter

  *   'giver' and 'getter'
     *   Rename to hasGiver and hasGetter
     *   Deprecate
     *   No longer subproperties of hasParty
  *   New property: hasParticipant
     *   No domain or range
     *   Has subproperties: hasGiver, hasGetter, hasParty, fromAgent and toAgent
  *   Added annotations on fromAgent and hasParty

uscholdm:
> Didn’t we decide to deprecate fromAgent and toAgent and redo message to be hasGiver and hasGetter?

If we did, I don't specifically recall. It was not in the summary notes from @marksem MarkW, or if it was, I missed it. Can others chip in and confirm one way or the other?

rjyounes:
@uscholdm I don't recall that, but rethinking it, as Dave suggests, I don't know why we would need fromAgent and toAgent when we have hasGiver and hasGetter.

uscholdm:
> @rjyounes   I don't know why we would need fromAgent and toAgent when we have hasGiver and hasGetter.

Its tempting, but I'm not sure it is always correct.  Suppose I make a change to the beneficiary on my life insurance policy from person A to person B.  If there is an instance of a class called BeneficiaryChange the `fromAgent` is person A but they are not a giver.

In the case of a package or a message, the `fromAgent` is the sending agent or the originating agent which might not correspond to 'giving' anything.  The sender might be returning a borrowed item to its owner.  Both would nicely fit in a totally generic `gist:from` property, but I am the only proponent of that idea.

rjyounes:
@uscholdm  "Give" doesn't have to mean transfer of ownership; we can say "give back" for returning a borrowed item. Anyway, the same argument could be applied to putting money in someone's bank account, which is one of the uses cases for hasGiver; if that were borrowed money, then according to your limited sense of "give," that is not giving either. Even the original sense of giving an obligation is not a transfer of ownership. 

As for the first example, I think you are incorrectly exploiting the multi-faceted meaning of "from" (and most prepositions) to broaden the meaning of "fromAgent" beyond the agreed-upon sense. If "fromAgent" means "the source of something" - I believe we are keeping that part of the definition and simply broadening it to include a wider scope than message or shipment - person A is not the _source_ of the beneficiary change either: _you_ would be the fromAgent (the source of the beneficiary change) and person B would be the toAgent. I think the BeneficiaryChange is a three-way relationship, and we don't have a specific predicate for that happened to A.

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,22 @@
...
 
+Release X.x.x
+-----
+
+### Minor Updates
+
+**Refactored `hasParty`, `giver` and `getter`**
+
+- `giver` and `getter`
+  - Rename to `hasGiver` and `hasGetter`
```
rjyounes, line 12: We're using past tense in release notes.

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,22 @@
...
+  - Deprecate
```
rjyounes, line 13: We're using past tense in release notes.

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,22 @@
...
+- `hasGiver` and `hasGetter` are not subproperties of `hasParty` (as were `giver` and `getter`)
+- New property: `hasParticipant`
+  - No domain or range
+  - Has subproperties: `hasGiver`, `hasGetter`, `hasParty`, `fromAgent` and `toAgent`
+- Added annotations on `fromAgent` and `hasParty`
```
rjyounes, line 18: A bit confusing, since when I look at the code I don't know what annotations have been added. Can we just leave this out - or if you want it there just a couple of words to indicate what's changed?

rjyounes (changes requested):
Requesting just a couple of tiny changes.

gistDeprecated.ttl
```diff
@@ -156,6 +156,23 @@ gist:geoDirectlyContains
...
+	owl:propertyDisjointWith gist:giver ;
+	<http://www.w3.org/2004/02/skos/core#definition> "The recipient"^^xsd:string ;
+	<http://www.w3.org/2004/02/skos/core#prefLabel> "Getter"^^xsd:string ;
+	.
+
+gist:giver
+	a owl:ObjectProperty ;
+	rdfs:subPropertyOf gist:hasParticipant ;
+	owl:deprecated "true"^^xsd:boolean ;
+	<http://www.w3.org/2004/02/skos/core#definition> "The active party, the one with the obligation or the one initiating the transfer"^^xsd:string ;
```
rjyounes, line 172: Could you add the skos prefix to the file header so we can use the prefixed form of the annotations here?

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,22 @@
 gist Release Notes
 =====
 
+Release X.x.x
+-----
+
+### Minor Updates
+
+**Refactored `hasParty`, `giver` and `getter`**
```
rjyounes, line 9: Please add issue number to the release note.

gistCore.ttl
```diff
@@ -3050,13 +3050,15 @@ gist:expressedIn
 
 gist:fromAgent
 	a owl:ObjectProperty ;
+	rdfs:subPropertyOf gist:hasParty ;
 	rdfs:range [
```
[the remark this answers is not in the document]
uscholdm: We've had this conversation before.  Redundancy is generally to be avoided, but sometimes it reduces errors. For example, if someone changed the range of the super-property and forgot to change the range of the sub-properties, an error could result.  Another example is relying on transitivity of import. I don't have a horse in this race, so I could go either  way. I do think we should make the decision consciously, not just exclude redundancy as an absolute principle.

gistCore.ttl
```diff
@@ -3050,13 +3050,15 @@ gist:expressedIn
...
 		a owl:Class ;
 		owl:unionOf (
```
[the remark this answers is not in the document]
uscholdm: See above.

docs/ReleaseNotes.md
uscholdm: > Please add issue number to the release note.

Done.

gistDeprecated.ttl
uscholdm: > Could you add the skos prefix to the file header so we can use the prefixed form of the annotations here?

Good idea. Done.

docs/ReleaseNotes.md
uscholdm: I purposely left out the gory details thinking it's easy enough for people to look to see for themselves.  At some point (maybe not this one) more detail detracts.  I added a scope note to one and an example on the other.  I also added one word to the `skos:defintion` which I  chose not to mention. Its a matter of taste, I suppose.  Do you prefer this level of detail:

- Added the word 'event' to the `skos:definition` of `fromAgent`
- Added a `skos:scopeNote` to `fromAgent` 
- Added a `skos:example` to `hasParty`

docs/ReleaseNotes.md
uscholdm: OK

docs/ReleaseNotes.md
uscholdm: OK

gistCore.ttl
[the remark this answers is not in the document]
rjyounes: I think this comment was made when fromAgent was a subproperty of hasParty. It's no longer relevant.

gistCore.ttl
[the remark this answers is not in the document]
rjyounes: I think this comment was made when fromAgent was a subproperty of hasParty. It's no longer relevant.

rjyounes (approved):

rjyounes:
Michael will touch base with Mark to see if his comment has been addressed and the PR can be merged.

marksem (changes requested):
All looks good except for the skos:definitions of fromAgent and toAgent.  I find them gramatically incorrect and/or too specific on implying a domain of shipment or message only.

Suggestions to correct this:

fromAgent:
     the party that is the source of something (e.g. a message, shipment, etc.)

toAgent:
     the party that is the recipient of something (e.g. a message, shipment, etc.)

.gitignore
```diff
@@ -1,3 +1,5 @@
 *.~vsd*
 *_webDownload/
 .DS_Store
+gistCore.ttl.orig
+docs/ReleaseNotes.md.orig
```
sa-bpelakh, line 5:
```suggestion
*.orig
```
will prevent more editing later.

sa-bpelakh (commented):
Yay, compliant labels :) Are you running a build to verify before pushing?

uscholdm:
> Yay, compliant labels :) Are you running a build to verify before pushing?

No. I merged develop into the branch first. I also read the manual and acted accordingly.

sa-bpelakh (approved):

docs/ReleaseNotes.md
```diff
@@ -4,6 +4,22 @@ gist Release Notes
 Release X.x.x [UPDATE AT RELEASE TIME]
 -----
 
+### Minor Updates
+
+**Refactored `hasParty`, `giver` and `getter`**
```
rjyounes, line 9: I pushed a small change to the release notes, moving patch updates below minor updates. I didn't want to bother you with it.

rjyounes (approved):

Issue #405, opened 2020-11-19
Broaden meaning of Specification to include how to do things


uscholdm:
Spawned from issue https://github.com/semanticarts/gist/issues/60

uscholdm:
The current intended meaning for `gist:Specification` is as a set of acceptance criteria for something, e.g. a product or service.  The criteria can be used to determine whether something is in compliance. If you bought something, did you get what you thought you paid for.  Another example is a formal specification of a program, in software engineering.

In the enterprise it is common to (formally or informally) document how to do things, e.g. recipe, method or process.  There was never a home in gist for such things, so they were either added as orphan classes, or put under `Specification`, which was not strictly correct.  This issue is about broadening the definition of `Specification` to include specifications of **how to do** something as well as specifying **what** it means to be a given thing.

uscholdm:
DECISION: 
- Broaden the definition.
- For now, do not create two main subclasses for the two kinds of specification. It may or may not prove to be useful.

uscholdm:
Commend from @mkumba 

> Admittedly the word specification is ambiguous in English in this regard, but what are we gaining by also becoming ambiguous. I think the issue is as it was if something was a specification we would know what it was, now we're not going to be sure if it is evaluative, or generative.

> There are behaviors that we can imagine implementing on instances of these classes. If the instance is template, then saying something like "new()" should create something. If something is a specification saying something like "conformsTo" we should get a boolean. But if we mix them all together we won't know which ones conform to which behaviors.

Response from @uscholdm 

> So maybe that's a good reason to have two major subclasses, which could be called EvaluativeSpecificaton and GenerativeSpecification (or maybe HowToSpecification). We decided to not do that till there was a good reason to. There are some things that I model as Specification that I would not make subclasses of either, so I'd not want them to be a disjoint union. For example, in Platts there is a class called TimeOfDayInTimeZone which means 'the time of day on a 24 hour clock in a given time zone, ignoring daylight savings time.' It is informational, part of the description of an assessment, when it takes place. Having a disjoint union would nudge one into thinking one had to choose. I would not want to be forced to think about whether TimeOfDayInTimeZone or SpecEntry or various other things necessarily go into one place or the other. Unhelpful.

uscholdm:
Comment from @marksem 

[...]

PR #406, opened 2020-11-19
Broaden meaning of Specification to include how to do things.

uscholdm:
Addressing issue #405

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -6,6 +6,8 @@ Release X.x.x [ Release manager: update version number here and below in import 
 
 ### Minor Updates
 
+*Broadened text definition for `Specification` to include specifications of how to do things (e.g. method, plan)*
+
 *Converted rdfs annotations to skos annotations*
 
 - Description:
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -2250,7 +2250,8 @@ gist:SimpleUnitOfMeasure
 gist:Specification
 	a owl:Class ;
 	rdfs:subClassOf gist:Requirement ;
-	skos:definition "A set of requirements to be satisfied by a material, design, product, or service."^^xsd:string ;
+	skos:definition "A characterization of either  1) *what* something is meant to be (e.g. acceptance criteria) or 2) how something will, should or must be done to achieve some outcome (e.g. a plan, recipe or method)"^^xsd:string ;
+	skos:example "Catalog item for a product or service. Program specification. Recipe or process for manufacturing shampoo. Method for assessing prices."^^xsd:string ;
 	skos:prefLabel "Specification"^^xsd:string ;
 	.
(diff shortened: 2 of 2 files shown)


mkumba (commented):
Admittedly the word specification is ambiguous in English in this regard, but what are we gaining by also becoming ambiguous.  I think the issue is as it was if something was a specification we would know what it was, now we're not going to be sure if it is evaluative, or generative. 

There are behaviors that we can imagine implementing on instances of these classes.  If the instance is template, then saying something like "new()" should create something.  If something is a specification saying something like "conformsTo" we should get a boolean.  But if we mix them all together we won't know which ones conform to which behaviors.

uscholdm:
> @mkumba  Admittedly the word specification is ambiguous in English in this regard, but what are we gaining by also becoming ambiguous. I think the issue is as it was if something was a specification we would know what it was, now we're not going to be sure if it is evaluative, or generative.

So maybe that's a good reason to have two major subclasses, which could be called EvaluativeSpecificaton and GenerativeSpecification (or maybe HowToSpecification).  We decided to not do that till there was a good reason to.  There are some things that I model as Specification that I would not make subclasses of either, so I'd not want them to be a disjoint union. For example, in Platts there is a class called TimeOfDayInTimeZone which means 'the time of day on a 24 hour clock in a given time zone, ignoring daylight savings time.' It is informational, part of the description of an assessment, when it takes place. Having a disjoint union would nudge one into thinking one had to choose. I would not want to be forced to think about whether TimeOfDayInTimeZone or SpecEntry or various other things necessarily go into one place or the other.  Unhelpful.

rjyounes (commented):
Michael, I think this PR should be closed in order to consider Dave's comments on the issue. I think they have merit.

uscholdm:
@rjyounes They do have merit, and any extended discussions ought take place back at the issue. However, if we can reach quick agreement, we can probably stay here on the PR.

marksem:
I concur with the merit of Dave's concerns.  I'm sorry to have missed the issue review yesterday where the DECISION was made.  Because I still think  a Specification is a "what" but not a "how" (process) for the reasons Dave brings up. Furthermore, Specification is sub of Requirement.  That again makes me think it is a "what" but not a "how".  I think processes fit under Task Template. I thought that's what Task Template was pretty much about. 

As for "EvaluativeSpecificaton and GenerativeSpecification", my initial reaction is to strongly object.  These are starting to sound like BFO classes, i.e., not natural, common concepts, and I think that would start to undermine Gist's appeal.

uscholdm:
CLOSING due to lack of consensus. Discussion continuing at issue #405

rjyounes:
@marksem Totally agree about EvaluativeSpecification and GenerativeSpecification. Frankly, I can barely understand what they mean unless I think harder than I should have to.

PR #414, opened 2020-12-04
Merge develop to master for release 9.5.0

rjyounes:


marksem:
In the past, I have created a release branch, done the release (like have Jamie publish all stuff), and when that is done, merge the release branch to the master.   That way the master is never truly updated till "after the fact", when we *know* it really is 
 what was released.    Do we not do it that way?  I didn't see any mention of how we do this in the  ChangeAndReleaseManagement.md doc.

rjyounes:
OK, sounds good. I was wondering about the best way of doing this.

rjyounes:
@marksem Does this mean the release package doesn't get reviewed by another ontologist before sending it to the developers? Where would that come in?

Issue #227, opened 2020-04-22
Update all labels according to a standard convention (replaces #20)


DanCarey404:
Add/replace rdfs:label values according to the agreed-to standard.

rjyounes:
@DanCarey404 Can you please add a summary of the agreed-upon standard so that it is documented here? Then can #20 be closed?

rjyounes:
Replaces #20

DanCarey404:
**Per Rebecca's request, these are the labeling standards being implemented.**
#### Classes
- Sentence case
- Normalized to natural language standards. E.g., hyphens inserted, acronyms in all caps, etc.
  - Examples: _AMA guideline_, _ISBN-10_
  
#### Properties
- Same as classes, but initial lowercase 
- Examples: _has unit of measure_, _has SSN_.

rjyounes:
Just to clarify: this was not my _request_: it was my _proposal_, which the group discussed and agreed on. :)

rjyounes:
Why is this a minor rather than a patch change?

rjyounes:
This task needs an assignee.

uscholdm:
@sa-bpelakh Boris has a query for this. Can you pop it into this issue, for convenience?

sa-bpelakh:
What I have are SHACL rules that validate that labels are conformant: https://github.com/semanticarts/platts-ontology/blob/develop/shapes/ontologyShapes.ttl. They enforce the policy described above, and even detect acronyms in all caps (minimum of 2 letters, I believe) and ignore their casing. The current version does _not_ allow for numbers in class or property names, so if that's a requirement, we'll have to make some changes.

rjyounes:
I think we should allow numbers; e.g., hypothetically we could define classes like Shimano105Components, Iso639 (subclass of Category), TourDeFrance2020Racers, CharactersIn1984, ...

sa-bpelakh:
I can see that for a specific domain, but for the base gist? We can set up a fun game of [regex golf](https://alf.nu/RegexGolf) for labels.

rjyounes:
Maybe it's less likely in an upper ontology, but why exclude it in principle?

uscholdm:
> Maybe it's less likely in an upper ontology, but why exclude it in principle?

We get to choose our own stylistic conventions, as does each client project.  I don't think we want gist to have numbers in IRIs and a rule for this would find a one the looks exactly like a lower case el. So I vote to put it in our gist checks as a warning.

[...]

PR #428, opened 2021-02-01
Label validation

sa-bpelakh:
Created rules for Title Case, Lower Case and Sentence Case, and associated with classes and properties. Also mandated at least one `skos:definition` for each ontology entity. You can run `onto_tool` to see how the validation outputs, and we can convert these to warnings for the time being so they don't abort the build, until we can actually change the labels. 

I think the regex work with the current labels, but we may need to augment to support numbers, dashes, etc, which are currently all excluded.

### DIFF
--- a/ontologyShapes.ttl
+++ b/ontologyShapes.ttl
@@ -44,7 +44,8 @@ gshapes:ClassShape
 
 gshapes:LowerCase
 	a sh:PropertyShape ;
-	skos:definition "Enforces Lower Case, allowing for all-caps words for acronyms."^^xsd:string ;
+	skos:definition "Enforces Lower Case, allowing for all-caps words for acronyms, as well as numbers."^^xsd:string ;
+	skos:editorialNote "Entities which require a prefLabel that does not conform to this rule should be annotated with gist:nonConformingLabel to pass validation."^^xsd:string ;
 	sh:maxCount "1"^^xsd:integer ;
 	sh:minCount "1"^^xsd:integer ;
 	sh:name "Lower Case Label"^^xsd:string ;
@@ -87,6 +88,7 @@ gshapes:PropertyShape
 gshapes:SentenceCase
 	a sh:PropertyShape ;
 	skos:definition "Enforces Sentence Case, allowing for all-caps words for acronyms."^^xsd:string ;
+	skos:editorialNote "Entities which require a prefLabel that does not conform to this rule should be annotated with gist:nonConformingLabel to pass validation."^^xsd:string ;
 	sh:maxCount "1"^^xsd:integer ;
 	sh:minCount "1"^^xsd:integer ;
 	sh:name "Sentence Case Label"^^xsd:string ;
@@ -96,7 +98,8 @@ gshapes:SentenceCase
 
 gshapes:TitleCase
 	a sh:PropertyShape ;
-	skos:definition "Enforces Title Case, allowing for all-caps words for acronyms and lower case conjunctions and prepositions."^^xsd:string ;
+	skos:definition "Enforces Title Case, allowing for numbers, all-caps words for acronyms and lower case conjunctions and prepositions."^^xsd:string ;
+	skos:editorialNote "Entities which require a prefLabel that does not conform to this rule should be annotated with gist:nonConformingLabel to pass validation."^^xsd:string ;
 	sh:maxCount "1"^^xsd:integer ;
 	sh:minCount "1"^^xsd:integer ;
 	sh:name "Title Case Label"^^xsd:string ;
--- a/ontologyShapes.ttl
+++ b/ontologyShapes.ttl
@@ -74,6 +74,16 @@ gshapes:PropertyShape
 		;
 	.
 
+gshapes:SentenceCase
+	a sh:PropertyShape ;
+	skos:definition "Enforces Sentence Case, allowing for all-caps words for acronyms."^^xsd:string ;
+	sh:maxCount "1"^^xsd:integer ;
+	sh:minCount "1"^^xsd:integer ;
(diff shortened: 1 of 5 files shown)


rjyounes (changes requested):
Can you look at my title case proposal on #227 as of 11:15 this morning and incorporate it into your code? Then I can do a full review.

sa-bpelakh:
@rjyounes I updated the regex to be more flexible, support numbers + the list of prepositions, articles etc you listed. Here are the positive examples I tested:

```
One-Off
Seems OK
Single
Dangling And
Lower and More
Proper-Dash
With a 123 Number
Trailing Number 345
SSN 123-345-890
Various Articles a an the Conjunctions and but if for or nor so yet Prepositions as at by cum ere for in of off on out per pre pro qua re sub to up via Done
```
and here are some negative examples
```
With an Improper-dash
not capitalized
15 Leading Number
```

ontologyShapes.ttl
```diff
@@ -0,0 +1,106 @@
...
+	.
+
+gshapes:TitleCase
+	a sh:PropertyShape ;
+	skos:definition "Enforces Title Case, allowing for all-caps words for acronyms and lower case conjunctions and prepositions."^^xsd:string ;
+	sh:maxCount "1"^^xsd:integer ;
+	sh:minCount "1"^^xsd:integer ;
+	sh:name "Title Case Label"^^xsd:string ;
+	sh:path skos:prefLabel ;
+	sh:pattern "^([A-Z]([a-z]+|[A-Z]+)|[A-Z]([a-z]+|[A-Z]+)([- ]([A-Z]([a-z]+|[A-Z]+)|[0-9]+)| a[nst]?| the| and| but| i[fn]| [fn]?or| so| yet| by| cum| ere| o(ff?|n|ut)| p(re|er|ro)| qua| re| sub| to| up| via)*[ -]([A-Z]([a-z]+|[A-Z]+)|[0-9]+))$"^^xsd:string ;
```
rjyounes, line 104: If I were writing this I'd opt for readability over conciseness and spell out each word separately, but it's a stylistic decision.

ontologyShapes.ttl
```diff
@@ -0,0 +1,106 @@
...
+		a sh:SPARQLTarget ;
+		skos:definition "Target only non-blank node classes to avoid noise from OWL constructs."^^xsd:string ;
+		sh:prefixes <http://www.w3.org/2002/07/owl#> ;
+		sh:select "SELECT $this WHERE { $this a owl:Class . filter(!isblank($this)) }"^^xsd:string ;
+	] ;
+	.
+
+gshapes:LowerCase
+	a sh:PropertyShape ;
+	skos:definition "Enforces Lower Case, allowing for all-caps words for acronyms."^^xsd:string ;
```
rjyounes, line 47: Have you accounted for capitalized proper nouns?

ontologyShapes.ttl
```diff
@@ -0,0 +1,106 @@
...
+	sh:targetClass
+		owl:AnnotationProperty ,
+		owl:DatatypeProperty ,
+		owl:ObjectProperty
+		;
+	.
+
+gshapes:SentenceCase
+	a sh:PropertyShape ;
+	skos:definition "Enforces Sentence Case, allowing for all-caps words for acronyms."^^xsd:string ;
```
rjyounes, line 89: Have you accounted for capitalized proper nouns?

rjyounes (changes requested):
Regarding capitalized proper nouns in sentence and lower case, I didn't check the regexes closely, but at a minimum the comments need to be updated.

ontologyShapes.ttl
```diff
@@ -0,0 +1,106 @@
...
+	] ;
+	.
+
+gshapes:LowerCase
+	a sh:PropertyShape ;
+	skos:definition "Enforces Lower Case, allowing for all-caps words for acronyms."^^xsd:string ;
+	sh:maxCount "1"^^xsd:integer ;
+	sh:minCount "1"^^xsd:integer ;
+	sh:name "Lower Case Label"^^xsd:string ;
+	sh:path skos:prefLabel ;
```
rjyounes, line 51: It occurs to me that we should include ```skos:altLabel``` as well.

ontologyShapes.ttl
sa-bpelakh: This regex has reached levels of complexity where readability would always be a challenge. I felt that a concise form would make maintenance easier.

ontologyShapes.ttl
sa-bpelakh: Not in the regex, as there is no way to recognize a proper noun. I will add a `skos:editorialNote` to indicate that `nonConformingLabel` should be used to handle exceptions.

ontologyShapes.ttl
sa-bpelakh: Do we want any sort of governance for `altLabel`? It can be used for any purpose, and does not fall into the scope of the primary labels being standardized.

rjyounes: Good question; I'm not sure. We did say that if we want implementation-specific labels such as for column headers, we should define specific annotations for that. I think of `altLabel` as something like "USA" for the entity with `prefLabel` "United States of America." 

SKOS defines `altLabel` as "Acronyms, abbreviations, spelling variants, and irregular plural/singular forms may be included among the alternative labels for a concept."

I guess we should merge this PR and then bring up `altLabel` as a separate issue.

ontologyShapes.ttl
rjyounes: Never mind, this and the other comment to this effect were momentary lapses on my part.

rjyounes (approved):

PR #429, opened 2021-02-15
Add property gist:description. Fixes #425.

### LINKED ISSUE #425: Add gist:description
We have many places where we'd like to add a user description field on various instance data and there doesn't seem to be a good existing choice for this. It seems general enough that it could be a candidate to go into gist.

rjyounes:
Add datatype property ```gist:description```.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,6 +1,15 @@
 gist Release Notes
 =====
 
+Release X.x.x [ADD APPROPRIATE VERSION NUMBER FOR RELEASE]
+-----
+
+### Minor Updates
+
+- Added datatype property `gist:description` for describing instance data. Issue [#425](https://github.com/semanticarts/gist/issues/425).
+
+Import URL: <https://ontologies.semanticarts.com/o/gistCoreX.x.x>. [ADD APPROPRIATE VERSION NUMBER FOR RELEASE]
+
 Release 9.5.0
 -----
 
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -2978,6 +2978,13 @@ gist:describedIn
 	skos:prefLabel "Described In"^^xsd:string ;
 	.
 
+gist:description
+	a owl:DatatypeProperty ;
+	skos:definition "A statement about someone or something's attributes or characteristics."^^xsd:string ;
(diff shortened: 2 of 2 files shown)


docs/ReleaseNotes.md
```diff
@@ -1,6 +1,15 @@
...
 =====
 
+Release X.x.x [ADD APPROPRIATE VERSION NUMBER FOR RELEASE]
+-----
+
+### Minor Updates
+
+- Added datatype property `gist:description` for describing instance data. Issue [#425](https://github.com/semanticarts/gist/issues/425).
+
+Import URL: <https://ontologies.semanticarts.com/o/gistCoreX.x.x>. [ADD APPROPRIATE VERSION NUMBER FOR RELEASE]
```
uscholdm, line 11: Why is this here? There is non corresponding change in any files.

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,15 @@
 gist Release Notes
 =====
 
+Release X.x.x [ADD APPROPRIATE VERSION NUMBER FOR RELEASE]
+-----
+
+### Minor Updates
+
+- Added datatype property `gist:description` for describing instance data. Issue [#425](https://github.com/semanticarts/gist/issues/425).
```
uscholdm, line 9: Here is says for instance data. There is no mention of this in the actual ontology. Hard to know what is intended. Maybe add a scope note about intended usage?

uscholdm (changes requested):
Clarify meaning of `gist:description`.
Remove or explain redundant text in release notes.

docs/ReleaseNotes.md
rjyounes: This is the URL we provide for users to import gist. The bracketed note is addressed to the release manager and will be deleted in the final text.

sa-bpelakh (approved):
Perhaps a `skos:scopeNote` on the property to indicate that it is a description of an instance entity providing greater detail than a `prefLabel`? Otherwise, looks fine.

docs/ReleaseNotes.md
rjyounes: Good idea, I added a scope note.

PR #442, opened 2021-03-05
Don't include extraneous content in release build. Fixes #438

### LINKED ISSUE #438: Modify build to not include validation artifacts
Directory built for release should not include:
* SHACL RDF
* ValidationAnnotations RDF
* Validation results

sa-bpelakh:
Specifically enumerated included files so that random RDF does not get included in the release. Reordered serialization action to streamline transform.

@dhurlburtusa Can you please run a sample build and make sure only the expected contents make it into the generated distribution?

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -5,7 +5,6 @@ variables:
   input: "."
   rdf-toolkit: "{input}/tools/rdf-toolkit.jar"
   output: "{name}{version}_webDownload"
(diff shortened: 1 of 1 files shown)


dhurlburtusa (approved):
LGTM.
I ran build from develop branch and noted the files it puts in the bundle.
Then I ran build from PR branch and noted that the extraneous files were not in the bundle.

PR #447, opened 2021-04-05
Change gist:identifies from InverseFunctionalProperty to FunctionalProperty (bug fix). Fixes #180.

### LINKED ISSUE #180: gist:identifies should be FunctionalProperty, not InverseFunctionalProperty
If ID-A identifies Person-P1, and ID-A identifies Person-P2, then Person-P1 owl:sameAs Person-P2. This is a functional property, not an inverse functional property.

The inverse, identifiedBy, is correctly asserted to be an inverse functional property, and the inverse of an inverse functional property is a functional property.

Despite modifying inferences, it's a patch fix because it corrects an error.

rjyounes:


sa-bpelakh (approved):
👍

PR #448, opened 2021-04-05
Remove gist:Room. Fixes #102

### LINKED ISSUE #102: Problems with gist:Room

rjyounes:
Since this is for major release 10.0.0, the class is removed rather than deprecated.

### DIFF
--- a/docs/ChangeAndReleaseManagement.md
+++ b/docs/ChangeAndReleaseManagement.md
@@ -43,7 +43,7 @@ Additional notes:
 
 - Correction of an error, even if not backward-compatible, does not require a major release. The expectation is that users will not have implemented against an obvious error. This would be a patch.
 - When a local name is altered, the original term is deprecated to make it a minor rather than major change. The deprecated term receives an `owl:equivalentClass` or `owl:equivalentProperty` assertion to the new term. Deprecated terms may be removed in a future major release. Deprecated terms reside in the `gistDeprecated.ttl` file. If a user wants to use a deprecated term, he/she should import this file into his/her ontology, which in turn imports `gistCore.ttl` and thus all of gist.
-- Major changes should have a significant impact aside from technically modifying inferencing if this is low-impact. E.g.,when an equivalent class axiom to a union class is changed to a subclass axiom in order to allow new subclasses to be defined.
+- Major changes should have a significant impact aside from technically modifying inferencing if this is low-impact. E.g., when an equivalent class axiom to a union class is changed to a subclass axiom in order to allow new subclasses to be defined.
 
 Releases
 -----
--- a/docs/ChangeAndReleaseManagement.md
+++ b/docs/ChangeAndReleaseManagement.md
@@ -33,7 +33,7 @@ Version numbers are of the form X.x.x (major.minor.patch). We follow [Semantic V
 - **Major:** Non-backward-compatible (i.e., reasoning produces different results).
   - Examples: adding a restriction, domain, range.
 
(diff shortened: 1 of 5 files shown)


docs/ReleaseNotes.md
```diff
@@ -1,23 +1,36 @@
...
+  - `giver` and `getter`
+    - Renamed to `hasGiver` and `hasGetter`
+    - The newly named versions are no longer subproperties of `hasParty`
+    - Deprecated `giver` and `getter`
+  - New property: `hasParticipant`
+    - No domain or range
+    - Has subproperties: `hasGiver`, `hasGetter`, `hasParty`, `fromAgent` and `toAgent`
+  - Added a `skos:scopeNote` to `fromAgent`
+  - Added a `skos:example` to `hasParty`
+  - Updated `skos:definition`s for `toAgent` and `fromAgent`
```
sa-bpelakh, line 33: Please revert this change. It _seems_ like the whitespace is extraneous, but when converted to HTML (via build), the old indentation gives the desired result:
image
but without the spaces, the nested lists don't render properly:
image

docs/ChangeAndReleaseManagement.md
```diff
@@ -43,7 +43,7 @@ Additional notes:
 
 - Correction of an error, even if not backward-compatible, does not require a major release. The expectation is that users will not have implemented against an obvious error. This would be a patch.
 - When a local name is altered, the original term is deprecated to make it a minor rather than major change. The deprecated term receives an `owl:equivalentClass` or `owl:equivalentProperty` assertion to the new term. Deprecated terms may be removed in a future major release. Deprecated terms reside in the `gistDeprecated.ttl` file. If a user wants to use a deprecated term, he/she should import this file into his/her ontology, which in turn imports `gistCore.ttl` and thus all of gist.
-- Major changes should have a significant impact aside from technically modifying inferencing if this is low-impact. E.g.,changing an equivalent class axiom to a union class to a subclass axiom in order to allow new subclasses to be defined. 
+- Major changes should have a significant impact aside from technically modifying inferencing if this is low-impact. E.g.,when an equivalent class axiom to a union class is changed to a subclass axiom in order to allow new subclasses to be defined.
```
sa-bpelakh, line 46:
```suggestion
- Major changes should have a significant impact aside from technically modifying inferencing if this is low-impact. E.g., when an equivalent class axiom to a union class is changed to a subclass axiom in order to allow new subclasses to be defined.
```

sa-bpelakh (changes requested):
Actual changes look fine, minor issue w/ documentation formatting.

sa-bpelakh (approved):
You are going to have to educate me about this Markdown lint tool you are using 🤔

docs/ReleaseNotes.md
rjyounes: I've fixed this. Can you check if there are similar problems with other nested lists in the file? We have a two-level nested list in 9.5.0 notes. This is the only three-level one we have.

PR #450, opened 2021-04-06
Fix formatting so that markdown2html properly creates nested lists.

sa-bpelakh:


rjyounes (approved):
I removed some extraneous spaces I found based on warnings from my markdown de-linter. Otherwise all good.

PR #464, opened 2021-04-28
Has direct x has x pattern. Fixes #115.

### LINKED ISSUE #115: The hasX/hasDirectX pattern is problematic
The hasX/hasDirectX pattern is problematic in that hasDirectX often reflects a modeling choice about granularity that has nothing to do with the world. A different modeling choice can make a prior assertion untrue.   For example:

ORIGINAL GRANULARITY:
	USA 			hasDirectPart 		WashingtonState
	WashingtonState 	hasDirectPart 		Seattle

REVISED GRANULARITY: 
	USA 			hasDirectPart 		WashingtonState
	WashingtonState 	hasDirectPart 		KingCounty
	KingCounty 		hasDirectPart 		Seattle

Suddenly WashingtonState hasDirectPart Seattle is no longer true, but the world did not change, only the choice of granularity. We normally think of hasX being inferred from hasDirectX assertions. That may be backwards.  Maybe we need to infer hasDirectX assertions from the fact that there are no intermediary assertions. 

So, from the following
USA 			hasPart 		WashingtonState
WashingtonState 	hasPart 		KingCounty
KingCounty 		hasPart 		Seattle
A closed world assumption that there are no other hasPart links involving the above
 
We can infer: 
USA 			hasDirectPart 		WashingtonState
WashingtonState 	hasDirectPart 		KingCounty
KingCounty 		hasDirectPart 		Seattle

I’m not suggesting we do this, I'm just illustrating the problem.

At the moment I have not worked out a good proposal that addresses the above concern and still retains the intended benefits of the original pattern. Does anyone have experience that demonstrates the intended benefits have ever become real?   If not, we might want to get rid of the hasDirectX / hasX pattern and just use a single hasX transitive property?

uscholdm:
Addresses Issue #115


sa-bpelakh (changes requested):
Please reformat these very long annotations as multi-line literals - we should not have paragraphs worth of text run on in a single line. Would make the review easier as well.

Also I take issue with the 'conversational' tone of the `scopeNote` - 2nd person imperatives don't belong in formal documentation, IMHO.

rjyounes:
I agree with Boris.

rjyounes (changes requested):
See inline comment.

rjyounes:
Please make sure to add "Fixes #xxx" in the title of the PR so that it can easily be correlated with the issue. Thanks.

uscholdm:
> Please reformat these very long annotations as multi-line literals

Is there a recommended number of characters per line? Used to be 80 was a standard, but that seems a bit long these days.

sa-bpelakh:
> Is there a recommended number of characters per line? Used to be 80 was a standard, but that seems a bit long these days.

80 is still fine, easily viewable on most screens. For example on my (large) monitor this comment window is 120 chars wide.

marksem:
In my experience, embedding newlines in ontology annotations (or other string content) has been considered "bad practice" b/c we don't know how it will be rendered.  It is easier to "line-wrap" a string w/o embedded newlines."

marksem:
In multi-paragraph content, I understand the need for embedded double new-lines to break up the paragraphs.

uscholdm:
> Also I take issue with the 'conversational' tone of the `scopeNote` - 2nd person imperatives don't belong in formal documentation, IMHO.

I don't mind either way.  If it's important, we might put something into the style guide.  No second person imperatives.
- What about second person non-imperatives, 
- What about imperatives that are not second person (if there is such a thing).

In the mean time, for expedience, I made an attempt at pretending that I did not want to have a conversation with the reader.  If its still not what you want, change it to suit.

rjyounes:
I would say: no second person period. No imperatives (too strong, too imperious).

rjyounes (changes requested):
@uscholdm I took the liberty of committing some very minor punctuation and wording adjustments in order to reduce the cycles we have to go through to merge this. However, I still don't understand what is meant by "rather than choosing appropriate granularity" in a couple of the scopeNotes. Can you possibly state these more clearly?

uscholdm:
The idea about granularity is this.  If there is a direct part link that one wishes to break apart by inserting an intermediate part, that is more granular.   This is abundantly clear with the geo example in the original issue #115 I was torn between leaving the comment out about granularity completely, or to explain it ever further.  I chose middle ground hoping that people would figure it out. Maybe too optimistic.  

=============
Original explanation 

The hasX/hasDirectX pattern is problematic in that hasDirectX often reflects a modeling choice about granularity that has nothing to do with the world. A different modeling choice can make a prior assertion untrue.   For example:

ORIGINAL GRANULARITY:
- USA 			hasDirectPart 		WashingtonState
- WashingtonState 	hasDirectPart 		Seattle

REVISED GRANULARITY: 
- USA 			hasDirectPart 		WashingtonState
- WashingtonState 	hasDirectPart 		KingCounty
- KingCounty 		hasDirectPart 		Seattle

Suddenly WashingtonState hasDirectPart Seattle is no longer true, but the world did not change, only the choice of granularity.

rjyounes:
I can live with the phrase about granularity, though I think your following example explains it more clearly.

uscholdm:
> In my experience, embedding newlines in ontology annotations (or other string content) has been considered "bad practice" b/c we don't know how it will be rendered. It is easier to "line-wrap" a string w/o embedded newlines."

@marksem makes an excellent point (albeit one that will soon be blunted). Mark warns against this sort of ugliness:
image

It may be ugly but it is easy to fix by widening the window a bit. It's impossible to do that with long comment lacking explicit line breaks.  So I'm in favor of leaving in the line breaks for long comments.

sa-bpelakh (approved):
Looks good.

Jamie-SA:
@uscholdm it is not impossible, many/most text editors will wrap long lines if configured to do so.

But the real question is are you changing the formatting based on your convenience or based on what is the correct way to format text?

I don't really care that much which way you choose but I do find it interesting that I know several of you argue against making ontology decisions based on developer requirements, but I would say this is exactly what you are doing by choosing to put line formatting into the annotation because it is easier for you to read.

My $0.02 is I agree with Mark that you shouldn't put arbitrary formatting line breaks into annotations. It makes it harder to display correctly (with proper text flow) when using (or creating) other tooling.

rjyounes:
I agree with @Jamie-SA and @marksem. How are we going to resolve this? @pwin is anxious to do the release.

uscholdm:
> I agree with @Jamie-SA and @marksem. How are we going to resolve this? @pwin is anxious to do the release.

I don't care too much either  way.  The argument about putting things into the ontology for the convenience of developers is reasonable.   @sa-bpelakh  raised the issue. Are you ok with the long comments?

As far as this release goes, don't let this hold it up. We can always add or remove line breaks on a later release, if anyone cares enough.

rjyounes:
OK, this is now up to @sa-bpelakh, since he initially asked for the multi-line strings.

sa-bpelakh:
OK, since the majority favors not breaking up the strings, and letting the tools take care of it, I am willing to cede on this point. It was just a matter of convenience for me, and not worth the argument. I have already approved the PR, so whether you decide to undo the line breaks before merging or not, I OK either way - the text itself looks good.

rjyounes (approved):

rjyounes:
BTW, I think it's only in GitHub that the single line doesn't wrap. My text editor wraps automatically (or you can set it to do so).

uscholdm:
> BTW, I think it's only in GitHub that the single line doesn't wrap. 

@rjyounes It's true that most tools will wrap lines. Another exception is DiffMerge - to my annoyance.

PR #467, opened 2021-05-01
Fix nested list indents in ReleaseNotes.

sa-bpelakh:
I had to make a small fix in ReleaseNotes for proper HTML rendering. This is due to a bug in the python Markdown module that messes up nested lists. As soon as you inspect, I can proceed w/ the release process, I have already regenerated the archive as part of testing this patch.


rjyounes (approved):
This is the same auto-correction we've had to make before, though I've turned off that rule in the delinter config (markdown.json), so I'm not sure why it keeps reappearing. I'll look into it at some point.

Issue #463, opened 2021-04-27
Add predefined MimeType instances for the most common types


sa-bpelakh:
I suggest:
* `text/plain`, `text/html`, `text/csv`
* `application/json`, `application/ld+json`
* `application/rdf+xml`, `application/turtle`, `text/turtle`, `application/n-triples`
* `application/sparql-results+json`,`application/sparql-results+xml`

Need a good way to create local names for these, e.g.
```
gist:_MimeType_ApplicationJson
    a gist:MimeType ;
    skos:prefLabel "application/json" .
```

rjyounes:
Aha! We are developing a set of conventions for that (as you know). Alphanumeric only, word boundaries (I.e., capital letter) at site of removed alphanum. So exactly what you show.

I like these.

Jamie-SA:
URIs for the mime types already exist, but I haven't found a document of them so I created triples I needed as follows:
```
@prefix mime: <https://www.iana.org/assignments/media-types/> .
mime: {
  mime:text\/html rdf:type gist:MimeType
  ; skos:prefLabel "HTML"
  ; gist:uniqueText "text/html"
  .
  mime:text\/turtle rdf:type gist:MimeType
  ; skos:prefLabel "Turtle"
  ; gist:uniqueText "text/turtle"
  .
  mime:application\/trig rdf:type gist:MimeType
  ; skos:prefLabel "TriG"
  ; gist:uniqueText "application/trig"
  .
  mime:application\/n-quads rdf:type gist:MimeType
  ; skos:prefLabel "N-Quads"
  ; gist:uniqueText "application/n-quads"
  .
  mime:application\/n-triples rdf:type gist:MimeType
  ; skos:prefLabel "N-Triples"
  ; gist:uniqueText "application/n-triples"
  .
  mime:application\/rdf\+xml rdf:type gist:MimeType
  ; skos:prefLabel "RDF/XML"
  ; gist:uniqueText "application/rdf+xml"
  .
  mime:application\/json rdf:type gist:MimeType
  ; skos:prefLabel "JSON"
  ; gist:uniqueText "application/json"
  .
  mime:application\/ld\+json rdf:type gist:MimeType
  ; skos:prefLabel "JSON-LD"
  ; gist:uniqueText "application/ld+json"
  .
}
```

Jamie-SA:
If you access the URIs you will get information about it.
[...]

PR #477, opened 2021-05-17
Rename MimeType to MediaType, create ontology to declare instances. Fixes #434, #463

### LINKED ISSUE #434: Out with MimeType, In with MediaType Proposal
Based on what I've read in [RFC 6838](https://tools.ietf.org/html/rfc6838) (which obsoletes [RFC 4288](https://tools.ietf.org/html/rfc4288), which obsoletes [RFC 2048](https://tools.ietf.org/html/rfc2048)), especially the historical note: https://tools.ietf.org/html/rfc6838#section-1.1, seems like we should be using the term "media type" instead of "mime type". I couldn't find an article actually stating that "mime type" is obsoleted. Since MIME is still valid/relevant, I would take "mime type" to mean a subset of "media type". That is, a MIME Type is a media type used in the MIME context.

I assume that `gist:MimeType` is actually intended to be any "media type" as defined by RFC 6838.

So, how about the following for Gist v9.6:
- Adding `gist:MediaType`.
- Deprecating `gist:MimeType` in favor of `gist:MediaType`.
- Adding `gist:MediaType owl:equivalentClass gist:MimeType`.

In Gist v10:
Remove `gist:MimeType`.

Even [RFC 2046](https://tools.ietf.org/html/rfc2046) - "Multipurpose Internet Mail Extensions (MIME) Part Two: Media Types" doesn't use "MIME Type". It uses "media type". Same with [RFC 2045](https://tools.ietf.org/html/rfc2045) - "Multipurpose Internet Mail Extensions (MIME) Part One: Format of Internet Message Bodies".

Very few of the MIME related RFCs actually use "MIME type". Many of the RFCs that do contain "media type" have been obsoleted by other newer RFCs.

https://en.wikipedia.org/wiki/Media_type starts with "A media type (formerly known as MIME type)" with a footnote to https://www.iana.org/assignments/media-types/media-types.xhtml.

Maybe something to bring up at the next Gist Council.

We can also consider redefining MimeType to be the type of a [Mime artist](https://en.wikipedia.org/wiki/Mime_artist). 😅

sa-bpelakh:
@mkumba There was extensive debate about this issue, and your input was requested. I followed the approach @Jamie-SA and I favored, which was to create annotations for existing IANA media type pages by adding types and labels.

@rjyounes and some others said we should not be creating statements using entities we do not own as subjects, and suggested I follow this pattern instead:

```
gist:_MediaType_TextHtml 
    a gist:MediaType; 
    gist:uniqueText "text/html";
    owl:sameAs iana:text\/html .
```

Creating a duplicate instance in the `gist` namespace for each of the media types. Do you have strong feeling in either direction?

### DIFF
--- a/gistMediaTypes.ttl
+++ b/gistMediaTypes.ttl
@@ -1,0 +1,93 @@
+# imports: https://ontologies.semanticarts.com/o/gistCore
+
+@prefix gist: <https://ontologies.semanticarts.com/gist/> .
+@prefix mimeapp: <https://www.iana.org/assignments/media-types/application/> .
+@prefix mimetxt: <https://www.iana.org/assignments/media-types/text/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix xml: <http://www.w3.org/XML/1998/namespace> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+<https://ontologies.semanticarts.com/o/gistMediaTypes>
+	a owl:Ontology ;
+	owl:imports <https://ontologies.semanticarts.com/o/gistCore> ;
+	owl:versionIRI <https://ontologies.semanticarts.com/o/gistMediaTypesX.x.x> ;
+	skos:definition "Definitions of IANA Media Types."^^xsd:string ;
+	skos:prefLabel "gistMediaTypes"^^xsd:string ;
+	gist:license "https://creativecommons.org/licenses/by-sa/3.0/"^^xsd:string ;
+	.
+
+mimeapp:json
+	a gist:MimeType ;
+	skos:prefLabel "JSON"^^xsd:string ;
+	gist:uniqueText "application/json"^^xsd:string ;
+	.
+
+<https://www.iana.org/assignments/media-types/application/ld+json>
+	a gist:MimeType ;
+	skos:prefLabel "JSON-LD"^^xsd:string ;
+	gist:uniqueText "application/ld+json"^^xsd:string ;
+	.
+
+mimeapp:n-quads
+	a gist:MimeType ;
+	skos:prefLabel "N-Quads"^^xsd:string ;
+	gist:uniqueText "application/n-quads"^^xsd:string ;
+	.
+
+mimeapp:n-triples
+	a gist:MimeType ;
+	skos:prefLabel "N-Triples"^^xsd:string ;
+	gist:uniqueText "application/n-triples"^^xsd:string ;
+	.
+
+<https://www.iana.org/assignments/media-types/application/rdf+xml>
+	a gist:MimeType ;
+	skos:prefLabel "RDF/XML"^^xsd:string ;
+	gist:uniqueText "application/rdf+xml"^^xsd:string ;
+	.
+
+<https://www.iana.org/assignments/media-types/application/sparql-results+json>
+	a gist:MimeType ;
+	skos:prefLabel "SPARQL 1.1 Query Results JSON"^^xsd:string ;
+	gist:uniqueText "application/sparql-results+json"^^xsd:string ;
+	.
+
+<https://www.iana.org/assignments/media-types/application/sparql-results+xml>
+	a gist:MimeType ;
+	skos:prefLabel "SPARQL 1.1 Query Results XML"^^xsd:string ;
+	gist:uniqueText "application/sparql-results+xml"^^xsd:string ;
+	.
+
+mimeapp:trig
+	a gist:MimeType ;
+	skos:prefLabel "TriG"^^xsd:string ;
+	gist:uniqueText "application/trig"^^xsd:string ;
+	.
+
+mimetxt:csv
+	a gist:MimeType ;
+	skos:prefLabel "CSV"^^xsd:string ;
+	gist:uniqueText "text/csv"^^xsd:string ;
+	.
+
+mimetxt:html
+	a gist:MimeType ;
+	skos:prefLabel "HTML"^^xsd:string ;
+	gist:uniqueText "text/html"^^xsd:string ;
+	.
+
+mimetxt:plain
+	a gist:MimeType ;
+	skos:prefLabel "Plain Text"^^xsd:string ;
+	gist:uniqueText "text/plain"^^xsd:string ;
+	.
+
+mimetxt:turtle
+	a gist:MimeType ;
+	skos:prefLabel "Turtle"^^xsd:string ;
+	gist:uniqueText "text/turtle"^^xsd:string ;
+	.
+
--- a/gistMediaTypes.ttl
+++ b/gistMediaTypes.ttl
@@ -20,73 +20,73 @@
 	.
 
 media-app:json
-	a gist:MimeType ;
+	a gist:MediaType ;
 	skos:prefLabel "JSON"^^xsd:string ;
 	gist:uniqueText "application/json"^^xsd:string ;
 	.
 
 <https://www.iana.org/assignments/media-types/application/ld+json>
-	a gist:MimeType ;
+	a gist:MediaType ;
 	skos:prefLabel "JSON-LD"^^xsd:string ;
 	gist:uniqueText "application/ld+json"^^xsd:string ;
 	.
 
 media-app:n-quads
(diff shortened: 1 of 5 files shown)


gistMediaTypes.ttl
```diff
@@ -0,0 +1,93 @@
...
+	a owl:Ontology ;
+	owl:imports <https://ontologies.semanticarts.com/o/gistCore> ;
+	owl:versionIRI <https://ontologies.semanticarts.com/o/gistMediaTypesX.x.x> ;
+	skos:definition "Definitions of IANA Media Types."^^xsd:string ;
+	skos:prefLabel "gistMediaTypes"^^xsd:string ;
+	gist:license "https://creativecommons.org/licenses/by-sa/3.0/"^^xsd:string ;
+	.
+
+media-app:json
+	a gist:MimeType ;
```
Jamie-SA, line 23: Should this be gist:MediaType?

sa-bpelakh: Whoops, copy-pasta 🤦 will fix.

mkumba:
I don’t think I’m qualified enough to make a call between mimeType and mediaType (as long as we keep GeneralMediaType, where we just distinguish video, from audio, image and text and one or two others)

And I never heard of iana, is that now the official place for these things.

In terms of adding statements to other peoples URIs I would have thought that as long as we don’t change the meaning we should be ok.  I think I can say that dbpedia:Lourve gist:geoLocatedIn geonames:Paris without violating depedia.  I think if I said dbpedia:Louvre rdf:type gist:Person that would be a bit over the line.

What is the use case for having mimeTypes / mediaTypes?  I was thinking for the few that we natively support (csv, excell, txt, word, ppt, pdf) that we would know how to process them.  Not sure what to do with the several thousand others

On May 17, 2021, at 11:18 AM, Boris Pelakh ***@***.******@***.***>> wrote:


@mkumba<https://github.com/mkumba> There was extensive debate about this issue, and your input was requested. I followed the approach @Jamie-SA<https://github.com/Jamie-SA> and I favored, which was to create annotations for existing IANA media type pages by adding types and labels.

@rjyounes<https://github.com/rjyounes> and some others said we should not be creating statements using entities we do not own as subjects, and suggested I follow this pattern instead:

gist:_MediaType_TextHtml
    a gist:MediaType;
    gist:uniqueText "text/html";
    owl:sameAs iana:text\/html .


Creating a duplicate instance in the gist namespace for each of the media types. Do you have strong feeling in either direction?

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/semanticarts/gist/pull/477

Commit Summary

  *   Rename MimeType to MediaType, create ontology to declare instances.
  *   Clean up labels and namespaces.

File Changes

  *   M bundle.yaml<https://github.com/semanticarts/gist/pull/477/files#diff-73c670418fb192fb77d3a1c95ada8448ee256f1a6afa0e6ede380a7c63221184> (2)
  *   M docs/ReleaseNotes.md<https://github.com/semanticarts/gist/pull/477/files#diff-53d9d97604b51916cd0e08d61a4e86d6357b9a672ee28c624df8aca4ec50647d> (17)
  *   M gistCore.ttl<https://github.com/semanticarts/gist/pull/477/files#diff-39a54dbf6b089d1dd97252ccc000763b9f2d81a08fa49fa8403933b547d23b40> (19)
  *   A gistMediaTypes.ttl<https://github.com/semanticarts/gist/pull/477/files#diff-004dbdee8ec4ae7cad8aaae285da1f7fb245d391782264ff2c2aa0d621cde171> (93)

Patch Links:

  *   https://github.com/semanticarts/gist/pull/477.patch
  *   https://github.com/semanticarts/gist/pull/477.diff

sa-bpelakh:
@mkumba IANA appears to be the master clearing house for registered media types, of which there are MANY. I picked a subset specifically relevant to any application that deals with semantic graphs (all RDF formats, all SPARQL result formats, and a few common text formats, such as `text/plain`), but any gist user would be free to follow the pattern and extend it to whichever media types are relevant to their use case - our unit list is not exhaustive either 😄 . And I absolutely agree that as long as the assertion is reasonable, asserting facts about external entities (such media types) is not wrong.

gistMediaTypes.ttl
sa-bpelakh: Corrected, and expanded `ontologyShapes` to make sure this doesn't happen again. Prior to the fix, build validation reported the following:
```
ERROR:root:SHACL verification produced non-empty results:
Focus                                                                          Path     Value         Severity     Message
<https://www.iana.org/assignments/media-types/application/json>                rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/application/ld+json>             rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/application/n-quads>             rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/application/n-triples>           rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/application/rdf+xml>             rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/application/sparql-results+json> rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/application/sparql-results+xml>  rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/application/trig>                rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/text/csv>                        rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/text/html>                       rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/text/plain>                      rdf:type gist:MimeType sh:Violation Instance references undefined class.
<https://www.iana.org/assignments/media-types/text/turtle>                     rdf:type gist:MimeType sh:Violation Instance references undefined class.
```

Jamie-SA (approved):

gistMediaTypes.ttl
```diff
@@ -0,0 +1,93 @@
...
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix xml: <http://www.w3.org/XML/1998/namespace> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+<https://ontologies.semanticarts.com/o/gistMediaTypes>
+	a owl:Ontology ;
+	owl:imports <https://ontologies.semanticarts.com/o/gistCore> ;
+	owl:versionIRI <https://ontologies.semanticarts.com/o/gistMediaTypesX.x.x> ;
+	skos:definition "Definitions of IANA Media Types."^^xsd:string ;
+	skos:prefLabel "gistMediaTypes"^^xsd:string ;
```
rjyounes, line 18: Our convention dictates that `prefLabel`s are human readable - I.e., in this case, "gist media types" or "gist Media Types" rather than "gistMediaTypes."

ontologyShapes.ttl
```diff
@@ -42,6 +42,39 @@ gshapes:ClassShape
 	] ;
 	.
 
+gshapes:InstanceShape
+	a sh:NodeShape ;
+	rdfs:label "Instance"^^xsd:string ;
```
rjyounes, line 47: Why are we using `rdfs:label` as well as `skos:prefLabel`? 

I would prefer a convention in which we use "Shape" in the label of a shape: I.e., in this case "Instance Shape" rather than "Instance." Thoughts?

ontologyShapes.ttl
```diff
@@ -42,6 +42,39 @@ gshapes:ClassShape
...
+		) ;
+		sh:path a ;
+	] ;
+	sh:severity sh:Violation ;
+	sh:targetSubjectsOf a ;
+	.
+
 gshapes:LowerCase
 	a sh:PropertyShape ;
 	skos:definition "Enforces Lower Case, allowing for all-caps words for acronyms, as well as numbers."^^xsd:string ;
```
rjyounes, line 80: As per convention: "Enforces lower case." That is, "Lower Case" should be lower case.

rjyounes (changes requested):
Requesting minor typographical changes.

gistMediaTypes.ttl
sa-bpelakh: I just followed the pattern of `gistCore`, which has a label of (surprisingly) `gistCore`. Should both be adjusted?

sa-bpelakh: As per issue review discussion, will modify both labels.

rjyounes (approved):

PR #484, opened 2021-05-25
Basis for issue139. Fixes #139.

### LINKED ISSUE #139: Fix definition of basisFor
Current definition: "Reason for an event."

I may want to use this predicate in other circumstances: e.g., a training course is the basis for expertise in a particular area. 

And in fact, the inverse basedOn has a broader definition: "pointer to the thing something was derived from"  This could also use some refinement, but it does allow application to a wider range of resources.

uscholdm:
NOTE:  To avoid merge conflicts, I merged the branch for PR: https://github.com/semanticarts/gist/pull/479 into the branch for this PR. So, before merging this PR it may be best to merge PR: https://github.com/semanticarts/gist/pull/479. 

If there is a better way to do this, I'm all ears.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -6,10 +6,13 @@ Release X.x.x
 
 
 ### Minor Updates
-- Updated annotations for `basedOn` and `basisFor` properties. Issue [#139](https://github.com/semanticarts/gist/issues/139)
 - Added annotation properties which provide usage guidance for properties without the constraints imposed by `rdfs:range` and `rdfs:domain`: Issue [#389](https://github.com/semanticarts/gist/issues/389)
   - domainIncludes
   - rangeIncludes
+
+
+### Patch Updates
+- Updated annotations for `basedOn` and `basisFor` properties. Issue [#139](https://github.com/semanticarts/gist/issues/139)
 
 
 Release 9.7.0
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -5,6 +5,7 @@ Release X.x.x
 -----
 
 ### Minor Updates
+- Updated annotations for `basedOn` and `basisFor` properties.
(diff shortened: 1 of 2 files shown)


rjyounes:
@uscholdm Merging one feature branch into another feature branch effectively means it's impossible to accept one without the other, which defeats the purpose of feature branches. What's supposed to happen is:

- Merge one PR to develop.
- Merge develop back to the feature branch for the second PR.
- Merge the second feature branch into develop. There may be merge conflicts which must be resolved before doing so.

Jamie-SA:
@uscholdm & @rjyounes, you can also do "stacked PRs". This makes sense if one PR is dependent upon another. It means your starting point for the second branch is the first branch (but it hasn't been merged yet). Not sure if that is what MFU was doing but sounds like maybe it was.

uscholdm:
@rjyounes @Jamie-SA I was trying to avoid having to address a conflict when merging independently modified versions of the release notes file.  So it's as if there was a dependency, in that I wanted one PR to merge before the other, but there was no real dependency.  Maybe this not the recommended behavior to avoid a avoid a merge conflict, and we just have to live with it.

Jamie-SA:
@uscholdm Avoiding a merge conflict can be a valid reason to use stacked PRs. However, it comes with the downsides that @rjyounes mentions in that if something holds up the first PR it also holds up the second, or if the first PR is rejected you have change the merge branch and address the un-conflict that you started from a branch that you don't want those changes anymore.

rjyounes:
IMO stacked PRs should only be used in case of dependency, not to avoid merge conflicts of independent PRs. There is a standard process for the latter that I've described above. @uscholdm I'll be happy to step you through that if you want.

uscholdm:
> I'll be happy to step you through that if you want.

I think I got it now.  The punch line seems to be: the way I proposed to avoid merge conflicts has the potential for worse problems than resolving the merge conflicts -- and hence should be avoided.

sa-bpelakh (approved):
Looks good. I updated with `develop` and moved release notes description to `Patch`-level.

Issue #488, opened 2021-05-28
Addition to change and release management documentation for minor vs patch change


rjyounes:
Any addition, even of annotation properties, is considered a minor rather than patch change.

Came up in the context of PR #479 when considering whether this was a minor or patch change.

PR #491, opened 2021-06-08
Clarifications to change and release management doc. Fixes #485, #488.

### LINKED ISSUE #485: Clarify deprecation policy in change and release management doc
Question is: do we always delete in a major release, or do we leave it around until some undefined future time?

New policy: Major releases never include deprecated items.

rjyounes:


rjyounes:
@uscholdm I wasn't sure about this: "For example, changing an equivalent class assertion pointing to a union class to a subclass axiom in order to allow new subclasses to be defined is not a major change." The example was in the previous version but I couldn't tell whether it was an example of a major change or one that wasn't a major change. See what you think.

rjyounes:
@uscholdm I'm confused: did you actually _make_ the changes you are suggesting? The idea is to provide a review rather than an implementation, in case I'm not in full agreement. :)

sa-bpelakh (approved):
Looks good, very clear phrasing.

PR #495, opened 2021-06-24
Refactor Ordered Collections. Fixes #112.

### LINKED ISSUE #112: Address semantic issues related to OrdinalCollection and OrdinalMember
OrdinalMember is not so much a _type_ of thing as the relationship between a thing and an OrdinalCollection. Something can be a member of more than one OrdinalCollection just as it can be a member of more than one Collection in general. For example, 3 is a member of the ordinal collection of integers and of prime numbers.

Granted, the semantics of OrdinalCollection will not be as tightly constrained as they are now without the OrdinalMember class, but they will be more accurate.

This also requires that precedence relationships not be stated directly between the members, but through ordering proxies -- e.g., the [OAI-ORE ontology](http://www.openarchives.org/ore/1.0/vocabulary). That is, in the collection of integers 3 directly precedes 4, but in the collection of prime numbers it directly precedes 5. The relationships don't make sense unless stated relative to a specific collection.

On the other hand, the intent of the OrdinalCollection and OrdinalMember classes may be to limit them to sequences like the months of the year or days of the week, where members are part of _only one_ (well-established) collection. In that case, we must say that the sequences of integers and primes are OrderedCollections but not OrdinalCollections. Then the restriction on OrdinalMember that it can be part of only one OrdinalCollection should be formalized in the OWL definition since it's crucial to the semantics.

daliadahleh:


gistCore.ttl
```diff
@@ -3473,6 +3461,15 @@ gist:identifies
...
 
+gist:isProxyFor
+	a
+		owl:FunctionalProperty ,
+		owl:ObjectProperty
+		;
+	skos:definition "The subject is used as a proxy, or substitute, for the object."^^xsd:string ;
+	skos:prefLabel "is proxy for"^^xsd:string ;
+	.
+
```
uscholdm, line 3472: How about adding the following:
```
	skos:scopeNote "Used to link members of ordered collections to the main item that can appeear in different places in different ordered collections."^^xsd:string ;
```

gistDeprecated.ttl
```diff
@@ -19,6 +19,49 @@
 	owl:versionIRI <https://ontologies.semanticarts.com/o/gistDeprecatedX.x.x> ;
 	.
 
+gist:OrdinalCollection
```
uscholdm, line 22: Add the following to these newly deprecated terms, as in the others.
`owl:deprecated "true"^^xsd:boolean ;`

gistCore.ttl
uscholdm: Also, how about adding an examples file as I did when playing around on my own. See the branch: 
`feature\issue112-MFU-ordered-member-test` in the origin.

See the file `gistExamplesOrderedMembers.ttl` in the `examples` folder. BorisP and I started doing this at Platts. Very helpful, just communicating internally.

uscholdm (changes requested):
A small fix and a couple of suggestions.

gistCore.ttl
daliadahleh: Examples are a good idea. Did we decide on a place for usage examples? I know it's been discussed but I don't know if we've reached a decision.

rjyounes (changes requested):
I'm happy to discuss any of the proposed changes if you would like to.

docs/ReleaseNotes.md
```diff
@@ -8,6 +8,7 @@ Release 10.0.0
 
 - Renamed `MimeType` to `MediaType` to be consistent with [IANA guidelines](https://www.iana.org/assignments/media-types/media-types.xhtml)
   and [RFC6838](https://tools.ietf.org/html/rfc6838). Issue [#434](<https://github.com/semanticarts/gist/issues/434>).
+- Refactored `gist:OrderedCollection` and `gist:OrderedMember`. Added `gist:isProxyFor`. Deprecated `gist:OrdinalCollection` and `gist:OrdinalMember`. Issue [#112](<https://github.com/semanticarts/gist/issues/112>).
```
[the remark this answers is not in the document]
uscholdm: This is a good alternataive.

gistCore.ttl
```diff
@@ -3473,6 +3461,16 @@ gist:identifies
 	skos:prefLabel "identifies"^^xsd:string ;
 	.
 
+gist:isProxyFor
+	a
+		owl:FunctionalProperty ,
+		owl:ObjectProperty
+		;
+	skos:definition "The subject is used as a proxy, or substitute, for the object."^^xsd:string ;
```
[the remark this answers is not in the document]
uscholdm: > Need to state domain Ordered Member.

This reduces the use of isProxyFor for other things.  I prefer not to do that. 

Need to resolve as group before merge PR.

gistCore.ttl
```diff
@@ -3307,7 +3295,7 @@ gist:hasOrderedMember
 		;
 	rdfs:subPropertyOf gist:hasMember ;
 	owl:inverseOf gist:orderedMemberOf ;
-	skos:definition "An inverse functional version of hasMember to ensure that no OrderedMember can be in more than one OrderedCollection., which can quickly lead to problems."^^xsd:string ;
+	skos:definition "An inverse functional version of hasMember to ensure that no OrderedMember can be in more than one OrderedCollection."^^xsd:string ;
```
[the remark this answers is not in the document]
uscholdm: > The range of `hasOrderedMember` should be stated as `OrderedMember`.

This breaks our policy of not including the range in the meaning of a property.  I suggest to say `hasOrderedItem` with no range and then locally have a range in the restriction for the definition of `OrderedCollection`.

rjyounes:
@daliadahleh: @uscholdm and I have been discussing the domain/range issue and we disagree. We think this should be brought to the group before the PR is merged. I will put it on the top of the agenda for the next meeting.

gistCore.ttl
```diff
@@ -1728,51 +1728,39 @@ gist:Offer
 gist:OrderedCollection
 	a owl:Class ;
 	rdfs:subClassOf gist:Collection ;
+	owl:equivalentClass [
+		a owl:Restriction ;
+		owl:onProperty gist:hasOrderedMember ;
+		owl:someValuesFrom gist:OrderedMember ;
+	] ;
```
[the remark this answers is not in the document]
sa-bpelakh: @rjyounes @daliadahleh Should we not remove the `someValuesFrom`? The `allValueFrom` enforces the object type on the property, and removing `someValuesFrom` allows for an empty collection.

gistCore.ttl
```diff
@@ -1728,51 +1728,39 @@ gist:Offer
...
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:sequence ;
+				owl:someValuesFrom xsd:integer ;
+			]
+		) ;
 	] ;
-	skos:definition "A member of an Ordinal Collection.  It necessarily precedes or is preceded by another Ordinal Member in the same collection. (This last condition cannot be formally stated in OWL)."^^xsd:string ;
-	skos:prefLabel "Ordinal Member"^^xsd:string ;
+	skos:definition "A member of an Ordered Collection. It either precedes or is preceded by another Ordered Member in the same collection, or it has a stated integer position in the collection."^^xsd:string ;
```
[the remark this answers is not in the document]
sa-bpelakh: I would use 'linking a member of a collection to that collection' to bypass the 'real-world item' awkwardness.

gistCore.ttl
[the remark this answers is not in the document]
sa-bpelakh: I agree, leave the property general. Add  a `owl:cardinality 1` for `isProxyFor` on `OrderedMember`?

sa-bpelakh (changes requested):
Just a few minor things, @uscholdm and @rjyounes got most of it.

gistCore.ttl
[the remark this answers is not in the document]
daliadahleh: @sa-bpelakh that seems confusing because the class name (and the first sentence of the definition) suggests that the proxy is the 'member', while your definition suggests that the real-world item is the 'member'. I think 'real-world item' is clearer.

gistCore.ttl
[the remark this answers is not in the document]
rjyounes: While I agree with the general principle, I disagree in this particular case. This is a self-contained model, and this predicate is specifically used to link the collection to the proxy. Can you provide examples of other contexts where this predicate would be used _with the same meaning_?

gistCore.ttl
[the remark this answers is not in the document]
rjyounes: Same response as above: While I agree with the general principle, I disagree in this particular case. This is a self-contained model, and this predicate is specifically used to link the ordered member  to the thing it's a proxy for. Can you provide examples of other contexts where this predicate would be used _with the same meaning_? We don't really want people using it for, say, legal proxies, do we? That's not the same thing at all, despite using the same word "proxy." I don't think this principle of not overspecifying usage should lead to predicates that don't have a particular meaning at all.

gistCore.ttl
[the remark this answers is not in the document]
rjyounes: @uschold and I discussed the possibility of empty sets and opened a new issue to raise the question: https://github.com/semanticarts/gist/issues/498. In theory I'm in favor of allowing empty collections, but we noted that if we allow empty collections, then the restriction must also be removed from `Collection` (`gist:hasMember some owl:Thing`), which leaves it as a class without a formal definition.

gistCore.ttl
```diff
@@ -1729,12 +1729,18 @@ gist:OrderedCollection
...
 	owl:equivalentClass [
-		a owl:Restriction ;
-		owl:onProperty gist:hasOrderedMember ;
-		owl:someValuesFrom gist:OrderedMember ;
+		owl:intersectionOf (
+			gist:Collection
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:hasOrderedMember ;
+				owl:someValuesFrom gist:OrderedMember ;
```
rjyounes, line 1737: This is a bit nuanced: the `someValuesFrom` statement makes it appear that we are restricting the range for this class, which we are not, since the range is already defined on the predicate. There are two ways to do indicate the cardinality in a less misleading way:

`owl:someValuesFrom owl:Thing`

and
`owl:minCardinality "1"^^xsd:nonNegativeInteger`

I prefer the second way as more straightforward and find the first odd. @uscholdm  prefers the first way because he prefers `someValuesFrom` to `minCardinality` for reasons that I don't fully agree with. I'd say it's your choice. Maybe @uscholdm will have something to say here.

gistCore.ttl
```diff
@@ -1729,12 +1729,18 @@ gist:OrderedCollection
...
+			]
+		) ;
 	] ;
 	skos:definition "A collection where the members are in a fixed sequence."^^xsd:string ;
 	skos:prefLabel "Ordered Collection"^^xsd:string ;
+	skos:scopeNote "Empty collections are not supported at the time."^^xsd:string ;
```
rjyounes, line 1743: "at this [not the] time"

rjyounes (changes requested):
Fix one typo and consider another change to the `someValuesFrom` restriction.

gistCore.ttl
daliadahleh: I'll go with the second. When I see owl:Thing I assume it specifically shouldn't be restricted to one class.

sa-bpelakh (approved):
Looks good, thanks for steering this barge into the port 👏

gistCore.ttl
```diff
@@ -1728,51 +1728,45 @@ gist:Offer
...
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:sequence ;
+				owl:someValuesFrom xsd:integer ;
+			]
+		) ;
 	] ;
-	skos:definition "A member of an Ordinal Collection.  It necessarily precedes or is preceded by another Ordinal Member in the same collection. (This last condition cannot be formally stated in OWL)."^^xsd:string ;
-	skos:prefLabel "Ordinal Member"^^xsd:string ;
+	skos:definition "A member of an ordered collection. This resource serves as a proxy linking a real-world item in the collection to that collection, allowing the expression of positional and other information that applies to the item only in the context of the collection."^^xsd:string ;
```
uscholdm, line 1768: > linking a real-world item in the collection to that collection,

Confusing wording, it does not seem to be saying what we want it say. Why does an item that is already in a collection need to be linked to that same collection?

uscholdm (changes requested):

gistCore.ttl
daliadahleh: How about 'This resource serves as a proxy that represents an item in the collection to allow the expression of positional and other information that applies to the item only in the context of the collection'?

rjyounes: I still think the phrase "an item in the collection" causes confusion, since it's technically only the proxy that's in the collection. 
How about "A member of an ordered collection. This resource links an item to a collection and allows the expression of positional and other information that applies to the item only in the context of the collection."?

uscholdm: This is annoyingly difficult to do well. Here is one more attempt. The example is from the IMF.

```
skos:definition "A member of an ordered collection. It is a proxy for a real world item that can appear in different orders in different collections. The ordered member appears in exactly one ordered collection."

skos:scopeNote "An ordered member points to the real world item via the providesOrderFor property and has ordering information either as a number in a sequence, or by preceding or following another ordered member."^^xsd:string ;  

skos:example "A document template may have parts such as table of contents, introduction, main body, appendix etc, but not all documents have all parts.  The appendix might be the third part of one document template, but the 5th part of another. This would correspond to one instance of DocumentPart for an appendix, and two instances of ordered member."^^xsd:string ; 
```

uscholdm (changes requested):
The fact that defining OrderedMember is so difficult to do well, is all the more reason to get it as good as we can.

rjyounes:
I think Michael has mostly nailed it. I propose a couple of tweaks:

```
skos:definition "A member of an ordered collection that serves as a proxy for a real world item, which can appear in different orders in different collections. The ordered member appears in exactly one ordered collection."

skos:scopeNote "An ordered member points to the real world item via the providesOrderFor property and represents ordering information either as a number in a sequence, or by preceding or following another ordered member.  The ordered collection is linked to the ordered member via the property hasOrderedMember. The ordered member may also represent data about the real world item that applies to it only in the context of the ordered collection."

skos:example "A person may rank 12th in the Boston Marathon but 29th in the New York City Marathon."
```

Michael, I've provided a different example which I think is more straightforward and easier to understand. See what you think. We could of course include both.

uscholdm:
Yes, much better example. Below is a slightly revised version. I split a long sentence in the scope note into two. I also got rid of the last sentence. First, I found it confusing. When I realized what it says, it kind of goes without saying, once the idea is grasped, which is hard enough. If we insist on saying it, then we need to have a compelling example that somehow does not add more complexity and confusion.  Even if it was a good example, it still may not be worth saying.  Less is more.  See also the variation at the end.
```
skos:definition "A member of an ordered collection that serves as a proxy for a real world item, which can appear in different orders in different collections. The ordered member appears in exactly one ordered collection."

skos:scopeNote "An ordered member points to the real world item via the providesOrderFor property.  Ordering information is represented either as a number in a sequence, or by preceding or following another ordered member.  The ordered collection is linked to the ordered member via the property hasOrderedMember."

skos:example "A person may rank 12th in the Boston Marathon but 29th in the New York City Marathon."
```

**VARIATION**: I moved last sentence of definition to first sentence of scope note and replaced it with a new sentence, which I think captures the essence.  The scope note is rewriten to exactly echo the last sentence in the definition.

```
skos:definition "A member of an ordered collection that serves as a proxy for a real world item, which can appear in different orders in different collections. An ordered member exactly corresponds to a particular item in a particular order in a particular (ordered) collection."

skos:scopeNote "An ordered member points to the real world item via the providesOrderFor property. It is linked to exactly one ordered collection via the hasOrderedMember property.    Ordering information is represented either as a number in a sequence, or by preceding or following another ordered member."

skos:example "A person may rank 12th in the Boston Marathon but 29th in the New York City Marathon."
```

_Why was this so fricking hard to do well?_

rjyounes (approved):

uscholdm:
I saw that Dalia made changes to the annotations, but they were not the latest draft.  Can @rjyounes and @daliadahleh have a look to see what we want to go with?

daliadahleh:
@uscholdm, the annotations are from your latest draft. I used the first variation in [your latest comment ](https://github.com/semanticarts/gist/pull/495#issuecomment-876008994).

rjyounes:
I'm happy to go with this. @uscholdm please give the final go-ahead.

uscholdm:
> the annotations are from your latest draft. I used the first variation

I prefer the second variation. Can you say what you prefer about the first?

daliadahleh:
I thought 'An ordered member exactly corresponds to a particular item in a particular order in a particular (ordered) collection' is unnecessarily wordy. The point comes across just as clearly in the first variation. I'm not too picky though, happy to go with either.

uscholdm:
> I thought 'An ordered member exactly corresponds to a particular item in a particular order in a particular (ordered) collection' is unnecessarily wordy.

I think I agree, at least for the definition. We could just remove that sentence all together, or put it in the scope note.

Its Rebecca's call. I'm not too bothered either way.

rjyounes:
Let me look at it. It's not actually correct to say it corresponds to a particular item "in a particular order." It is linked to the same item regardless of the order. It will just change its properties (sequence or directlyPrecedes) accordingly.

uscholdm:
OK - maybe its just helpful to me as a way to think about it, but confusing otherwise.  Probably best to just leave that verbiage out.

Issue #506, opened 2021-07-08
Remove inverses unless there's a compelling reason


rjyounes:
As per the decision in #294, we should purge gist of inverse properties that don't have a compelling reason to exist. In general, the guidance is that if there's a tree-like structure, keep the property that goes up rather than down the tree, since cardinality generally will be greater going downward. If there's no tree-like structure, it's dependent on individual context.

This will be a minor change if we deprecate the inverses. 

[Added]
From Issue #551 
`precedesDirectly` has inverse `followsDirectly`, but `precedes` has no inverse `follows`. Should be consistent one way or another.

rjyounes:
We currently define 17 inverses. 

We have to decide which ones to keep. @uscholdm and @mdfeickert will take a pass at a proposal and present.

Rebecca: use directionality guidelines in style guide.

Peter: If we don't provide inverses, people will be rolling their own. We should put recommendations in place for when it's appropriate. 

Peter: could also have an optional import of the inverses.

Michael: or add an annotation with suggestions of how to name the inverses.

Rebecca: Why do we care what other people name properties in their own namespaces?

Rebecca: We don't need both the annotation and the import. The latter seems better.

Borislav: Remove all inverses from gist and put in a separate file.

Michael: There are some inverses I use all the time that I would like to keep, but I wouldn't want to use the whole file. So each person would have to pick and choose the ones they want. This is inconvenient, and argues for keeping the commonly used ones in gistCore.

Michael: May also propose bringing a few back.

rjyounes:
Consider in conjunction with #551.

JonathonGist:
Glad for these decisions--especially that all of them are to be available in an imports file. And I'm glad that the special ones, used often, are being preserved in gistCore.

JonathonGist:
It seems that with the availability of inverse paths in SPARQL and SHACL, there is limited or no need for adding inverse predicates to an ontology. My argument for keeping them in a separate file has evaporated.

rjyounes:
I've now come across a case or two where the inverse appears to be necessary. For example, if we have a `gist:ControlledVocabulary`, we probably want to assert `<term> gist:isMemberOf <term>` rather than using a value restriction on `gist:hasMember` on the vocabulary itself. This results in duplication and therefore maintenance problems.

Having thought of this case, I now think we have to proceed with some caution, though at the moment I haven't thought of any others.

rjyounes:
@marksem will provide a candidate list of removals and run it by the group.

[...]

Issue #444, opened 2021-03-11
parentOf property: definition needs work


uscholdm:
The current definition for `gist:parentOf` is: 
```
gist:parentOf
	a owl:ObjectProperty ;
	skos:definition "Biological Parent"^^xsd:string ;
	skos:prefLabel "parent of"^^xsd:string ;
	.
```

The term parent is used much more broadly in a hierarchy. For example:
- directSubTaskOf is defined to be an immedate parent task
-  we have a property called `gist:hasNavigationalParent`

SUGGESTION: rename to `biologicalParentOf` (or use the inverse `hasBiologicalParent`

rjyounes:
Not just that, but more to the point, there are adoptive parents who have legal responsibility for the child. I don't know if step-parent would ever be important in an enterprise context without legal adoption.

Anyway, I heartily agree with the suggestion.

Jamie-SA:
Related to the discussion is gist:offspringOf which sounds like it would be the inverse of the the current gist:parentOf (and the proposed gist:biologicalParentOf). Neither existing properties have owl:inverseOf declared.

Not relevant to the proposal but I am curious how often gist:parentOf got used in a non-biological hierarchies. :)

rjyounes:
This is subsumed by #534 but I'm leaving it open so we make sure to consider the specific issues addressed here.

uscholdm:
@rjyounes This has been addressed, suggest to close. 
Here is the current implementation:
```
gist:hasBiologicalParent
	a owl:ObjectProperty ;
	rdfs:domain gist:LivingThing ;
	rdfs:range gist:LivingThing ;
	skos:definition "Relates a living thing to its biological parent."^^xsd:string ;
	skos:prefLabel "has biological parent"^^xsd:string ;
	.
```

PR #535, opened 2021-08-17
Revise local names according to adopted conventions. Fixes #188

### LINKED ISSUE #188: Define and implement conventions for local names

rjyounes:


rjyounes:
In the interests of getting this done and the `10.0.0` release out the door, I suggest that reviewers flag only errors, and changes they cannot live with. I have kept the changes conservative to minimize disagreement, even in cases where I felt a fuller rewording would be an improvement. We will have the opportunity to make changes in future major releases. In particular, we will likely release a revamp of times, in which case all or many of the predicates will change and names can be re-evaluated then. In addition, we will probably be removing a lot of inverses (#506). 

I added 'biological' to the parent/offspring predicates to align with the definitions, but there is an issue about whether this limitation is appropriate (#444). I also reversed the order to hasX vs isXOf.

rjyounes:
@daliadahleh, because @uscholdm is off for a while, I'm requesting a review from you. As I mentioned to the other reviewers, unless you find something egregious or that you can't live with, or have a minor comment that can be easily fixed, let's try to pass this one through. It's holding up the release and we have been working on this for 18 months! Thanks!

marksem (changes requested):
Shouldn't `offsetToUniversal` have been changed to something like `hasOffsetToUniversal` ?

`precedesDirectly' appears to be missing annotations. This is because it is not declared anywhere, but only referenced as the inverse of `followsDirectly`.

Why is `latitude` an object property as well as a data property when I view in protege? Ah, b/c its range is xsd:double, and all the datatypes are classes now, at least according to Protege.  Same issue for longitude, sequence, baseConversionFactor, and more.

rjyounes:
I just noticed that all the datatypes are classes in Protege. That's an error we need to track down. They're also shown as datatypes. It's possibly a Protege bug. Do you have time to look into that, since you know Protege better than I do?

You're right about `offsetToUniversal`. And I'll check on `precedesDirectly`. Good catches!

rjyounes:
Changed target branch from `develop` to `release-`10.0.0`.  This PR was created before the latter was split off from the former; it now needs to be merged into the release branch, and will be merged to `develop` along with all other commits on the release branch.

marksem:
@rjyounes , yes I can investigate to see if Protege is doing something wrong, or if it is a logical entailment somehow.

rjyounes:
Fixed `offsetToUniversal` and restored `precedesDirectly`.

rjyounes:
@marksem The ontology changes we discussed are ready for re-review. Then we're awaiting your investigation of the Protege issue.

rjyounes:
@marksem @sa-bpelakh  I now recall that I've seen this problem before, and it's a protege bug.  The relevant OWL is:
```
gist:numericValue
	a owl:DatatypeProperty ;
	rdfs:domain gist:Magnitude ;
	rdfs:range [
		owl:unionOf (
			owl:rational
			owl:real
			xsd:byte
			xsd:decimal
			xsd:double
			xsd:float
			xsd:int
			xsd:integer
			xsd:long
			xsd:negativeInteger
			xsd:nonNegativeInteger
			xsd:nonPositiveInteger
			xsd:positiveInteger
			xsd:short
			xsd:unsignedByte
			xsd:unsignedInt
			xsd:unsignedLong
			xsd:unsignedShort
		) ;
	] ;
        .
```
This is valid OWL because the range of owl:unionOf is rdf:List. Protege doesn't handle it properly. I've checked out TBC and it has no problem with it. This is very unfortunate because many of our users will view the ontology in Protege. Thoughts? @sa-bpelakh We could add a note to the property, at least.

sa-bpelakh:
> @marksem @sa-bpelakh I now recall that I've seen this problem before, and it's a protege bug. The relevant OWL is:
> 
> ```

Should we remove `rdfs:range` and replace it w/ `gist:rangeIncludes`, so we document it but not confuse Protege? How important is this correction?

marksem:
Apparently Protege is ok with something like this:
`<http://ns/myprop1> rdf:type owl:DatatypeProperty ;
                    rdfs:range [ rdf:type rdfs:Datatype ;
                                 owl:unionOf ( owl:rational
                                               owl:real
                                             )
                               ] .

`

So maybe we're just missing the `rdf:type rdfs:Datatype` part and then we're all good.  Did not get time to fully test yet.

rjyounes:
@marksem I've moved the Protege issue over to #552. I assigned it to myself, but feel free to reassign it to yourself if you have a chance to test it.

rjyounes:
The release note might be a bit confusing: I included a note about updates to the gist style guide, although those are not in this PR, in order to avoid one merge conflict.

marksem (approved):

PR #546, opened 2021-09-02
Remove redundant subclass assertion

rjyounes:


sa-bpelakh (approved):
👍

sa-bpelakh:
If this is going into release10.0, it should move into that project, rather than being in October 2021. Otherwise, merge into develop instead.

rjyounes:
Yes, I got the wrong project. It's 10.0.0.

PR #553, opened 2021-09-03
Fix numericValue range declaration (Fixes #552)

### LINKED ISSUE #552: Protege doesn't handle a unionOf datatypes in gist:numericValue correctly
The definition of `gist:numericValue` is:

```
gist:numericValue
	a owl:DatatypeProperty ;
	rdfs:domain gist:Magnitude ;
	rdfs:range [
		owl:unionOf (
			owl:rational
			owl:real
			xsd:byte
			xsd:decimal
			xsd:double
			xsd:float
			xsd:int
			xsd:integer
			xsd:long
			xsd:negativeInteger
			xsd:nonNegativeInteger
			xsd:nonPositiveInteger
			xsd:positiveInteger
			xsd:short
			xsd:unsignedByte
			xsd:unsignedInt
			xsd:unsignedLong
			xsd:unsignedShort
		) ;
	] ;
	skos:definition "The actual value of a magnitude."^^xsd:string ;
	skos:prefLabel "numeric value"^^xsd:string ;
	.
```

This is valid OWL (the range of `owl:unionOf` is `rdf:List`). However, Protege doesn't recognize it and:
* Considers all the datatypes as classes as well as datatypes
* Considers the property as both a datatype and object property

This is unfortunate because many of our users view the ontology in Protege. The only stopgap I can think of is to add a `skos:note` to the property definition explaining the issue. Is there any other approach? Is there any way to change the axiom in a way that Protege accepts?

marksem:


sa-bpelakh (approved):
I am assuming you verified that this fixes Protege. I had to do a small patch to fix validation.

marksem:
@sa-bpelakh yes, it fixes things in Protege.  Thanks for fixing the validation.  The auto-validate is new to me.  I guess I haven't edited gist in a while. :o

PR #554, opened 2021-09-05
Local and endpoint migration for renamed and removed entities.

sa-bpelakh:
This is in anticipation of #535, based on the most recent version.

In order to run these, you will need to make sure that [onto-tool](https://pypi.org/project/onto-tool/) is upgraded to v1.4.1.

I ran the local migration using the following command:
```bash
onto_tool bundle -v input test-migration -v output test-migration/migrated -v report test-migration/report migration/v10.0/migrate_local.yaml
```
which updates all `.ttl` and `.owl` files in the `input` directory (not recursively, is that an issue?) and places the updated RDF in the `output` directory. ValidationReport RDF is placed in the `report` directory.

To run the migration against a SPARQL endpoint:
```bash
onto_tool bundle -v user USER -v password PWD -v endpoint https://agraph.semanticarts.com/catalogs/demo-catalog/repositories/ci-test -v report test-migration/endpoint-report migration/v10.0/migrate_endpoint.yaml
```

For databases where the UPDATE endpoint is different, it can be provided via `-v update_endpoint URI` in addition to the query endpoint provided via `-v endpoint`.

Do we want to include detailed documentation for the migration in the release notes of each major release, since users will not have access to our Confluence pages?


migration/v10.0/action_rename_classes_default.rq
```diff
@@ -0,0 +1,23 @@
...
+        (gist:MimeType gist:MediaType)
+        # Issue #483
+        (gist:BuildingAddress gist:StreetAddress)
+    }
+    {
+        ?ref ?refProp ?oldClass
+    } UNION {
+        ?oldClass ?defProp ?defObj
+    }
+}
```
rjyounes, line 23: Why do we have both `action_rename_classes.rq` and `action_rename_classes_default.rq` and similarly for properties?

migration/v10.0/detect_removed.rq
```diff
@@ -0,0 +1,67 @@
...
+            ] .
+}
+WHERE {
+    {
+        select distinct ?deleted where {
+            values ?deleted {
+                gist:PhysicalThing
+                gist:Room
+                gist:SocialBeing
+                gist:TemmplateTask
```
rjyounes, line 26: Typo

migration/v10.0/detect_removed.rq
```diff
@@ -0,0 +1,67 @@
...
+                gist:_unitedNations
+                gist:date
+                gist:dateTime
+                gist:geoDirectlyContainedIn
+                gist:geoDirectlyContains
+                gist:getter
+                gist:giver
+                gist:hasA
+                gist:localDate
+                gist:hasA
```
rjyounes, line 38: duplicate

migration/v10.0/action_rename_classes_default.rq
sa-bpelakh: We do renaming in two steps:
1. First rename everything that is in a named graph, retaining the modified triples in their original graphs, and then
2. Rename all that is remaining, which would be in the default graph
We do it in this manner because the default graph does not have a universally recognized reference URI, and in some triple stores it doesn't have a URI at all. Also, when processing local file data, there are no named graphs, so only the default queries (which don't specify a graph) are used.

migration/v10.0/detect_removed.rq
sa-bpelakh: Done

migration/v10.0/detect_removed.rq
sa-bpelakh: Good catch

migration/v10.0/action_rename_classes_default.rq
Jamie-SA: A local .trig or N-quads file can have named graphs. I assume that works correctly for them as well.

sa-bpelakh: @Jamie-SA There is no current support in onto-tool for .trig or .nq inputs. Do you view this as a deal breaker for this functionality? There is no way I can code and test this before Monday

Jamie-SA: It is not a deal breaker for me, I had assumed the library you use to read/write would support all of the main formats. But if not, I wouldn't hold this up because of it.

I have switched almost entirely to .trig because of it's support for named graphs.

sa-bpelakh: The toolkit (`rdflib`) does support it, but I have not enabled it - and there is an existing issue (https://github.com/semanticarts/ontology-toolkit/issues/65), to which I will give some attention after this release.

migration/v10.0/detect_removed.rq
```diff
@@ -0,0 +1,65 @@
...
+            {
+                ?s ?deleted ?o
+            } UNION {
+                ?s ?p ?deleted
+            } UNION {
+                ?deleted ?p ?o
+            }
+        }
+    }
+    bind(<urn:deleted-entity-vaidation-report> as ?report)
```
Jamie-SA, line 62: Minor typo: 
```suggestion
    bind(<urn:deleted-entity-validation-report> as ?report)
```

sa-bpelakh:
@rjyounes Please read over my documentation. @bolerio Should be good to go as soon as @rjyounes approves the verbiage.

docs/MajorVersionMigration.md
```diff
@@ -0,0 +1,49 @@
...
+## Migrating RDF Data in Local Files
+
+Once you have the Ontology Toolkit installed, issue the following command from the directory where
+you cloned [gist](https://github.com/semanticarts/gist):
+```shell
+onto_tool bundle -v input INPUT-DIR 
+                 -v output OUTPUT-DIR
+                 -v report REPORT-DIR migration/v10.0/migrate_local.yaml
+```
+where _INPUT-DIR_ is the directory in which your RDF data is located, _OUTPUT-DIR_ is the directory where
```
rjyounes, line 29: Can you indicate whether the directories have to exist in advance, or will they be created if they don't exist? Or is that too obvious to state?

rjyounes (changes requested):
Looks great except for one or two suggested rewordings in the documentation. Use at your own discretion.

bolerio:
Thanks @sa-bpelakh  and @rjyounes .

@rjyounes , waiting on your final approval to merge and seal the release.

sa-bpelakh:
@rjyounes Adjusted documentation as per your suggestions

rjyounes (approved):

PR #561, opened 2021-09-11
Validation query for hasCommunicationAddress

Jamie-SA:
I was working on queries to validate the new restrictions on various changes made in this release but after working on them a bit I realized that these types of checks would be better to do in SHACL. If OntoTool can do SHACL validation as part of the migration process I could create some shapes to enforce the new restrictions. Shapes could also be added later.

This PR has a couple of queries that can be included.


sa-bpelakh (changes requested):
onto-tool can handle SHACL on local execution (see https://github.com/semanticarts/gist/blob/develop/bundle.yaml#L59), but since there is no uniform way to invoke SHACL across different triple stores, there is no support for endpoint SHACL validation. For that reason I have gone with SPARQL for the migration, so that the queries can be reused.

Good catch not assuming that triple store queries the union of all named graphs, I need to add a version of my detect query that looks inside graphs.

I would simplify the detection query as follows:
```
CONSTRUCT {
    ?report a sh:ValidationReport ;
        sh:conforms false ;
        sh:result
            [
                a sh:ValidationResult ;
                sh:focusNode ?failedEntity ;
                sh:resultMessage "Domain restriction on gist:hasCommunicationAddress violated" ;
                sh:resultSeverity sh:Warning ;
                sh:sourceConstraintComponent <urn:constraint:communicatonAddress-domain>
            ] .
}
WHERE {
    { ?failedEntity gist:hasCommunicationAddress ?addr . }
    FILTER NOT EXISTS {
        { ?failedEntity a gist:Person } UNION { ?failedEntity a gist:Organization }
    }
    bind(<urn:new-domain-validation-report> as ?report)
}
```
There is no reason to include the URI in the error message, since it's automatically included in the report as the `focusNode`.

Since the detection queries are run via the following section of the script, if you prefix the file names with `detect_`, you won't need to make any additional changes.
```yaml
- action: "verify"
  message: "Find remaining problems"
  type: "construct"
  inference: "none"
  stopOnFail: false
  target: '{report}'
  endpoint:
    query_uri: "{endpoint}"
    user: "{user}"
    password: "{password}"
  queries:
    source: '{migration}/v10.0'
    includes:
      - 'detect_*.rq'
```

sa-bpelakh (approved):

Issue #579, opened 2021-09-27
gist unit of measure instances have no unit symbols


rjyounes:
Submit one PR for #370, #526, and #579.

uscholdm:
Do we want the full meal deal, as in: 
```
gist:_kilogram_per_cubic_meter
	a
		owl:Thing ,
		gist:DensityUnit
		;
	skos:prefLabel "Kilogram per cubic meter"^^xsd:string ;
	gist:convertToStandard "1.0"^^xsd:double ;
	gist:denominator gist:_cubic_meter ;
	gist:hasStandardUnit gist:_kilogram_per_cubic_meter ;
	gist:numerator gist:_kilogram ;
	gist:unitSymbol "kg/m^3"^^xsd:string ;
	gist:unitSymbolHTML "kg/m&sup3;"^^xsd:string ;
	gist:unitSymbolUnicode "kg/m?"^^xsd:string ;
	.
```
The difference is nicely showing sub/superscripts for exponents, or showing the small circle degrees symbol for temperature.  Come to think of it, it virtually all the gist units will be simple, no need for any special characters.

rjyounes:
I vote no.

rjyounes:
Would `gist:unitSymbolHtml` and `gist:unitSymbolUnicode` be good candidates for removal?

uscholdm:
More a case of not adding them.  I don't think they exist in gist now, just in special add-on units ontologies for various clients.  

I think we should not add them at this time.  Maybe an editorial comment on `gist:UnitOfMeasure` to suggest doing this.

Issue #526, opened 2021-08-04
Add definitions to unit of measure instances


rjyounes:
They should have definitions, and they are also failing onto-tool validation.

rjyounes:
Implement with #556 and #370.

rjyounes:
Submit one PR for #370, #526, and #556.

rjyounes:
#370 and #556 will go into a separate PR.

Issue #146, opened 2020-01-16
Clarify definition of CoherentRatioUnit or remove the class


rjyounes:
Current definition: "A ratio unit whose numerator and denominator reduce to 1."

Maybe I'm just ignorant here, but this doesn't mean anything to me. We should clarify it for the general population.

uscholdm:
Yeah, it requires more explanation and might not be technically accurate. It means that none of the units need any conversions, after simplification.  

See: http://www.eim.gr/metrology/measurement-units/si-derived-units/coherent-derived-units/ which says coherent derived units are:
> products of powers of base units. Coherent derived units are products of powers of base units that include no numerical factor other than 1.

Examples:
- meter / second:  both meter and second have conversion to base of 1
- km / killo-second has  1000 conversion for numerator, and 1000 for denominator which reduces to 1. It is equiv to meter / sec
- km/sec.   km has conversion of 1000, sec has conversion of 1, so this does not reduce to 1

sa-bpelakh:
DECISION: add `skos:example` (see above) that clarifies meaning of definition.

rjyounes:
This is subsumed by #534 but I'm leaving it open so we make sure to consider the specific issues addressed here.

rjyounes:
See discussion in #331.

stevenchalem:
On hold until major measurement changes are implemented

uscholdm:
This should be moot, as ProductUnit and RationUnit are ambigous and likely going away.

uscholdm:
Defer until units work has been decided.

philblackwood:
Will be addressed in new Units of Measure (concept not needed).

PR #586, opened 2021-09-29
Issues 370 526 579 instances

uscholdm:
- Updated annotations for coherent unit classes, per issue [#146]
- Added definitions for unit of measure instances per issue [#526] 
- Added unit symbols for unit instances instances per issue [#579]

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -3,7 +3,9 @@ gist Release Notes
 
 ### Minor Updates
 
-- update annotations for coherent unit classes, per issue [#146](https://github.com/semanticarts/gist/issues/146)
+- Updated annotations for coherent unit classes, per issue [#146](https://github.com/semanticarts/gist/issues/146)
+- Added definitions for unit of measure instances per issue [#526] https://github.com/semanticarts/gist/issues/526 
+- Added unit symbols for unit instances instances per issue [#579] https://github.com/semanticarts/gist/issues/579 
 
 Release 10.0.0
 -----
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,11 +1,14 @@
 gist Release Notes
 =====
 
+### Minor Updates
+
+- Added unit symbols for unit instances instances per issue [#579] (https://github.com/semanticarts/gist/issues/579) 
+
 ### Patch Updates
 
-- Updated annotations for coherent unit classes, per issue [#146](https://github.com/semanticarts/gist/issues/146)
-- Added definitions for unit of measure instances per issue [#526] https://github.com/semanticarts/gist/issues/526 
-- Added unit symbols for unit instances instances per issue [#579] https://github.com/semanticarts/gist/issues/579 
+- Added labels to gist instances, per issue [#370](https://github.com/semanticarts/gist/issues/370)
+- Added definitions for unit of measure instances per issue [#526] (https://github.com/semanticarts/gist/issues/526) 
 
 Release 10.0.0
 -----
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,7 +1,7 @@
 gist Release Notes
 =====
 
-### Minor Updates
+### Patch Updates
 
 - Updated annotations for coherent unit classes, per issue [#146](https://github.com/semanticarts/gist/issues/146)
 - Added definitions for unit of measure instances per issue [#526] https://github.com/semanticarts/gist/issues/526 
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -2637,6 +2637,8 @@ gist:_USDollar
 		owl:Thing ,
 		gist:BaseUnit
 		;
+	skos:definition "The base unit for currency."^^xsd:string ;
+	skos:prefLabel "US Dollar"^^xsd:string ;
 	gist:baseConversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_USDollar ;
 	.
@@ -2647,6 +2649,8 @@ gist:_ampere
 		owl:Thing ,
 		gist:BaseUnit
 		;
+	skos:definition "The base unit for electrical current."^^xsd:string ;
+	skos:prefLabel "ampere"^^xsd:string ;
 	gist:baseConversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_ampere ;
 	.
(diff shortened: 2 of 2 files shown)


docs/ReleaseNotes.md
```diff
@@ -1,6 +1,12 @@
 gist Release Notes
 =====
 
+### Patch Updates
+
+- Updated annotations for coherent unit classes, per issue [#146](https://github.com/semanticarts/gist/issues/146)
+- Added definitions for unit of measure instances per issue [#526] https://github.com/semanticarts/gist/issues/526
```
rjyounes, line 7: Need parens around the URL

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,12 @@
 gist Release Notes
 =====
 
+### Patch Updates
+
+- Updated annotations for coherent unit classes, per issue [#146](https://github.com/semanticarts/gist/issues/146)
+- Added definitions for unit of measure instances per issue [#526] https://github.com/semanticarts/gist/issues/526 
+- Added unit symbols for unit instances instances per issue [#579] https://github.com/semanticarts/gist/issues/579
```
rjyounes, line 8: Need parens around the URL

gistCore.ttl
```diff
@@ -324,7 +326,8 @@ gist:CoherentUnit
 	skos:prefLabel "Coherent Unit"^^xsd:string ;
 	skos:scopeNote
 		"Coherent unit is the physics term for this, informally you might think of it as the standard unit for a given dimension."^^xsd:string ,
-		"In principle, the CoherentUnit for a ProductUnit or RatioUnit can be inferred by recursively decomposing the products and ratios into their respective CoherentUnits, bottoming out in SimpleUnits"^^xsd:string
+		"In principle, the CoherentUnit for a ProductUnit or RatioUnit can be inferred by recursively decomposing the products and ratios into their respective CoherentUnits, bottoming out in SimpleUnits"^^xsd:string ,
+		"See: http://www.eim.gr/metrology/measurement-units/si-derived-units/coherent-derived-units/"^^xsd:string
```
rjyounes, line 330: How about putting the "See" statement in an `rdfs:seeAlso`?

gistCore.ttl
```diff
@@ -2654,7 +2663,9 @@ gist:_bit
 		owl:Thing ,
 		gist:BaseUnit
 		;
-	skos:definition "A bit (short for binary digit) is the smallest unit of data in a computer. A bit has a single binary value, either 0 or 1."^^xsd:string ;
+	skos:definition "The base unit for measuring information."^^xsd:string ;
```
rjyounes, line 2666: I would say more specifically "digital information."

gistCore.ttl
```diff
@@ -2702,9 +2722,12 @@ gist:_kelvin
 		owl:Thing ,
 		gist:BaseUnit
 		;
+	skos:definition "The base unit for measuring temperature."^^xsd:string ;
+	skos:prefLabel "Kelvin"^^xsd:string ;
```
rjyounes, line 2726: Should the prefLabel here be lowercased, like the others?

gistCore.ttl
```diff
@@ -2761,6 +2799,9 @@ gist:_one_day
 		owl:Thing ,
 		gist:Duration
 		;
+	skos:definition "A duration equal to 24 hours."^^xsd:string ;
+	skos:prefLabel "one day"^^xsd:string ;
+	skos:scopeNote "This is a gist:Magnutude, not a unit. Semantically it is the same as the unit, day."^^xsd:string ;
```
rjyounes, line 2804: Cheeky. Though you know I agree, I wouldn't assert semantic equivalence until the team has decided to collapse the magnitudes and units generally.

gistCore.ttl
```diff
@@ -2771,6 +2812,9 @@ gist:_one_millisecond
 		owl:Thing ,
 		gist:Duration
 		;
+	skos:definition "A duration equal to one millisecond."^^xsd:string ;
+	skos:prefLabel "one millisecond"^^xsd:string ;
+	skos:scopeNote "This is a gist:Magnutude, not a unit. Semantically it is the same as the unit, millisecond."^^xsd:string ;
```
rjyounes, line 2817: Same as above

gistCore.ttl
```diff
@@ -2781,6 +2825,9 @@ gist:_one_minute
 		owl:Thing ,
 		gist:Duration
 		;
+	skos:definition "A duration equal to one minute."^^xsd:string ;
+	skos:prefLabel "one minute"^^xsd:string ;
+	skos:scopeNote "This is a gist:Magnutude, not a unit. Semantically it is the same as the unit, minute."^^xsd:string ;
```
rjyounes, line 2830: Same as above

gistCore.ttl
```diff
@@ -2791,8 +2838,11 @@ gist:_second
 		owl:Thing ,
```
rjyounes, line 2838: Why do we have these `owl:Thing` assertions everywhere? Does the serializer add them, or are they leftover. I recommend removing them.

rjyounes (changes requested):
Looks good overall. I've made a few suggestions.

docs/ReleaseNotes.md
uscholdm: ok

gistCore.ttl
uscholdm: Not a bad idea. There is exactly one other use of `seeAlso`, and no other uses of `scopeNote` that ought to be `seeAlso` instead.  This should be in the style guide, if it isn't already.

gistCore.ttl
uscholdm: ok

gistCore.ttl
uscholdm: I wondered about that. My thinking was that unlike all the the others, (k)Kelvin is pretty much always capitalized - perhaps due to its originally being a person's name.  It seemed appropriate for a label, it looked wrong in low case.

gistCore.ttl
uscholdm: I don't mean the formal semantics of OWL, I mean what it is representing in the world. I gave it a slight rewording. 
- `This is a gist:Magnutude, not a unit of measure, it means essentially the same thing as the unit, day.`

gistCore.ttl
uscholdm: I think an artifact of Protege, but this is a separate issue.

rjyounes (approved):

PR #596, opened 2021-10-28
Change skos:definition of Task to define it as an Event, not as a Task.

turingtest37:
Patch for a one-word change in the skos definition.


rjyounes:
Replaced by PR #597.

rjyounes:
Now fixed in PR #597.

PR #597, opened 2021-10-28
Proposed skos:definition of Task

turingtest37:
"A function or piece of work to be done or undertaken, usually within a certain time."


rjyounes (changes requested):
IMO we should go with @uscholdm's proposed definition.

rjyounes (approved):

PR #614, opened 2022-01-14
Updates to gist:requires. Fixes #183.

### LINKED ISSUE #183: Remove or loosen constraints on gist:requires
Current domain: Intention
Current range: Behavior
Definition: An intention that sets out a state of satisfaction (you are required to drive on right side of the road)

We may want to say "Plants require good soil and rain to grow." The plant itself does not have an intention, though we might have an intention for the plant. Or maybe the subject of requires here is growth, but this too doesn't have an intention.

The Cambridge Dictionary defines several meanings for required:
1. To need something or to make something necessary.
2. To order or demand something, or to order someone to do something, especially because of a rule or law.
3. To make it officially necessary for someone to do something.

gist:requires encompasses only definitions 2 and 3.

rjyounes:
- Remove domain and range constraints
- Change definition, add an example

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,13 +1,19 @@
 gist Release Notes
 =====
 
-Release X.x.x
+Release 11.0.0
 -----
+
+### Major Updates
+
+- Removed property  `gist:hasOrderedMember`. `gist:hasMember` should be used instead. Issue [#540](https://github.com/semanticarts/gist/issues/540).
+- Removed domain and range constraints from `gist:requires`. Issue [#183](https://github.com/semanticarts/gist/issues/183).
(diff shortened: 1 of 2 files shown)


rjyounes:
@dylan-sa Let me know if you can think of a better definition. I'm not thrilled with this one.

dylan-sa (approved):
@rjyounes Looks good to me. I think the definition is tough to improve, but I'll offer a couple alternatives in case they inspire some tweaks:

- 'The occurrence or existence of the object is a necessary condition for the occurrence or existence of the subject'

- 'The object's obtaining is a necessary condition for the subject's obtaining'

These options are a little more wordy, but they may help to paint a picture of what's really going on when we say 'x requires y'. For example, 'the contract requires two signatures' might mean 'the existence of two signatures is a necessary condition for the existence of the contract'.

PR #619, opened 2022-01-26
New property-based time model. Fixes #499

### LINKED ISSUE #499: Remodel times to use datatype properties rather than TimeInstant
Proposal by @mkumba. Documents attached.

[gistTime-apptx.pptx](https://github.com/semanticarts/gist/files/6735390/gistTime-apptx.pptx)
[experimentsWithDateDatatypes.ttl.txt](https://github.com/semanticarts/gist/files/6735394/experimentsWithDateDatatypes.ttl.txt)

rjyounes:


rjyounes:
Anyone who wants to review in addition to the assigned reviewers, please do!

I've added a few `skos:editorialNotes` to weigh in on.

mkumba (approved):
I'm having a heck of a time figuring out how and where to put comments, hope this is it.

I'm good with changes.  In terms of the editorial notes that were in the presentation and didn't make the file, yeah I think that's a good idea.  

Also on the question of atDateTime being just gist:dateTime I'm ok with that.

rjyounes:
@mkumba @marksem @DanCarey404: @uscholdm and I both think `CurrentEvent` is preferable to and clearer than `ContemporaryEvent`, and I've made that update to the file. Please comment if you don't like it.

uscholdm:
I suggest using "OngoingEvent" because a very recent event will generally be regarded a current. The term 'current event' refers to a general time period not whether the event is still happening.  E.g. a trial that was just completed, its over but it's in the news and is regarded as a current event.

rjyounes:
@uscholdm I think I agree with you. However, does "ongoing" lend a sense of continuing over some significant period of time? E.g., I am currently stopped at a red light, but I wouldn't call that an ongoing event. Is this too much of a quibble?

rjyounes:
I think the advantage of Contemporary is it contrasts more naturally with Historical. But I'm OK with Contemporary, Ongoing, or Current, whatever others prefer.

uscholdm:
Let's hope the event of discussing this is not ongoing for too much longer :-)

rjyounes:
Question: If the date predicates have precision +/- 12 hours, shouldn't the minute predicates have precision +/- 30 seconds rather than 1 minute?

bpelakh:
Just a voice from the peanut gallery, but it's about time (heh) that this change was made. 👍

rjyounes:
@bpelakh Since this is a public repository, you would have been welcome to submit a PR!

DanCarey404:
I went with the term "contemporary" because I had an essentially boolean category :Contemporaneity.  And I had that because there was no real word for "currentness".  I prefer contemporary or current, but not ongoing.

DanCarey404:
I used :atDateTime versus just :dateTime because the latter looks too much like a datatype, in my opinion.  Also, the subproperties I have created in my projects feel more natural with the "at" included.  :startAtDateTime, validateAtDateTime, changeAtDateTime, etc.

rjyounes:
@DanCarey404 Regarding the property names, datatype property names are nouns by convention, and `startDateTime` is a noun, whereas `startAtDateTime` is verbal. It's also less awkward, IMO. I had the same thought about `gist:dateTime` but Dave thought it was fine, and I wanted to keep the names parallel. I'd be OK with `gist:atDateTime` and `gist:startDateTime`, though.

DanCarey404:
This property name is definitely not a hill to die on.  I just wanted to explain the origin in case anyone found it persuasive.

gistDeprecated.ttl
```diff
@@ -0,0 +1,32 @@
...
+		owl:InverseFunctionalProperty
+		;
+	rdfs:subPropertyOf gist:hasMember ;
+	rdfs:range gist:OrderedMember ;
+	owl:deprecated "true"^^xsd:boolean ;
+	skos:definition "Relates an ordered collection to an ordered member that belongs to it."^^xsd:string ;
+	skos:prefLabel "has ordered member"^^xsd:string ;
+	skos:scopeNote "This property is inverse functional because no ordered member can be in more than one ordered collection."^^xsd:string ;
+	.
+
```
uscholdm, line 32: ContemporaneousEvent should be deprecated.

gistCore.ttl
```diff
@@ -266,7 +267,8 @@ gist:CoherentProductUnit
 			]
 		) ;
 	] ;
-	skos:definition "A ratio unit whose numerator and denominator reduce to 1"^^xsd:string ;
+	skos:definition "A product unit both of whose factors are coherent units. The conversion factor is 1."^^xsd:string ;
+	skos:example "Square meter, an area unit."^^xsd:string ;
```
uscholdm, line 271: Why is this showing up in this PR?  Makes it  harder to review.

gistCore.ttl
```diff
@@ -3856,10 +3805,16 @@ gist:recognizes
 
 gist:requires
 	a owl:ObjectProperty ;
-	rdfs:domain gist:Intention ;
-	rdfs:range gist:Behavior ;
-	skos:definition "An intention that sets out a state of satisfaction (you are required to drive on right side of the road)"^^xsd:string ;
+	skos:definition "The subject needs the object or makes it necessary, mandatory, or compulsory."^^xsd:string ;
+	skos:example "Humans require air; solar power requires sunshine."^^xsd:string ;
```
uscholdm, line 3809: Why this PR?

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,33 @@
 gist Release Notes
 =====
 
+Release 11.0.0
+-----
+
+### Major Updates
+
+- Implemented new time model based on datatype properties rather than `TimeInstant`, which has been deleted. Issue [#499](https://github.com/semanticarts/gist/issues/499).
```
uscholdm, line 9: Suggest adding something like this:

- Changed the way to refer to dates and times by moving precision information from subclasses of `TimeInstant` to a new set of subproperties of a generic `atDateTime` datatype property.
  - Previously, we used the class `TimeInstant` and its subclasses to refer to points in time with different precisions, e.g. day, minute microsecond.  We used object properties to connect things to points in time.  
  - We have removed the class `TimeInstant` and converted the time properties to datatype properties.  We added versions of the time properties for different precisions: date, minute, and microsecond. We retained properties that distinguish start vs. end, and planned vs. actual.    
  - We reationalized the naming of the properties and the annotations describing them.
  - We renamed ContemporaneousEvent to ContemporaryEvent

uscholdm (changes requested):
See inline.  A few key things.
- there is some stuff in this PR that belongs elsewhere - how did it get here?
- need more details in release notes

docs/ReleaseNotes.md
rjyounes: Good, thanks. I've adopted with adaptations.

gistDeprecated.ttl
rjyounes: We do not deprecate in a major release. In fact, all existing deprecations are removed.

gistCore.ttl
rjyounes: Something weird happened in the merge from develop and I had to recreate some changes from earlier PRs.

gistCore.ttl
rjyounes: Between develop not having been merged into Dave's branch, and some rebase problems, a number of changes from previous merges got into this PR. Unfortunate, but it would be too difficult to fix now.

rjyounes:
Closing this PR because history has gotten screwed up. Have re-opened as PR #629 with a clean history.

PR #621, opened 2022-01-27
Removing min cardinality of zero restriction on ProductUnit. Fixes #582.

### LINKED ISSUE #582: Remove all minCardinality 0 axioms
Currently there is one on `ProductUnit`. There may be others.

I propose that these hints belong either in domain-specific SHACL or in a `domainIncludes` annotation on the property.

JessSing:
Added scopeNote to use SHACL if needed.

Also updated release notes.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -16,7 +16,7 @@ Release X.x.x
 - Added labels to gist instances, per issue [#370](https://github.com/semanticarts/gist/issues/370).
 - Added definitions for unit of measure instances per issue [#526](https://github.com/semanticarts/gist/issues/526).
 - Deprecated property `gist:hasOrderedMember`. `gist:hasMember` should be used instead. Issue [#540](https://github.com/semanticarts/gist/issues/540).
-- Removed min cardinality of zero restriction on ProductUnit and added scopeNote, per issue [#582](https://github.com/semanticarts/gist/issues/582).
+- Removed min cardinality of zero restriction on `ProductUnit`, per issue [#582](https://github.com/semanticarts/gist/issues/582).
   
 Import URL: <https://ontologies.semanticarts.com/o/gistCoreX.x.x>.
 
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -16,6 +16,7 @@ Release X.x.x
 - Added labels to gist instances, per issue [#370](https://github.com/semanticarts/gist/issues/370).
 - Added definitions for unit of measure instances per issue [#526](https://github.com/semanticarts/gist/issues/526).
 - Deprecated property `gist:hasOrderedMember`. `gist:hasMember` should be used instead. Issue [#540](https://github.com/semanticarts/gist/issues/540).
+- Removed min cardinality of zero restriction on ProductUnit and added scopeNote, per issue [#582](https://github.com/semanticarts/gist/issues/582).
   
 Import URL: <https://ontologies.semanticarts.com/o/gistCoreX.x.x>.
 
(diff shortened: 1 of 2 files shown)


gistCore.ttl
```diff
@@ -2055,17 +2055,12 @@ gist:ProductUnit
...
-				owl:onProperty gist:baseConversionFactor ;
-				owl:minQualifiedCardinality "0"^^xsd:nonNegativeInteger ;
-				owl:onDataRange xsd:double ;
-			]
 		) ;
 	] ;
 	skos:definition "A unit of measure that is the product of two simpler ones."^^xsd:string ;
 	skos:example "Area and Volume are the classic cases.  But other, more exotic cases exist, such as Newtons."^^xsd:string ;
 	skos:prefLabel "Product Unit"^^xsd:string ;
+	skos:scopeNote "Base conversion factor must be a nonzero value; use SHACL for data validation as needed."^^xsd:string ;
```
rjyounes, line 2063: That's not the meaning of the minQualifiedCardinality restriction. It is a hint that `ProductUnit` should have a value for `baseConversionFactor`. What we want to say in the scope note is:

It is intended that a ProductUnit will have a value for baseConversionFactor.

rjyounes (changes requested):
Jess, please also add a note to the release notes. This is a patch fix.

JessSing:
Updated scope note and release notes.

docs/ReleaseNotes.md
```diff
@@ -16,6 +16,7 @@ Release X.x.x
 - Added labels to gist instances, per issue [#370](https://github.com/semanticarts/gist/issues/370).
 - Added definitions for unit of measure instances per issue [#526](https://github.com/semanticarts/gist/issues/526).
 - Deprecated property `gist:hasOrderedMember`. `gist:hasMember` should be used instead. Issue [#540](https://github.com/semanticarts/gist/issues/540).
+- Removed min cardinality of zero restriction on ProductUnit and added scopeNote, per issue [#582](https://github.com/semanticarts/gist/issues/582).
```
rjyounes, line 19: Sorry to be picky, but you can put `gist:ProductUnit` in backticks as you see in other notes. No need to mention the scope note, as it's not significant enough to call out on its own.

gistCore.ttl
```diff
@@ -2055,17 +2055,12 @@ gist:ProductUnit
...
-				owl:onProperty gist:baseConversionFactor ;
-				owl:minQualifiedCardinality "0"^^xsd:nonNegativeInteger ;
-				owl:onDataRange xsd:double ;
-			]
 		) ;
 	] ;
 	skos:definition "A unit of measure that is the product of two simpler ones."^^xsd:string ;
 	skos:example "Area and Volume are the classic cases.  But other, more exotic cases exist, such as Newtons."^^xsd:string ;
 	skos:prefLabel "Product Unit"^^xsd:string ;
+	skos:scopeNote "A ProductUnit is intended have a (nonzero) value for baseConversionFactor."^^xsd:string ;
```
rjyounes, line 2063: You need to remove the "(nonzero)" part here. What the min cardinality restriction is hinting at is that  a product unit is expected to have _a value_, without any indication of what that value should be. I.e., the 0 refers to the number of assertions, not the values asserted. Never mind that a `baseConversionFactor` of 0 makes no sense - that doesn't need to be stated here but is a general fact about the meaning of that predicate.

rjyounes (changes requested):

docs/ReleaseNotes.md
JessSing: No I should have matched the formatting, sorry about that.

rjyounes (approved):

PR #630, opened 2022-02-16
Add predicates with year precision. Fixes #388.

### LINKED ISSUE #388: Add properties for year precision to new time model
Just like DateInstant, but with precision of a year. Useful when we want to store data about which year something happened.

rjyounes:
Added predicates:

```
actualStartYear
actualEndYear
plannedStartYear
plannedEndYear
```

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -6,10 +6,10 @@ Release 11.0.0
 
 ### Major Updates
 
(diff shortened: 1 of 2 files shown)


Jamie-SA:
How would you say that something happened in a year? Or that data (revenue, GDP, etc) was for the year 2022? Would it be expected to use both a start and end property with the same data value?

daliadahleh (approved):

PR #639, opened 2022-03-15
GovernedGeoRegions and GovernmentOrganizations. Fixes #215.

### LINKED ISSUE #215: Consider adding Country, which is different from CountryGovernment
A country has a government. The government could be disbanded and a new government created without changing the identity of a country. Also, as Dan points out, it's a concept used by many of our clients.

rjyounes:
Applied serializer, merged in changes from develop, and made a few other changes to formal definitions and annotations. 

@DanCarey404 Please review this draft PR before I turn it over to anyone else for review, as I've made some changes to both formal and textual definitions. I've added some `skos:editorialNote`s starting with "DAN" to flag a couple of issues for you. If necessary we can jump on a call to iron out details.

My first step was to run the serializer and merge in develop (commit ee5864b). The second commit involves some actual updates (dc2ade1). So the best way to view the changes that I've made is to compare those to commits.


gistCore.ttl
```diff
@@ -692,11 +689,18 @@ gist:CountryGovernment
...
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:isGovernedBy ;
+				owl:onClass gist:GovernmentOrganization ;
+				owl:maxQualifiedCardinality "0"^^xsd:nonNegativeInteger ;
 			]
 		) ;
 	] ;
 	skos:definition "A Government Organization which asserts both sovereignty (i.e., it is not governed by some other government organization) and governance over an entity generally recognized as a 'country'."^^xsd:string ;
+	skos:editorialNote "DAN: Changed the second restriction to CountryGeoRegion, not GeoRegion. I don't understand the first restriction: it states that a country government is the set of all things are are not governed by government organizations, which includes chairs. I think you want a qualified max cardinality of 0 restriction, which I've added above for your review. Is it correct to also say that it governs exactly one GeoRegion, rather than one or more? I.e., a qualified cardinality restriction of 1 on governs CountryGeoRegion."^^xsd:string ;
```
DanCarey404, line 703: I don't understand your qualification about "chairs".  Any chair would be established by/subject to/governed by that government.  
I thought I had a restriction that a CountryGovernment is governed by 0 GovernmentOrgs.  But I don't see that in the comparison.  So I think your restriction is okay.
I'm on the fence about a CountryGov governing more than one country.  But since we don't have a Country class anymore, it is probably safer to make it a min 1 restriction.

gistCore.ttl
```diff
@@ -1163,17 +1166,22 @@ gist:GovernedGeoRegion
...
+	skos:prefLabel "Governed Geo-Region"^^xsd:string ;
+	skos:scopeNote "Geographic regions do not need not be physically contiguous in order to constitute a governed geo-region; e.g., Alaska and Hawaii."^^xsd:string ;
 	.
 
 gist:GovernmentOrganization
 	a owl:Class ;
 	rdfs:subClassOf gist:Organization ;
-	skos:definition "An Organization which exercises political and/or regulatory authority over a political unit, people, georegion, etc, as well as performs certain functions for this unit or body. Differs from a corporation in that it cannot be owned."^^xsd:string ;
+	owl:disjointWith
+		gist:IntergovernmentalOrganization ,
```
DanCarey404, line 1178: Oooh, good catch on the disjoints.

gistCore.ttl
```diff
@@ -663,20 +662,18 @@ gist:CountryGeoRegion
...
 				owl:onClass gist:CountryGovernment ;
-				owl:maxQualifiedCardinality "1"^^xsd:nonNegativeInteger ;
+				owl:qualifiedCardinality "1"^^xsd:nonNegativeInteger ;
 			]
 		) ;
 	] ;
-	skos:definition "A defined geographical area (or areas) governed by only one country government."^^xsd:string ;
-	skos:prefLabel "Country Geo Region"^^xsd:string ;
+	skos:definition "A defined geographical area (or areas) governed by exactly one country government."^^xsd:string ;
+	skos:editorialNote "DAN: changed qualified cardinality restriction from max 1 to exactly 1. Is there a problem with that?"^^xsd:string ;
```
DanCarey404, line 670: I _think_ there was a reason I used max 1.  But I can't remember it.  So let's go with exactly 1.  If someone can later demonstrate a real problem with that, then we will change it.

gistCore.ttl
```diff
@@ -2545,7 +2552,9 @@ gist:TreatyOrganization
 			]
 		) ;
 	] ;
-	skos:definition "An organization the membership of which is comprised of Country Governments."^^xsd:string ;
+	skos:definition "An organization whose members are country governments."^^xsd:string ;
+	skos:editorialNote "DAN: do we want to say that *only* country governments are members?"^^xsd:string ;
```
DanCarey404, line 2556: It is my understanding that only countries can enter into treaties.

DanCarey404 (approved):
See my responses below.  But all your changes are okay with me.

Issue #148, opened 2020-01-21
Fix definition of hasJurisdiction 


rjyounes:
"when laws and contracts are meeted out"

1. Definition seems wrong altogether
2. "meted" not "meeted"

rjyounes:
Dictionary definitions of "jurisdiction":

Google:
"The official power to make legal decisions and judgements."

Merriam-Webster:
"The power, right, or authority to interpret and apply the law."
"the authority of a sovereign power to govern or legislate"
"the power or right to exercise authority"
"the limits or territory within which authority may be exercised"

Cambridge:
"the official authority to make (esp. legal) decisions and judgments"

Wikipedia:
"the practical authority granted to a legal body to administer justice within a defined field of responsibility...Colloquially it is used to refer to the geographical area to which such authority applies, e.g. the court has jurisdiction over all of Colorado. The legal term refers only to the granted authority, not to a geographical area."

rjyounes:
Recommendation:

skos:definition: "The legal body has been granted the practical authority to administer justice within a defined field of responsibility."
skos:scopeNote: "While colloquially 'jurisdiction' is used to refer to the geographical area to which such authority applies, e.g. the court has jurisdiction over all of Colorado. The legal term refers only to the granted authority, not to a geographical area."

rjyounes:
I also recommend the property name be changed to hasJurisdictionOver, since hasJurisdiction does not imply the existence of an object.

uscholdm:
It turns out the concept of jurisdiction is multifaceted.  In 2010, Dave and I read a 20-page paper describing this, authored by Lexis Nexis. Jurisdiction can be over a region, but also a subject. A police department has jurisdicion over drugs but not prostitution, say.  Id be careful about restictiong the domain or range.

rjyounes:
Agree completely.

marksem:
+1

I agree the definition should change, and I'm good with the local name change as well.

rjyounes:
@uscholdm There's no domain or range stated - though possibly we could define an appropriate domain. If you're referring to my proposed scope note, that can be reworded:

"While colloquially 'jurisdiction' is sometimes used in relation to geographic areas, such as jurisdiction over the State of Colorado, legal jurisdiction pertains to the administrative area rather than the geographic one."

- The United States District Court for the Eastern District of New York has federal jurisdiction over the counties of Kings, Nassau, Queens, Richmond, and Suffolk.
[...]

PR #640, opened 2022-03-15
Rename hasJurisdictionOver. Fixes #637.

### LINKED ISSUE #623: Move gist instances to a separate file?
Consider in conjunction with #305 and #556.

### LINKED ISSUE #637: Rename `hasJurisdictionOver`
This was incorrectly renamed in gist 10 due to a misunderstanding of the intended meaning. Given the definition of `Contract`:

```
gist:Contract
	a owl:Class ;
	owl:equivalentClass [
		a owl:Class ;
		owl:intersectionOf (
			gist:Agreement
			[
				a owl:Restriction ;
				owl:onProperty gist:hasJurisdictionOver ;
				owl:someValuesFrom gist:GovernmentOrganization ;
			]
		) ;
	] ;
.
```
The change to `hasJurisdictionOver` reverses the intended meaning and the restriction is now incorrect. This issue involves a name change; providing a better definition is included in issue #148.

rjyounes:
Corrected the local name and made a quick fix to the definition so that it's intelligible. Further work on the definition is called for in #148.

The history of this is that when local names were updated in gist 10, the meaning of the predicate was misunderstood and so changed. I realized the mistake and have now changed it back.

### DIFF
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -515,7 +515,7 @@ gist:Contract
 			gist:Agreement
 			[
 				a owl:Restriction ;
-				owl:onProperty gist:hasJurisdiction ;
+				owl:onProperty gist:isUnderJurisdiction ;
 				owl:someValuesFrom gist:GovernmentOrganization ;
 			]
 		) ;
@@ -3137,12 +3137,6 @@ gist:hasIncumbent
 	skos:prefLabel "has incumbent"^^xsd:string ;
 	.
 
-gist:hasJurisdiction
-	a owl:ObjectProperty ;
-	skos:definition "Relates a law, contract, etc., to the system of law or government which has the power, right, or authority to interpret and apply it."^^xsd:string ;
-	skos:prefLabel "has jurisdiction"^^xsd:string ;
-	.
-
 gist:hasMagnitude
 	a owl:ObjectProperty ;
 	rdfs:range gist:Magnitude ;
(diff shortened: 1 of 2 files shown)


gistCore.ttl
```diff
@@ -3137,10 +3137,10 @@ gist:hasIncumbent
 	skos:prefLabel "has incumbent"^^xsd:string ;
 	.
 
-gist:hasJurisdictionOver
+gist:hasJurisdiction
 	a owl:ObjectProperty ;
-	skos:definition "When laws and contracts are meted out"^^xsd:string ;
-	skos:prefLabel "has jurisdiction over"^^xsd:string ;
+	skos:definition "Relates a law, contract, etc., to the system of law or government which has the power, right, or authority to interpret and apply it."^^xsd:string ;
+	skos:prefLabel "has jurisdiction"^^xsd:string ;
```
uscholdm, line 3143: The reason for the confusion is that 'hasJurisdiction' sounds like it means has jurisdiction over. Perhaps a better name would be along the lines of isGovernedByJurisdiction

uscholdm (changes requested):
The reason for the confusion is that 'hasJurisdiction' sounds like it means has jurisdiction over. Perhaps a better name would be along the lines of isGovernedByJurisdiction. Consider an alternative.

gistCore.ttl
rjyounes: Good idea. I'll use that alternative since it seems pretty good and I can't think of a better one.

rjyounes: @uscholdm Changed name to `isUnderJurisdiction`. Not sure if it should be `isUnderJurisdictionOf`.

PR #641, opened 2022-03-15
Modifications to biological parent/child predicates and related restrictions. Fixes #638.

### LINKED ISSUE #638: Delete `hasBiologicalOffspring`
And change the restriction on LivingThing:

```
			[
				a owl:Restriction ;
				owl:onProperty gist:hasBiologicalParent ;
				owl:someValuesFrom gist:LivingThing ;
			]
```
to

```
			[
				a owl:Restriction ;
				owl:onProperty gist:hasBiologicalParent ;
				owl:minCardinality "1"^^xsd:nonNegativeInteger ;
			]
```
`gist:rangeIncludes gist:Person` also seems unnecessary in that case.

rjyounes:


gistCore.ttl
```diff
@@ -1285,7 +1285,7 @@ gist:LivingThing
 			[
 				a owl:Restriction ;
 				owl:onProperty gist:hasBiologicalParent ;
-				owl:someValuesFrom gist:LivingThing ;
+				owl:minCardinality "1"^^xsd:nonNegativeInteger ;
```
uscholdm, line 1288: Do you have in mind a non-living thing that might be a parent? I prefer leave the restriction unchanged.

If you really think that is possible, then its best to use someValuesFrom owl:Thing rather than a min 1, which reduces flexibility and can cause inference problems in some cases, the ones you least expect to getcha :-)

gistCore.ttl
```diff
@@ -1295,6 +1295,7 @@ gist:LivingThing
 		"Negative examples: fictional life forms such as unicorns or Mickey Mouse."^^xsd:string
 		;
 	skos:prefLabel "Living Thing"^^xsd:string ;
+	skos:scopeNote "Not all life forms have exactly 2 parents, so the restriction only specifies a minimum of one."^^xsd:string ;
```
uscholdm, line 1298: BUT: all biological parents surely are (or were) living.

gistCore.ttl
```diff
@@ -1769,7 +1770,12 @@ gist:Person
 			[
 				a owl:Restriction ;
 				owl:onProperty gist:hasBiologicalParent ;
-				owl:someValuesFrom gist:Person ;
+				owl:allValuesFrom gist:Person ;
+			]
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:hasBiologicalParent ;
+				owl:cardinality "2"^^xsd:nonNegativeInteger ;
```
uscholdm, line 1778: Is says **2** parents, but you said 1.

uscholdm (changes requested):

gistCore.ttl
rjyounes: How is this relevant?

gistCore.ttl
rjyounes: A Person has two parents, but not all LivingThings do (e.g., asexual reproduction).

gistCore.ttl
rjyounes: I can go with `someValuesFrom` for a minCardinality, but in that case I prefer `owl:someValuesFrom owl:Thing`, since the range of the predicate is `LivingThing`. I know, we've had that discussion many times and I don't think either of us will be convinced.

daliadahleh: I agree owl:someValuesFrom owl:Thing makes sense here so that there's no redundancy.

gistCore.ttl
daliadahleh: That's already taken care of with the domain and range of hasBiologicalParents. I don't think it's relevant here.

daliadahleh (changes requested):

PR #649, opened 2022-03-28
Combine standard and base conversion factor predicates into a single `conversionFactor` predicate. Fixes #624.

### LINKED ISSUE #624: Correct axioms related to baseConversionFactor and standardConversionFactor, and clarify definitions
@uschold and I think the following changes should be made, based on discussion related to issue #582:

* Make `baseConversionFactor` a subproperty of `standardConversionFactor` instead of the other way around (we got this right in making `hasBaseUnit` a subproperty of `hasStandardUnit`, interestingly enough.
* The class `UnitOfMeasure` should have cardinality restrictions of 1 for both `standardConversionFactor` and `hasStandardUnit`. Such restrictions should then be removed from the subclasses.
* In addition, the definitions need to be cleaned up and clarified:

The current definition of `standardConversionFactor` is a scope note and provides no clarification of the difference between it and `baseConversionFactor`. The language in the latter's definition could also be cleaned up and split into a definition, example, and scope note.

`baseConversionFactor`: The conversion factor used to get to the base unit.  E.g., multiplying by 0.0254 gets you from inches to meters. Divide by this number to go the other way. Used in conjunction with conversionOffset to convert from one unit to another. Degrees K =  (Degrees F - conversionOffset) * convertToBase. Or K = (F-(-469.67)) * (5/9.  To go the other way: F = (K * 9/5) -469.67.  Try it on Google.

`standardConversionFactor`: Note this kind of conversion will only work with temperatures if they are in Kelvin or Rankine (with a true 0).  You multiply to get to the base, divide to go from the base. mph to mps is .44704.  The multiple from kph to mps is .277778 .  To convert 60 mph to kph is (60 * .44704 / .277778 or 96.56056 kph

wslooney:
Eliminated :standardConversionFactor and :baseConversionFactor. Created :conversionFactor instead with cleaned up annotations.

Added cardinality restrictions of 1 to :UnitOfMeasure for :conversionFactor and :hasStandardUnit

Removed cardinality restriction on :hasBaseUnit in :SimpleUnitOfMeasure

### DIFF
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -2008,7 +2008,7 @@ gist:ProductUnit
 	skos:definition "A unit of measure that is the product of two simpler ones."^^xsd:string ;
 	skos:example "Area and Volume are the classic cases.  But other, more exotic cases exist, such as Newtons."^^xsd:string ;
 	skos:prefLabel "Product Unit"^^xsd:string ;
-	skos:scopeNote "A ProductUnit is intended have a value for baseConversionFactor."^^xsd:string ;
+	skos:scopeNote "A ProductUnit is intended have a value for conversionFactor."^^xsd:string ;
 	.
 
 gist:Project
@@ -2178,12 +2178,6 @@ gist:SimpleUnitOfMeasure
 				a owl:Restriction ;
 				owl:onProperty gist:hasBaseUnit ;
 				owl:onClass gist:BaseUnit ;
-				owl:qualifiedCardinality "1"^^xsd:nonNegativeInteger ;
-			]
-			[
-				a owl:Restriction ;
-				owl:onProperty gist:baseConversionFactor ;
-				owl:someValuesFrom xsd:double ;
 			]
 		) ;
 	] ;
@@ -2509,6 +2503,22 @@ gist:TreatyOrganization
 
 gist:UnitOfMeasure
 	a owl:Class ;
+	owl:equivalentClass [
+		a owl:Class ;
+		owl:intersectionOf (
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:hasStandardUnit ;
+				owl:cardinality "1"^^xsd:nonNegativeInteger ;
+			]
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:conversionFactor ;
+				owl:cardinality "1"^^xsd:nonNegativeInteger ;
+			]
+		) ;
+	] ;
+
 	skos:definition "Standard unit by which we measure things"^^xsd:string ;
 	skos:prefLabel "Unit of Measure"^^xsd:string ;
 	.
@@ -2562,7 +2572,7 @@ gist:_USDollar
 		;
 	skos:definition "The base unit for currency."^^xsd:string ;
 	skos:prefLabel "US Dollar"^^xsd:string ;
-	gist:baseConversionFactor "1.0"^^xsd:double ;
+	gist:conversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_USDollar ;
 	gist:unitSymbol "USD"^^xsd:string ;
 	.
@@ -2575,7 +2585,7 @@ gist:_ampere
 		;
 	skos:definition "The base unit for electrical current."^^xsd:string ;
 	skos:prefLabel "ampere"^^xsd:string ;
-	gist:baseConversionFactor "1.0"^^xsd:double ;
+	gist:conversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_ampere ;
 	gist:unitSymbol "A"^^xsd:string ;
 	.
@@ -2589,7 +2599,7 @@ gist:_bit
 	skos:definition "The base unit for measuring digital information."^^xsd:string ;
 	skos:prefLabel "bit"^^xsd:string ;
 	skos:scopeNote "A bit (short for binary digit) is the smallest unit of data in a computer. A bit has a single binary value, either 0 or 1."^^xsd:string ;
-	gist:baseConversionFactor "1.0"^^xsd:double ;
+	gist:conversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_bit ;
 	.
 
@@ -2601,7 +2611,7 @@ gist:_candela
 		;
 	skos:definition "The base unit for luminous intensity."^^xsd:string ;
 	skos:prefLabel "candela"^^xsd:string ;
-	gist:baseConversionFactor "1.0"^^xsd:double ;
+	gist:conversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_candela ;
 	gist:unitSymbol "cd"^^xsd:string ;
 	.
@@ -2613,7 +2623,7 @@ gist:_day
 		;
 	skos:definition "A duration unit that is 24 hours long."^^xsd:string ;
 	skos:prefLabel "day"^^xsd:string ;
-	gist:baseConversionFactor "86400.0"^^xsd:double ;
+	gist:conversionFactor "86400.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_second ;
 	.
 
@@ -2625,7 +2635,7 @@ gist:_each
 		;
 	skos:definition "The base unit for count magnitudes."^^xsd:string ;
 	skos:prefLabel "each"^^xsd:string ;
-	gist:baseConversionFactor "1.0"^^xsd:double ;
+	gist:conversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_each ;
 	.
 
@@ -2647,7 +2657,7 @@ gist:_kelvin
 		;
 	skos:definition "The base unit for measuring temperature."^^xsd:string ;
 	skos:prefLabel "Kelvin"^^xsd:string ;
-	gist:baseConversionFactor "1.0"^^xsd:double ;
+	gist:conversionFactor "1.0"^^xsd:double ;
 	gist:conversionOffset "0.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_kelvin ;
 	gist:unitSymbol "K"^^xsd:string ;
@@ -2661,7 +2671,7 @@ gist:_kilogram
 		;
 	skos:definition "The base unit for measuring mass."^^xsd:string ;
 	skos:prefLabel "mass"^^xsd:string ;
-	gist:baseConversionFactor "1.0"^^xsd:double ;
+	gist:conversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_kilogram ;
 	gist:unitSymbol "kg"^^xsd:string ;
 	.
@@ -2674,7 +2684,7 @@ gist:_meter
 		;
 	skos:definition "The base unit for measuring distance."^^xsd:string ;
 	skos:prefLabel "distance"^^xsd:string ;
-	gist:baseConversionFactor "1.0"^^xsd:double ;
+	gist:conversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_meter ;
 	gist:unitSymbol "m"^^xsd:string ;
 	.
@@ -2686,7 +2696,7 @@ gist:_millisecond
 		;
 	skos:definition "A unit equal to a thousandth of a second."^^xsd:string ;
 	skos:prefLabel "millisecond"^^xsd:string ;
-	gist:baseConversionFactor "0.001"^^xsd:double ;
+	gist:conversionFactor "0.001"^^xsd:double ;
 	gist:hasBaseUnit gist:_second ;
 	gist:unitSymbol "ms"^^xsd:string ;
 	.
@@ -2698,7 +2708,7 @@ gist:_minute
 		;
 	skos:definition "A unit equal to 60 seconds."^^xsd:string ;
 	skos:prefLabel "minute"^^xsd:string ;
-	gist:baseConversionFactor "60.0"^^xsd:double ;
+	gist:conversionFactor "60.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_second ;
 	gist:unitSymbol "min"^^xsd:string ;
 	.
@@ -2711,7 +2721,7 @@ gist:_mole
 		;
 	skos:definition "The base unit for measuring molar quantities."^^xsd:string ;
 	skos:prefLabel "mole"^^xsd:string ;
-	gist:baseConversionFactor "1.0"^^xsd:double ;
+	gist:conversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_mole ;
 	gist:unitSymbol "mol"^^xsd:string ;
 	.
@@ -2763,7 +2773,7 @@ gist:_second
 		;
 	skos:definition "The base unit for measuring durations."^^xsd:string ;
 	skos:prefLabel "second"^^xsd:string ;
-	gist:baseConversionFactor "1.0"^^xsd:double ;
+	gist:conversionFactor "1.0"^^xsd:double ;
 	gist:hasBaseUnit gist:_second ;
 	gist:unitSymbol "s"^^xsd:string ;
 	.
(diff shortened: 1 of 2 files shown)


uscholdm (commented):
I looked at the related issue and found no mention of removing the property `baseConverstionFactor`.   I don't know if I agree with this. What as the rational for this? Was it agreed among several parties after a discussion? It so, it should be documented in the issue.

wslooney:
> 

Sorry, I thought it was documented in the Issue page (I know it's _somewhere_ maybe it was on a different page). We decided in the gist meeting on 3/24/22 that there was no need distinguish between :baseConversionFactor and :standardConversaionFactor . A unit of measure still bears the :hasStandardUnit or :hasBaseUnit property to its standard unit or base unit. It was agreed that there was nothing gained by also making that distinction in the conversionFactor to the unit.

gistCore.ttl
```diff
@@ -2893,11 +2903,13 @@ The conventions for precision that are repeated in each property name are as fol
...
 	.
 
-gist:baseConversionFactor
+gist:conversionFactor
 	a owl:DatatypeProperty ;
 	rdfs:domain gist:UnitOfMeasure ;
 	rdfs:range xsd:double ;
-	skos:definition "The conversion factor used to get to the base unit.  E.g., multiplying by 0.0254 gets you from inches to meters. Divide by this number to go the other way. Used in conjunction with conversionOffset to convert from one unit to another. Degrees K =  (Degrees F - conversionOffset) * convertToBase. Or K = (F-(-469.67)) * (5/9.  To go the other way: F = (K * 9/5) -469.67.  Try it on Google."^^xsd:string ;
+	skos:definition "The conversion factor used to get to the standard unit (SI unit when applicable)."^^xsd:string ;
+	skos:example "Inches should have a conversionFactor of 0.0254 in order to go from inches to meters."^^xsd:string ;
```
rjyounes, line 2911: This is awkward. Something more like "The unit of measure inch is multiplied by 0.0254 to convert to the base unit meter."

The definition should mention both base and standard units, something like "The conversion factor used to convert the unit to its standard or base unit."

rjyounes (changes requested):
A couple of minor comments on wording.

gistCore.ttl
wslooney: I left the base unit part out of the definition because a BaseUnit is a StandardUnit, but I'm happy to add it in if it makes things clearer/easier to use.

wslooney:
I adopted the suggestions from @rjyounes. I also changed the `skos:prefLabel` for `gist:conversionFactor` because I had forgotten to change it earlier from "base conversion factor."

gistCore.ttl
wslooney: I adopted the suggestions from @rjyounes. I also changed the `skos:prefLabel` for `gist:conversionFactor` because I had forgotten to change it earlier from "base conversion factor."

rjyounes: I realize it's redundant but it helps to clarify things.

gistCore.ttl
```diff
@@ -2893,12 +2903,14 @@ The conventions for precision that are repeated in each property name are as fol
...
 
-gist:baseConversionFactor
+gist:conversionFactor
 	a owl:DatatypeProperty ;
 	rdfs:domain gist:UnitOfMeasure ;
 	rdfs:range xsd:double ;
-	skos:definition "The conversion factor used to get to the base unit.  E.g., multiplying by 0.0254 gets you from inches to meters. Divide by this number to go the other way. Used in conjunction with conversionOffset to convert from one unit to another. Degrees K =  (Degrees F - conversionOffset) * convertToBase. Or K = (F-(-469.67)) * (5/9.  To go the other way: F = (K * 9/5) -469.67.  Try it on Google."^^xsd:string ;
-	skos:prefLabel "base conversion factor"^^xsd:string ;
+	skos:definition "The conversion factor used to convert a unit to its standard or base unit."^^xsd:string ;
+	skos:example "The unit of git measure inch is multiplied by a conversion factor of 0.0254 to convert to its base unit meter."^^xsd:string ;
```
rjyounes, line 2911: typo - gist not git. But it's not gist-specific, so let's leave that word out. There's still something odd about this sentence. Let's try something like this:

"A value expressed in inches is multiplied by a conversion factor of 0.0254 to express the value in the base unit, meters."

gistCore.ttl
```diff
@@ -2177,12 +2177,6 @@ gist:SimpleUnitOfMeasure
 				a owl:Restriction ;
 				owl:onProperty gist:hasBaseUnit ;
 				owl:onClass gist:BaseUnit ;
-				owl:qualifiedCardinality "1"^^xsd:nonNegativeInteger ;
```
rjyounes, line 2180: What got changed here?

rjyounes (changes requested):
Still some wording changes, and I forgot to ask you to add a line to the release notes; just follow the patterns used there. For now we'll assume this is going into 11.0.0, but we need to get approval from Michael and Dan before we can merge it, because it's a major change. (I'm not sure why the title of the PR says "minor." This is complicated. If it goes into a minor version then it's a minor change, because we keep it but deprecate it. In major versions we remove all deprecations, so then it's a major change.)

gistCore.ttl
wslooney: `:UnitOfMeasure` now how as cardinality restriction of 1 on `:hasStandardUnit` and `:SimpleUnitOfMeasure` is a subclass of `:UnitOfMeasure`. I left the restriction that `:SimpleUnitOfMeasure` `:hasBaseUnit`  `:BaseUnit` but removed the cardinality restriction because `:hasBaseUnit` is a subproperty of `:hasStandardUnit`. So the cardinality is already entailed.

gistCore.ttl
wslooney: What I was aiming for with the skos:example is to make sure it's clear in the example that conversionFactor is linking the unit of measure (inch) to the specific xsd:double (0.0254). I didn't want the example to wrongly suggest that the conversion factor links the magnitude to the conversion factor.

wslooney:
Sorry. Misunderstanding on my part what counts as minor/major. I'll change the subject too, to reflect that.

gistCore.ttl
wslooney: Adopted @rjyounes suggestion for commit 2051d3fb1ad5815727a5b5eafdc3363893986963

gistCore.ttl
```diff
@@ -2893,11 +2903,13 @@ The conventions for precision that are repeated in each property name are as fol
 """^^xsd:string ;
 	.
 
-gist:baseConversionFactor
+gist:conversionFactor
 	a owl:DatatypeProperty ;
 	rdfs:domain gist:UnitOfMeasure ;
 	rdfs:range xsd:double ;
-	skos:definition "The conversion factor used to get to the base unit.  E.g., multiplying by 0.0254 gets you from inches to meters. Divide by this number to go the other way. Used in conjunction with conversionOffset to convert from one unit to another. Degrees K =  (Degrees F - conversionOffset) * convertToBase. Or K = (F-(-469.67)) * (5/9.  To go the other way: F = (K * 9/5) -469.67.  Try it on Google."^^xsd:string ;
+	skos:definition "The conversion factor used to get to the standard unit (SI unit when applicable)."^^xsd:string ;
```
pwin: I'm wondering if this needs a little more in the definition - e.g. that to go to the standard unit there is a multiplication of the inches value with this conversion factor.  Also, should we mention that it is reversible through division, or otherwise reversing the process of forward conversion (e.g. with the Fahrenheit to Kelvin)?

wslooney: The skos:definition on the most recent commit is: "The conversion factor used to convert a unit to its standard or base unit."

This is very laconic, but, I think, sufficient. I think what's most important is that it's clear what the direction the conversion factor goes (from some unit **to** its standard unit and only to the standard unit). The word "factor" already indicates that it has to be multiplied, and the skos:example makes that clearer if needed.

I also don't think anything is gained by indicating that "x * y = z" is equivalent to "z / y = x. " But maybe there's a case where that needs to be explained that I'm not imagining. Does anyone else have thoughts?

rjyounes: I don't think we need to explain that a multiplication can be reversed by division. We're not teaching 5th grade arithmetic, after all. :)

rjyounes: What I do think should be made explicit is to add base - even though this is technically redundant - and not everyone knows what SI is. So along the lines of

"The conversion factor used to get to the standard unit (which could be a base unit) in the International System of Units (SI)."

gistCore.ttl
rjyounes: It's also redundant to include the object in the SimpleUnitOfMeasure restriction. You can use `owl:Thing` instead.

rjyounes (changes requested):
Still a couple of small questions. We can talk through them if you want.

gistCore.ttl
wslooney: The most recent commit does include that the conversion factor converts to standard or base unit.

wslooney:
@rjyounes OK, I see. Does this look right? The restriction should be:

```
			[
				a owl:Restriction ;
				owl:onProperty gist:hasBaseUnit ;
				owl:someValuesFrom owl:Thing ;
			]
```
What I had before was (a) semantically wrong because I used `owl:onClass` and (b) redundant for the reason you gave.

gistCore.ttl
uscholdm: > The conversion factor used to get to the standard unit (which could be a base unit) in the International System of Units (SI).

BTW, "Standard unit" is not a standard term for what we mean. That would be "coherent unit". We might add a scopeNote to that effect.

wslooney: It is funny that gist uses the property `hasStandardUnit `with a range of `CoherentUnit`

rjyounes:
@wslooney That's what I meant, but I've changed my mind.

Now I'm looking at the full file and I see what you proposed is incorrect:

```
			[
				a owl:Restriction ;
				owl:onProperty gist:hasBaseUnit ;
				owl:onClass gist:BaseUnit ;
			]
```

You have to have a cardinality or values assertion here. This by itself is not a valid restriction.

Now I'm seeing that the definition of `hasStandardUnit` must be changed. First, given that `hasBaseUnit` is a subproperty of `hasStandardUnit`, the latter should no longer refer only to complex units. 

But it seems to be more complicated. Since I'm no expert in this area, I've done a bit of research. I had thought - and apparently the gist authors did as well - that a base unit was for a simple unit of measure and a standard unit can apply to either simple or complex units.  But from what I am reading, standard unit is a generic concept not related to any particular system of units, while base units are standard units defined by SI. SI defines _derived_ units that complex units can be reduced to, but I don't see any mention of standard units. If this is correct (which I'm not sure of), the cardinality restriction on UnitOfMeasure.hasStandardUnit is incorrect - if there are different systems of standard units, a unit of measure could in theory have different standard units.

My recommendation is that this MR include only the reduction of `baseConversionFactor` and `standardConversionFactor` to `conversionFactor` - this is uncontroversial, I think - and that we leave the other issue to our future overhaul of units and magnitudes.  [Phil Blackwood](https://github.com/philblackwood) is very knowledgeable in this area and will surely have some insight. Since we are trying to get the release out next week, we don't have time to fully sort out the issues around base and standard units.

wslooney:
I realized my initial restriction was not valid hence my comment that it was semantically wrong above.

So just to make sure I understand what to do:

1) Leave the restrictions on UnitOfMeasure and its subclasses as they were EXCEPT change restrictions on `baseConversionFactor` to `conversionFactor`
2) Reduce `baseConversionFactor` and `standardConversionFactor` to `conversionFactor`
3) Replace mentions of `baseConversionFactor` and `standardConversionFactor` that appear in some annotations.

rjyounes (approved):
Looks good!

gistCore.ttl
```diff
@@ -2007,7 +2007,7 @@ gist:ProductUnit
 	skos:definition "A unit of measure that is the product of two simpler ones."^^xsd:string ;
 	skos:example "Area and Volume are the classic cases.  But other, more exotic cases exist, such as Newtons."^^xsd:string ;
 	skos:prefLabel "Product Unit"^^xsd:string ;
-	skos:scopeNote "A ProductUnit is intended have a value for baseConversionFactor."^^xsd:string ;
+	skos:scopeNote "A ProductUnit is intended have a value for conversionFactor."^^xsd:string ;
```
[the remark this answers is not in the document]
uscholdm: We might want to skip this for now -- it is tangential to this PR. If we skip it, it should be added as an issue.

gistCore.ttl
```diff
@@ -2971,6 +2963,16 @@ gist:contributesTo
 	skos:prefLabel "contributes to"^^xsd:string ;
 	.
 
+gist:conversionFactor
+	a owl:DatatypeProperty ;
+	rdfs:domain gist:UnitOfMeasure ;
+	rdfs:range xsd:double ;
+	skos:definition "The conversion factor used to convert a unit to its standard (i.e., coherent) or base unit."^^xsd:string ;
```
uscholdm, line 2970: Suggest slight variation to be one of the following:
1. "The conversion factor used to convert a unit to its standard (i.e., coherent) unit."
2. "The conversion factor used to convert a unit to its standard (i.e., coherent) unit (which could be a base unit.)"

gistCore.ttl
```diff
@@ -2971,6 +2963,16 @@ gist:contributesTo
...
+	skos:example "A value expressed in inches is multiplied by a conversion factor of 0.0254 to express the value in the base unit, meter."^^xsd:string ;
```
uscholdm, line 2971: Add a non-base unit example such as:
`skos:example "The area expressed as 7 square kilometers is multiplied by a conversion factor of one million which results in 7 million square-meters."^^xsd:string ;`

uscholdm (changes requested):
This is nearly good to go. A couple minor suggestions to change annotations and a suggestion which can wait till another time, if we wish.

gistCore.ttl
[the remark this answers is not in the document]
wslooney: Thanks for raising this. I agree this should be changed, but I think it gets us back into stuff that I changed then changed back, i.e., the two restrictions that should just be on `UnitOfMeasure`. For some both restrictions are in `SimpleUnitOfMeasure`. Here one of the restrictions (but not the one on `hasStandardUnit`) in `ProductUnit` is just noted in an annotation rather than being semantically encoded. In `RatioUnit` it's not even annotated. It's kind of a mess.

I think a new issue should be raised that was originally part of the one this addressed for fixing the restrictions on `ProductUnit`.

gistCore.ttl
wslooney: Is the suggestion to add a second skos:example or to replace the one already here? I'm reading you as requesting to add a second skos:example. I just wanted to make sure.

uscholdm: Yes. Add a second example to show two cases: base unit and non-base unit.  It could be a second skos:example annotation, or it could be rolled into one. I don't know if we have a convention on that yet. Look for other uses of skos:example to see if we are doing anything consistent there.  If not, then ontologist's choice!

gistCore.ttl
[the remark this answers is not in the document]
uscholdm: Yes, it should be a new issue.

rjyounes (changes requested):
We are still missing a release note.

wslooney:
@rjyounes It seemed like the only way I could edit the release notes was to create a branch. I had done that, but just gave that branch a more conspicuous name. Not sure what the procedures are for editing the notes (do I create a PR and assign people?)

https://github.com/semanticarts/gist/tree/Release-Note-Edit-ConversionFactor

wslooney:
Sorry, new to github. Dylan told me I should just edit the release notes on **this branch**. Which in hindsight is obvious.

gistCore.ttl
wslooney: I have made the changes @uscholdm quested but without the hyphen in square meter.

I don't love the way the skos:examples are phrased because both examples are phrased in a way that makes it sound like the conversion factor is applied to a specific magnitude rather than a unit, but my previous attempts to make a skos:example that didn't sound that way ended up with fairly tortured English.

rjyounes:
@wslooney The conversion factor is _defined on_ a unit but _applied to_ a magnitude, wouldn't you say?

wslooney:
@rjyounes It's a very minor point and I trust yours and Michael's instincts on what is or isn't misleading in an annotation. Given what you've said, I think what we have is good.

It's an interesting question whether or not you can multiply an inch. _Prima facie_ I would have thought you could because units like quarter-inches or inches squared suggest we do, but it's above my proverbial paygrade.

rjyounes:
Interesting point. Nevertheless, in the case of conversion factor we are not multiplying the unit, but the magnitude of _n_ inches.

PR #662, opened 2022-04-21
Fixes #661: update scopenote with isIdentifiedBy

### LINKED ISSUE #661: Update reference to identifiedBy
There is a scopeNote on gist:name that says:

> "For more formal use, consider using a sub property of the object property, identifiedBy."

That property was renamed to isIdentifiedBy. Update to:

"For more formal use, consider using a sub property of the object property, isIdentifiedBy."

Jamie-SA:
Fixes #661
Update scopenote on gist:name to use the correct property name that was previously updated.


uscholdm (approved):

rjyounes (approved):
I removed the scope note altogether since it was irrelevant to this property, and also fixed the property definition.

PR #667, opened 2022-05-06
Adding new property `hasFirstMember`. Resolves #549.

### LINKED ISSUE #549: Open World makes first thing in an OrderedCollection indeterminate
Given the open world assumption, it is not possible to tell which is the first thing in an OrderedCollection just by looking for the absence of a predecessor.   We may want an explicit `hasFirstMember` type property to remove this ambiguity and make it easy to find the first member of an OrderedCollection.

dylan-sa:


gistCore.ttl
```diff
@@ -3227,6 +3227,16 @@ gist:hasDirectSuperCategory
...
+gist:hasFirstMember
+	a owl:ObjectProperty ;
+	rdfs:subPropertyOf gist:hasMember ;
+	rdfs:domain gist:OrderedCollection ;
+	rdfs:range gist:OrderedMember ;
+	skos:definition "Relates an ordered collection to its first member."^^xsd:string ;
+	skos:prefLabel "has first member"^^xsd:string ;
+	skos:scopeNote "Given the open-world assumption, the absence of a predecessor does not entail that an ordered member is the first member of an ordered collection. This property is used to explicitly indicate the first member."^^xsd:string ;
+	.
+
```
rjyounes, line 3239: This looks nearly perfect. We didn't mention this at the meeting, but it occurs to me you could also add a restriction on `OrderedCollection` to the effect that it is either empty or has a first member. (Note that due to OWA the first member doesn't have to be stated.)

A little nitpicky, but most commonly I see the OWA spelled "Open World Assumption" rather than the hyphenated lowercase version, though I know this is also used.

dylan-sa: For `OrderedCollection`, maybe something like the following would work? If so, should I go ahead and include in this PR?

```
gist:OrderedCollection
	a owl:Class ;
	owl:equivalentClass [
		a owl:Class ;
		owl:intersectionOf (
			gist:Collection
			[
				a owl:Restriction ;
				owl:onProperty gist:hasMember ;
				owl:allValuesFrom gist:OrderedMember ;
			]
			[
				owl:unionOf (
					[
						owl:complementOf [
							a owl:Restriction ;
							owl:onProperty gist:hasMember ;
							owl:someValuesFrom gist:OrderedMember ;
						] ;
					]
					[
						a owl:Restriction ;
						owl:onProperty gist:hasFirstMember ;
						owl:someValuesFrom gist:OrderedMember ;
					]
				) ;
			]
		) ;
	] ;
```

uscholdm (approved):

rjyounes:
The complement is not correct: you need to use owl:Thing instead of gist:OrderedMember to say that the set has no members. Otherwise the complement includes things which have objects other than OrderedMembers as members.

rjyounes (changes requested):
See comments.

dylan-sa:
@rjyounes Ah, that's right! I will go ahead and work the correct version into the PR.

gistCore.ttl
uscholdm: I'm not sure I follow; it looks like it is literally saying the following:
```
skos:definition "A collection that only has ordered members and either has a first member that is an ordered member or has no members that are ordered members."^^xsd:string ; 
```

Is this your intention? A slightly simpler alternative to complement would be to use max cardinality of 0

uscholdm:
> Which makes me realize that `hasFirstMember` should be declared a functional property.

good point.

dylan-sa:
I have no objection to going the functional route, but in #549 it looks like we wanted to allow for "first-place ties," so to speak.

rjyounes:
Good point, you're right, we do want first place ties, so not functional.

rjyounes:
But it now occurs to me that it's an _inverse_ functional property, since each OrderedMember belongs to a unique OrderedCollection and points to a unique object.

uscholdm:
I think that's right, given an ordered member is just the proxy for the real thing.

dylan-sa:
OK, I think the latest two commits get it right. `hasFirstMember` is now inverse functional, and I've tweaked the restriction for `OrderedMember` in line with discussion above. (Namely, I went with the 'cardinality 0' construction over the clunky 'complement' construction.) If this looks good, I will rebase to catch up with the latest changes to develop.

rjyounes (changes requested):
Everything you've done is perfect. Can you add one sentence to the scope note along the lines of "Since ordered collections may not be strictly ordered, there can be more than one first member." 

After that it's good to go.

PR #670, opened 2022-05-13
Update annotations on PhysicalSubstance and PhysicalIdentifiableItem. Fixes #644.

### LINKED ISSUE #644: Fix definition of `PhysicalSubstance`
Current definition and examples:

```
skos:definition "Non corporeal material, i.e. 'stuff' which can be divided into parts where each part retains its essence."
skos:example "An amount of water, of penicillin, of sand, of gold. An actual piece of gold, not gold the concept."
```

@mkumba notes that 'non-corporeal' is incorrect, and should be changed to 'physical.'

`piece of gold ` is not correct as an example, it should be `some amount of gold`.

The distinction between `PhysicalSubstance` and `PhysicalIdentifiableItem` mirrors English mass and count nouns. One could also use the term `discrete` as in `DiscretePhysicalItem`, but I don't know what the opposite is.

However, it doesn't seem correct to say that count nouns like persons and a grain of sand lose their identity when divided. A person can have all his/her limbs amputated and still be a person. This needs clarification. There's a question of the granularity of the division - if a person is split into distinct atoms, then it is no longer a person.

rjyounes:
In addition to updating definitions and scope notes of these classes, I have removed the restriction on PhysicalIdentifiableItem that it has an ID. While, as the scope note indicates, PIIs are able to have RFIDs, they do not necessarily have any sort of identifier.


rjyounes:
@uscholdm Please note the comment in the description above.

gistCore.ttl
```diff
@@ -1877,13 +1872,21 @@ gist:PhysicalIdentifiableItem
 		gist:SchemaMetaData ,
 		gist:UnitOfMeasure
 		;
-	skos:definition "You could, at least in principle, put an RFID tag on members of this class. Physical things are made of something.  E.g., statues are made of bronze."^^xsd:string ;
+	skos:definition
+		"A discrete physical object which, if subdivided, will result in parts that are distinguishable in nature from the whole and in general also distinguishable from the other parts."^^xsd:string ,
```
uscholdm, line 1876: Say also here that it has weight and takes up space. Maybe:

`A  physical object that  has weight and takes up space and if subdivided, will result in parts that are distinguishable`...

or

`A  physical object (i.e. has weight and takes up space) that if subdivided, will result in parts that are distinguishable`...

gistCore.ttl
```diff
@@ -1901,12 +1904,12 @@ gist:PhysicalSubstance
 		]
 		;
 	owl:disjointWith gist:UnitOfMeasure ;
-	skos:definition "Non-corporeal material, i.e. 'stuff' which can be divided into parts where each part retains its essence."^^xsd:string ;
+	skos:definition "An undifferentiated amount of physical material which, when subdivided, results in each part being indistinguishable in nature from the whole and from every other part."^^xsd:string ;
```
uscholdm, line 1907: Shouldn't `material which` be `material that`?

gistCore.ttl
```diff
@@ -3704,9 +3707,8 @@ gist:longitude
 gist:name
 	a owl:DatatypeProperty ;
 	rdfs:range xsd:string ;
-	skos:definition "Relates an individual to a casual name."^^xsd:string ;
+	skos:definition "Relates an individual to (one of) its name(s)."^^xsd:string ;
```
uscholdm, line 3710: This is ok, but is unrelated to the PR topic - are we just tossing it in here for convenience?

uscholdm (changes requested):
Getting very close...

rjyounes:
Closing; this PR was meant to be merged to develop, not master.

PR #672, opened 2022-05-13
Update annotations on gist:uniqueText. Fixes #577.

### LINKED ISSUE #577: Broaden gist:uniqueText definition
Definition is: "This is used for the actual value of a key or ID where you don't want the possibility of having more than one."

There is no domain.

Proposal:
- add `gist:domainIncludes gist:ID`
- Change definition to something like: 'The unique string value of some text object."
- Add `skos:example`: "The string value of an identifier object; the three-segment number associated with a materialized version number."

rjyounes:


rjyounes:
Closing. The PR needs to be submitted to develop, not master.

PR #673, opened 2022-05-13
Update definition of gist:Building. Fixes #482.

### LINKED ISSUE #482: Definition of gist:Building is too broad
Definition is "A man-made structure for dwelling or working."

According to this, construction scaffolds, parking lots (where a parking lot attendant works),  elevators, and food trucks, among many other things, are all buildings.

rjyounes:


gistCore.ttl
```diff
@@ -186,7 +186,11 @@ gist:Building
 		gist:Artifact ,
 		gist:Landmark
 		;
-	skos:definition "A man-made structure for dwelling or working."^^xsd:string ;
+	skos:definition "A relatively permanent man-made structure situated on a plot of land, having a roof and walls, commonly used for dwelling, entertaining, or working."^^xsd:string ;
+	skos:example
```
uscholdm, line 190: Maybe add yurt, which is a nice example of relatively permanant. Some are put up and taken down seasonally - but that still counts as relatively permanent, to me.

uscholdm (changes requested):

rjyounes:
I'm not sure why these other changes are showing up...

gistCore.ttl
rjyounes: I think yurt may be a controversial case, so I'm going to leave it out as subject to user interpretation.

rjyounes:
I made a bad mistake - I submitted a PR to merge into master rather than develop. This is the result of working with another client that doesn't use a develop branch and merges directly to master, so I'm in the habit of doing that. I need to undo the merge to master and fix things again.

I also forgot release notes...

rjyounes:
The merge to master has been undone.

PR #724, opened 2022-09-17
Normalize gist for protege + serializer compatibility. Fixes #646

### LINKED ISSUE #646: Does the serializer give different results on a Protégé-generated file?
We've observed some differences in serialization output from commit to commit, generally having to do with ordering of triples. One hypothesis is that the serializer is producing different results on Protégé-generated files and manually edited files (for reasons unknown). The task is to run an experiment to determine whether this is true, and if so, figure out what if anything can be done about it.

marksem:


marksem:
I initially included a separate commit adding skos declarations (which could be contentious anyway) but the validation checker complained that these were not properly annotated.  So we would need to relax our validation for such a case. 

Rather than change the validation check, I "zagged" and added code to the pre-commit hook to remove Protege's inserted SKOS declarations.

I believe this still achieves the goal of avoiding repeated, annoying, extra work for those who use Protege to edit.

HOWEVER, this solution still leaves us with the problem of Protege removing the skos:prefLabel on the ontology IRI.

uscholdm (approved):
Lets go with this for now, we can always change our minds later if we wish. Not hving skos:prefLabel for ontology is a minor nit - unlikely to cause real problems.

PR #737, opened 2022-09-27
Changes to precedes and follows predicates. Fixes #300.

### LINKED ISSUE #300: Change the definition of 'precedes' to exclude same order
From Phil Blackwell via email.

Definition of gist:precedes is "A generic ordering relation indicating that the Subject has the same order as or comes before the Object. The 'greater than or equal to' symbol is often used for this relation."

Phil notes that the relevant symbol is <= rather than >=.

I am now wondering whether it shouldn't be simply < rather than <=. The former aligns better with our common sense understanding. Then the first part of the definition must also change.

rjyounes:
This PR includes:

1. Changed the definition of `gist:precedes` to exclude same order. 
2. Added OWL type assertions to `gist:precedes`, `gist:precedesDirectly`, and `gist:followsDirectly` to formally express this (asymmetric and irreflexive).
3. Updated ontologyShapes.ttl to accept these OWL types.
4. Added scope notes to `gist:precedesDirectly` and `gist:followsDirectly` about how to express same order.
5. Added predicate `gist:follows`, since during this work it was found to be missing.
6. A few miscellaneous modifications to gistDeprecated.ttl and the release notes.

Item 5 is technically beyond the scope of issue #300, but given that it's a bug fix it seemed acceptable to add it here.  Item 6 is also outside the scope but address small omissions I noted along the way.

### DIFF
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -3033,8 +3033,25 @@ gist:endDateTime
 		;
 	.
 
+gist:follows
+	a
+		owl:ObjectProperty ,
+		owl:AsymmetricProperty ,
+		owl:IrreflexiveProperty ,
+		owl:TransitiveProperty
+		;
+	skos:definition "A generic ordering relation indicating that the subject comes after the object. The greater-than symbol is often used for this relation."^^xsd:string ;
+	skos:prefLabel "follows"^^xsd:string ;
+	skos:scopeNote "This is the transitive version of gist:followsDirectly."^^xsd:string ;
+	.
+
 gist:followsDirectly
-	a owl:ObjectProperty ;
+	a
+		owl:ObjectProperty ,
+		owl:AsymmetricProperty ,
+		owl:IrreflexiveProperty
+		;
+	rdfs:subPropertyOf gist:follows ;
 	owl:inverseOf gist:precedesDirectly ;
 	skos:definition "A generic ordering relation indicating that the subject comes immediately after the object."^^xsd:string ;
 	skos:prefLabel "follows directly"^^xsd:string ;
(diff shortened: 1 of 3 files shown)


gistCore.ttl
```diff
@@ -3029,12 +3029,32 @@ gist:endDateTime
...
 	.
 
+gist:follows
+	a
+		owl:ObjectProperty ,
+		owl:AsymmetricProperty ,
+		owl:IrreflexiveProperty ,
+		owl:TransitiveProperty
+		;
+	skos:definition "A generic ordering relation indicating that the subject comes after the object. The greater-than symbol is often used for this relation."^^xsd:string ;
```
dylan-sa, line 3039: The part about the greater-than symbol might be better as `scopeNote`s, as it's more about usage. Ditto for the `precedes` definition.

dylan-sa (approved):
Had one thought about moving the greater-than/less-than comments into `scopeNote`s, but I can see it going either way. Otherwise looks good to me!

gistCore.ttl
rjyounes: Yes, I like that idea and have made that change.

uscholdm (changes requested):
Everyting looks good except that the release notes fail to mention the new instances of IrreflexivePRoperty and AsymmetricProperty. It only says annotations are changed, which is less central.

rjyounes:
@uscholdm In addressing your comment in the release notes I realized that adding those types impacts reasoning and therefore is technically a major change. Do you think I should roll that part out of this PR and save it for the next major release?

uscholdm (approved):
good

rjyounes:
Backed out the assertion of asymmetricity and irreflexivity on the predicates.

PR #739, opened 2022-09-27
Add prefix declarations. Fixes #684.

### LINKED ISSUE #684: Should `gist` describe its preferred prefix and namespace URI using VANN?
It's become very common practice in more recent years for ontologies to explicitly provide simple metadata for their **_preferred_** prefix and namespace URI using the long-standing, and stable, [VANN vocabulary](http://purl.org/vocab/vann/).

For gist, this would simply be including the following two new VANN triples:
```
prefix vann: <http://purl.org/vocab/vann/>

<https://ontologies.semanticarts.com/o/gistCore> a owl:Ontology;
    vann:preferredNamespacePrefix "gist" ;
    vann:preferredNamespaceUri "https://ontologies.semanticarts.com/gist/" .
```

**Note**: In the case of the prefix, this is just potentially helpful extra metadata - i.e., by explicitly stating a _preferred_ value it can help maintain a higher degree of consistency among people using that ontology.

**Note**: In the case of the namespace URI, this can be helpful in linking the ontology resource (i.e., `<https://ontologies.semanticarts.com/o/gistCore>` in the case of gist) back to the actual namespace URI it describes (i.e., `"https://ontologies.semanticarts.com/gist/"`).

**Note**: The `rdfs:range` of the VANN `preferredNamespaceUri` predicate (although not explicitly stated in the VANN documentation) is accepted to be an `xsd:string`, and not an IRI, or `xsd:anyURI`.
I think this is deliberate to allow processors perform simple string manipulation on the IRIs of terms from a vocabulary, but also because to-date the notion of a Namespace has not been explicitly denoted (i.e., there are no `rdf:Namespace` or `owl:Namespace` types, and no associated `rdf:namespace` or `owl:hasNamespace` predicates), and so dereferencing a Namespace URI in-and-of-itself has generally (but not always!) been expected to return the description of that ontology - but that description has never been expected to contain triples describing the Namespace itself (whether it **_should or not_** is a separate issue :) !).

rjyounes:
In modifying bundle.yaml I noticed a few small errors in imports and preflabels in the non-core ontology files, and have fixed those as well though not directly related to this issue.

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -62,11 +62,7 @@ actions:
   inference: "none"
   source: "{input}"
   includes:
-    - gistCore.ttl
-    - gistDeprecated.ttl
-    - gistMediaTypes.ttl
-    - gistPrefixDeclarations.ttl
-    - gistValidationAnnotations.ttl
+    - "gist*.ttl"
   shapes:
     source: "{input}/ontologyShapes.ttl"
 - action: "verify"
@@ -95,9 +91,7 @@ actions:
     from: "(.*)\\.ttl"
     to: "\\g<1>{version}.ttl"
   includes:
-    - gistCore.ttl
-    - gistMediaTypes.ttl
-    - gistDeprecated.ttl
+    - gist*.ttl
 - action: "definedBy"
   message: "Adding rdfs:definedBy."
   source: "{output}"
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -63,7 +63,9 @@ actions:
   source: "{input}"
   includes:
     - gistCore.ttl
+    - gistDeprecated.ttl
     - gistMediaTypes.ttl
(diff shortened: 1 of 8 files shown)


Jamie-SA (approved):
Looks correct.

Jamie-SA:
@rjyounes I am trying to figure out the commit history on this and other PRs of yours. Above it looks like the commits were done this morning, but I reviewed it yesterday. Normally, I wouldn't want to see commits happen after an approval without a re-review but from memory those commits seem to be the same things I reviewed yesterday. This isn't the first time I've seen this on PRs you've merged so I'm thinking you are doing something different than I am used to seeing and wondering what that is.

rjyounes:
I am doing a rebase rather than a merge from the develop, so something must be going wrong there. I'll switch to merges to see if that helps.

PR #742, opened 2022-09-29
Deprecated group and added Deprecated ttl file. Fixes #663

### LINKED ISSUE #663: Deprecate gist:Group
from @uscholdm (https://github.com/semanticarts/gist/issues/570#issuecomment-1104428832) where we were looking at `gist:Collection` and subclasses:

"The use of the word 'grouping' in the definition conflicts with gist:Group. I propose to get rid of gist:Group. I have never used it. If we did want it, it should have a more specific name, such as GroupOfPersons. It would be defined as a subclass of gist:Collection and have an allValuesFRom restirciton with gist:Person as the filter class. But I cannot see any reason to keep it."

Should Group be deprecated at next minor release?

wslooney:


gistCore.ttl
```diff
@@ -1137,6 +1137,7 @@ gist:GovernmentOrganization
 
 gist:Group
 	a owl:Class ;
+	owl:deprecated "true"^^xsd:boolean ;
```
rjyounes, line 1140: @wslooney The class needs to be removed from gistCore. The `owl:deprecated` assertion should be added to the definition in gistDeprecated.

rjyounes (changes requested):
In addition to the inline comment, you need to add a release note. This is a minor change.

rjyounes (changes requested):
It probably wasn't clear, the `gistDeprecated.ttl` file goes at the top level of the directlry, along with `gistCore.ttl. The Deprecated folder is created only by the release bundle.

Otherwise, your latest changes are good, and you still need a release note.

wslooney:
@rjyounes Thanks. I believe I have made the appropriate changes. Please check when you get the chance.

rjyounes (approved):

PR #743, opened 2022-09-29
Allow for empty collections, changed annotations (issue 570)

wslooney:


gistCore.ttl
```diff
@@ -339,14 +339,10 @@ gist:CoherentUnit
...
 	a owl:Class ;
-	rdfs:subClassOf [
-		a owl:Restriction ;
-		owl:onProperty gist:hasMember ;
-		owl:someValuesFrom owl:Thing ;
-	] ;
-	skos:definition "Any identifiable grouping of instances.  For instance, a jury is a collection of people."^^xsd:string ;
-	skos:example "A jury is a group of people,   a financial ledger is a collection of transaction entries; a route is an (ordered) collection of segments."^^xsd:string ;
+	skos:definition "A grouping of things."^^xsd:string ;
+	skos:example "A jury is a group of people, a financial ledger is a collection of transaction entries; a route is an (ordered) collection of segments. A deck of playing cards."^^xsd:string ;
```
uscholdm, line 343: Suggest remove deck of playing cards, unlikely to be relevant to an enterprise client.

uscholdm (approved):
One minor suggestion, but I'm approving anyway. So you can make the change and merge w/o another review.

rjyounes (changes requested):
You need to add a release note.

rjyounes (changes requested):
The branch is out of sync with develop, so you also need to merge in develop. I would do that before writing the release note, because you may get conflicts in that file without the merge.

gistCore.ttl
uscholdm: There already is such a note:

```This definition allows a collection to have zero members.```

rjyounes (approved):

PR #747, opened 2022-10-04
Update list of documents in bundle.yaml

### LINKED ISSUE #746: A couple issues with the bundle.yaml file
I noticed a couple of problems with but bundle.yaml file that should be fixed. Check the PR for specifics.

Jamie-SA:
Fixes #746 

Mainly the list of document files was outdated.

I noticed that you have to explicitly list all markdown files that you want converted to HTML. It would be better to not have to keep that list updated every release but I think that would require leaving them as .md files or making a change to onto-tool. Maybe this should be added to the release documentation as a step (if it hasn't been added previously).

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -185,11 +185,11 @@ actions:
     from: "\\(./(\\w+)\\.md\\)"
     to: "(./\\g<1>.html)"
   includes:
+    - "Contributing.md"
+    - "gistStyleGuide.md"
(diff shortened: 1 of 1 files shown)


rjyounes:
@Jamie-SA I'm going to close this PR. We will consider issue #746 for the next release. This PR may be updated by then, so I think it's easier to start fresh next time around.

PR #751, opened 2022-10-06
Final Changes for Release 11.1.0

stevenchalem:


Jamie-SA (approved):
Looks good to me... but I'm not the grammar police :)

docs/ReleaseNotes.md
```diff
@@ -1,37 +1,38 @@
...
 - Improved the clarity and accuracy of annotations on several terms, including definitions, examples, and scope notes:
-
   - `gist:Building`: Issue [#482](https://github.com/semanticarts/gist/issues/482).
   - `gist:PhysicalSubstance` and `gist:PhysicalIdentifiableItem`: Issue [#644](https://github.com/semanticarts/gist/issues/644).
   - `gist:uniqueText`: Issue[#577](https://github.com/semanticarts/gist/issues/577).
   - `gist:precedes`, `gist:precedesDirectly`, and `gist:followsDirectly`. Issue [#300](https://github.com/semanticarts/gist/issues/300).
   - `gist:Template` [#408](https://github.com/semanticarts/gist/issues/408).
 - Added SHACL `PrefixDeclaration`s for prefixes used in the gist ontology. The declaration for `gist:` is in the gist core ontology, and additional prefixes are included in a separate ontology file. Issue [#684](https://github.com/semanticarts/gist/issues/684).
 - Reformatted gist for Protege + serializer compatibility. Issue [#646](https://github.com/semanticarts/gist/issues/646).
+- Updated gist logo. [#713](https://github.com/semanticarts/gist/issues/713).
```
rjyounes, line 33: Lines 33 and 34 are only relevant internally, which is why I didn't add them to the release notes. They should be removed. I'm also tempted to say the same of line 32 - what do you think?

docs/ReleaseNotes.md
```diff
@@ -1,37 +1,38 @@
-gist Release Notes
+Gist Release Notes
```
rjyounes, line 1: gist is always written all lower-case.

rjyounes (changes requested):

docs/ReleaseNotes.md
justin2004: > gist is always written all lower-case.

That's like building a staircase where one step is one inch higher than all the rest.

rjyounes (approved):

PR #761, opened 2022-10-17
Update gistStyleGuide.md

pmcb55:
Just some pedantic suggestions (discard if you like!). Main 'contention' might be over my suggestion to replace 'Orthographic' with just 'Spelling' (i.e., I had to Google 'Orthographic', as I did initially think "What has gist got to do with 'map stuff'!?" :) !). I have a few more thoughts on this Style Guide, but this is just an initial attempt to suggest some 'low hanging fruit', as I see it - but in general, I really, *really* like (and agree with) the vast majority of it :) !

### DIFF
--- a/docs/gistStyleGuide.md
+++ b/docs/gistStyleGuide.md
@@ -48,17 +48,19 @@ Serialization
 Local Names
 -----
 
-### Orthographic Conventions for Class and Property Local Names
+### Spelling Conventions for Class and Property Local Names
 
 -----
 
 - Camelcase
-  - Classes initial uppercase
-  - Properties initial lowercase
+  - Classes initial uppercase.
+    - Example: `GeoPoint` not `geopoint` or `geoPoint`.
+  - Properties initial lowercase.
+    - Example: `hasDirectPart` not `HasDirectPart` or `hasdirectpart` or `has_direct_part`.
 - Alphanumeric characters only.
   - Example: `Isbn10`, not `Isbn-10` or `ISBN-10`.
 - Acronyms are also camelcased so that word boundaries are unambiguous.
-  - Examples: `AmaGuideline`, not `AMAGuideline`; `UriScheme`, not `URIScheme`
+  - Examples: `AmaGuideline`, not `AMAGuideline`; `UriScheme`, not `URIScheme`.
   - `ID` is an exception, because Merriam-Webster spells it in all-caps.
 - No non-standard abbreviations. E.g., `hasUoM` should be `hasUnitOfMeasure`.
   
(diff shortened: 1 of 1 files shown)


uscholdm:
Phil, it would be helpful if you summarized the changes at a high level with rationale. Such as,
- Added periods (they are not sentences, so why do you want periods?)
-

docs/gistStyleGuide.md
```diff
@@ -128,15 +130,15 @@ The following conventions apply to `skos:prefLabel` but not `skos:altLabel`.
 
 ### Classes
 
-- Title case (see definition of title case below)
+- Title case (see definition of title case below).
 - Normalized to natural language standards. E.g., hyphens inserted, acronyms in all caps, etc.
-  - Examples: *AMA Guideline*, *ISBN-10*
+  - Examples: `*AMA Guideline*`, `*ISBN-10*`.
```
uscholdm, line 135: Why this change?

docs/gistStyleGuide.md
```diff
@@ -48,30 +48,32 @@ Serialization
...
-  - Properties initial lowercase
+  - Classes initial uppercase.
+    - Example: `GeoPoint` not `geopoint` or `geoPoint`.
+  - Properties initial lowercase.
+    - Example: `hasDirectPart` not `HasDirectPart` or `hasdirectpart` or `has_direct_part`.
 - Alphanumeric characters only.
   - Example: `Isbn10`, not `Isbn-10` or `ISBN-10`.
 - Acronyms are also camelcased so that word boundaries are unambiguous.
-  - Examples: `AmaGuideline`, not `AMAGuideline`; `UriScheme`, not `URIScheme`
+  - Examples: `AmaGuideline`, not `AMAGuideline`; `UriScheme`, not `URIScheme`.
```
uscholdm, line 63: Why adding periods to non-sentences?

docs/gistStyleGuide.md
rjyounes: Agree, these are natural language terms and should not be in backticks.

docs/gistStyleGuide.md
rjyounes: Agree with Michael.

rjyounes:
Re "spelling" vs "orthography": Orthography is largely concerned with matters of spelling, and in particular the relationship between phonemes and graphemes in a language. Other elements that may be considered part of orthography include hyphenation, capitalization, word breaks, emphasis, and punctuation. (Wikipedia)

I.e., spelling has to do merely with the arrangement of letters in a word; orthography is broader than that. I stick by my word choice.

docs/gistStyleGuide.md
uscholdm: On the other hand, this text will be in OWL code this way in an annotation: `"AMA Guideline"^^xsd:string  `. May not want the *'s which causes italics.

rjyounes: Agreed - italics should be removed.

docs/gistStyleGuide.md
rjyounes: I see now there's a period on the line above. The main thing is to be consistent throughout the document.

Issue #770, opened 2022-11-02
should `gist:isCharacterizedAs` be a subprop of `gist:isCategorizedBy`?


justin2004:
I think `gist:isCharacterizedAs` is a specialization of  `gist:isCategorizedBy` where the domain and range are fixed.
Whereas `gist:isCategorizedBy` doesn't fix the domain and range.

uscholdm:
> I think `gist:isCharacterizedAs` is a specialization of `gist:isCategorizedBy`

If anything, it is the reverse. One way to characterize something is to say what category it is in. There are possibly many other ways. However, we might not need `gist:isCharacterizedAs` - I've never found a use case for it.  

Ironically, `gist:isCharacterizedAs`  is uncharacteristic of most gist properties, by having a very narrow domain and range. That goes against our guidelines.  If we keep  `gist:isCharacterizedAs` I hitnk it shoujld have no domain  or range and should be a supeproperty of `gist:isCategorizedBy`.

justin2004:
> I think it should have no domain or range and should be a supeproperty of gist:isCategorizedBy.

I'd be happy with that as well, @uscholdm .
I think they likely belong in the same prop hierarchy (not top level siblings). 

I was just using the presence of domain and range to be a cue that this is a specialization. Also `isCharacterizedAs` has "categorize" in its `skos:defintion`.

Also the `skos:definition` on `gist:isCharacterizedAs` seems to be backwards:
"A way to categorize a behavior."
While the domain is `gist:Event` and the range is `gist:Behavior`.
So I'f say that the Event is categorized by the Behavior -- the Behavior isn't being categorized.

rjyounes:
I agree, the current definition seems a complete anomaly. I'd be in favor of dropping it, as you initially suggest, @uscholdm. What is the use case when `isCategorizedBy` is not appropriate?

uscholdm:
> Also the `skos:definition` on `gist:isCharacterizedAs` seems to be backwards:
> "A way to categorize a behavior."

That's true. This is an ancient definition, needs to be updated if we keep the property.

uscholdm:
PROPOSAL: delete the property `isCharacterizedAs` because it is historical and was never used, to my knowledge. The narrow domain and range are also historical, we don't do that these days. The main purpose it serves is to enable Event to have some axioms.  That is not a super-compelling reason.

```
gist:Event
	a owl:Class ;
	owl:equivalentClass [
		a owl:Restriction ;
[...]

PR #777, opened 2022-11-21
Rework Event and Task classes. Fixes #760.

### LINKED ISSUE #760: Fix Project and Task
In gist 11.1.0 we have the following:

* `TaskExecution`, subclass of `Event`, defined as "An event characterized by a defined piece of work that is either scheduled, accomplished, or both." This is identical to gist 11.0.0 `Task`.
* `ProjectExecution`, subclass of `TaskExecution`, defined as "A project execution is a task execution (usually a longer duration task) made up of other task executions."
* `ScheduledTaskExecution`, subclass of `TaskExecution`, defined as "The execution of a task that was planned to occur at a given time. When it was scheduled, it would have been in the future, but now might be in the past."

We also have `TaskTemplate`, subclass of `Template`, defined as "An outline of a task of a particular type, which is the basis for executing such tasks." This is roughly like a process specification - a recipe or set of instructions.

My interpretation of `TaskTemplate` is that it's not meant to be a one-off - like it's superclass `Template`, it should be able to serve as the basis for a number of outputs. E.g., a recipe for chocolate cake can be the basis for a number of chocolate-cake making events that results in different chocolate cakes.

@mkumba asks: "If we define a Project[Execution], say the Juniper project, is that not a Project[Execution] either until it gets scheduled?"

I think we need a third type of object here, `DefinedTask` and `DefinedProject`. `Task` and `Project` are probably the preferred terms if starting from scratch, but that's not currently possible because they have been deprecated and replaced by `TaskExecution` and `ProjectExecution`.

`TaskExecution` would be related to a `Task` - use `isBasedOn`? Similarly for `ProjectExecution`. 

In a similar vein, @mkumba writes:

"Something got lost along the way.

A task template is a cookie cutter.  It is something you create tasks from.  In Pipedrive or Jira or a project management system, or a patient scheduling system, or a calendar, when you say “new” or “create” you create a task, and the nature and that task is determined by the template.  But the template isn’t a task or anything you can do, its a template. 

Used to be in gist we had three stages a task went through.  It went from being defined, to being scheduled to being completed.  So a defined task is like a Jira ticket in the backlog, we know what it is and what its about we just haven’t scheduled it yet.  We can then schedule it.  In some domains there is another step to resource it, and then finally we do it.  In most domains the task accretes more dates, resources and activity, but in some domains it makes sense to create a separate instance for the actual task (execution in 11.1 parlance) and refer to the schedule, but a lot of domains the task just evolves.

Where I think this thing drifted was maybe:
Somewhere along the way we dropped the concept “definedTask” and just said that’s just a task that hasn’t yet been scheduled or accomplished.
We tightened the definition of event, to where now it has to have and start and end date (we were kind of loose historically with some values, yeah this task (say a Jira ticket) may not have a start date yet, but it will eventually.  Weirdly the current definition should create an inconsistency, (it says an event has exactly 1 endDate, but one of its children, contemporaneousEvent is max 0)  I think its not throwing an exception because exactly 1 rarely does any inference.

My suggestion is we relax event to include things that happened, are scheduled to happened, or we expect to schedule or something like that."

rjyounes:
Fixes #760.

### DIFF
--- a/docs/release_notes/PR777.md
+++ b/docs/release_notes/PR777.md
@@ -1,8 +1,8 @@
 ### Minor Updates
 
 - Changes related to tasks, projects, and events. Issue [#760](https://github.com/semanticarts/gist/issues/760).
-  - Deprecated `gist:TaskExeuction`, `gist:ScheduledTaskExecution`, and `gist:ProjectExecution` and renamed to `gist:Task`, `gist:ScheduledTask`, and `gist:Project`, respectively.
+  - Renamed `gist:TaskExecution`, `gist:ScheduledTaskExecution`, and `gist:ProjectExecution` and renamed to `gist:Task`, `gist:ScheduledTask`, and `gist:Project`, respectively.
   - These classes are no longer limited to the execution of a task or project, but include proposed and planned events.
-  - Removed datetime-related restrictions on `gists:Event`.
-  - Deprecated `gist:PlannedEvent` and renamed to `gist:ScheduledEvent`.
+  - Removed datetime restrictions on `gist:Event`.
+  - Renamed `gist:PlannedEvent` to `gist:ScheduledEvent`.
   - Added `skos:scopeNote`s and `skos:example`s to clarify meaning and usage.
--- a/docs/release_notes/PR777.md
+++ b/docs/release_notes/PR777.md
@@ -1,0 +1,8 @@
+### Minor Updates
+
+- Changes related to tasks, projects, and events. Issue [#760](https://github.com/semanticarts/gist/issues/760).
+  - Deprecated `gist:TaskExeuction`, `gist:ScheduledTaskExecution`, and `gist:ProjectExecution` and renamed to `gist:Task`, `gist:ScheduledTask`, and `gist:Project`, respectively.
+  - These classes are no longer limited to the execution of a task or project, but include proposed and planned events.
+  - Removed datetime-related restrictions on `gists:Event`.
+  - Deprecated `gist:PlannedEvent` and renamed to `gist:ScheduledEvent`.
+  - Added `skos:scopeNote`s and `skos:example`s to clarify meaning and usage.
--- a/docs/release_notes/PR777.md
+++ b/docs/release_notes/PR777.md
@@ -1,7 +1,7 @@
 ### Minor Updates
 
 - Changes related to tasks, projects, and events. Issue [#760](https://github.com/semanticarts/gist/issues/760).
-  - Renamed `gist:TaskExecution`, `gist:ScheduledTaskExecution`, and `gist:ProjectExecution` and renamed to `gist:Task`, `gist:ScheduledTask`, and `gist:Project`, respectively.
+  - Renamed `gist:TaskExecution`, `gist:ScheduledTaskExecution`, and `gist:ProjectExecution` to `gist:Task`, `gist:ScheduledTask`, and `gist:Project`, respectively.
   - These classes are no longer limited to the execution of a task or project, but include proposed and planned events.
   - Removed datetime restrictions on `gist:Event`.
   - Replaced `gist:PlannedEvent` with `gist:ScheduledEvent`.
--- a/docs/release_notes/PR777.md
+++ b/docs/release_notes/PR777.md
@@ -4,5 +4,5 @@
   - Renamed `gist:TaskExecution`, `gist:ScheduledTaskExecution`, and `gist:ProjectExecution` and renamed to `gist:Task`, `gist:ScheduledTask`, and `gist:Project`, respectively.
   - These classes are no longer limited to the execution of a task or project, but include proposed and planned events.
   - Removed datetime restrictions on `gist:Event`.
-  - Renamed `gist:PlannedEvent` to `gist:ScheduledEvent`.
+  - Replaced `gist:PlannedEvent` with `gist:ScheduledEvent`.
   - Added `skos:scopeNote`s and `skos:example`s to clarify meaning and usage.
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -900,7 +900,7 @@ gist:Event
 	skos:definition "Something that occurs over a period of time, often characterized as an activity being carried out by some person, organization, or software application or brought about by natural forces."^^xsd:string ;
 	skos:example "A transaction, conference, baseball game, earthquake."^^xsd:string ;
 	skos:prefLabel "Event"^^xsd:string ;
-	skos:scopeNote "An event does not necessarily have either planned or actual start or end dates. For example, a conference can be in the planning phase without any dates selected, but is nevertheless an (unscheduled) event. The subclasses of Event state particular restrictions on planned and actual start and end dates."^^xsd:string ;
+	skos:scopeNote "An event does not necessarily have either planned or actual start or end datetimes. For example, a conference can be in the planning phase without any dates selected, but is nevertheless an (unscheduled) event. The subclasses of Event state particular restrictions on planned and actual start and end dates."^^xsd:string ;
 	.
 
 gist:Extent
@@ -2106,9 +2106,8 @@ gist:ScheduledEvent
 			]
 		) ;
 	] ;
-	skos:definition "An event which, at the time it is created, is to occur in the future. It remains a scheduled event after it has started or ended, but then is also either a historical or contemporary event as well."^^xsd:string ;
-	skos:prefLabel "Scheduled Event"^^xsd:string ;
-	skos:scopeNote "Not all planned events are scheduled events. For example, a conference can be in an early planning phase without any dates yet selected."^^xsd:string ;
+skos:definition "An event with a planned start datetime."^^xsd:string ;
+skos:scopeNote "If the event already started, but has not yet ended, it is a contemporary event with an actual start datetime. If the event is over, it is a historical event having an actual end datetime. The event always retains its planned start datetime, and thus continues to be a scheduled event."^^xsd:string ; 
 	.
 
 gist:ScheduledTask
@@ -2120,8 +2119,9 @@ gist:ScheduledTask
 			gist:Task
 		) ;
 	] ;
-	skos:definition "A task that was or is planned to occur at a given time. When it was scheduled, it would have been in the future, but might now be in the past."^^xsd:string ;
 	skos:prefLabel "Scheduled Task"^^xsd:string ;
+	skos:definition "A task with a planned start datetime."^^xsd:string ;
+skos:scopeNote "If work on the task has already started, but has not yet ended, it will have an actual start datetime. If the task is completed, it will also have an actual end datetime. The task always retains its planned start time, and thus continues to be a scheduled task."^^xsd:string ; 
 	.
 
 gist:SchemaMetaData
@@ -3481,7 +3481,11 @@ gist:isTriggeredBy
 	a owl:ObjectProperty ;
 	skos:definition "Relates a contingency, such as an event or obligation, to the event that gives rise to it."^^xsd:string ;
 	skos:prefLabel "is triggered by"^^xsd:string ;
-	skos:scopeNote "For obligations, this property describes what must happen to trigger the contingent obligation.  In most cases, before the contingent obligation becomes an actual obligation, the triggering event is scheduled to occur in the future, since if it had already happened, the contingency would no longer be contingent. Other uses include controls, processes, etc."^^xsd:string ;
+skos:scopeNote "For obligations, this property describes what must happen to trigger the contingent obligation. Other uses include controls, processes, etc."^^xsd:string ;
+	skos:example
+		"Fire insurance is contingent on a particular building burning down"^^xsd:string ,
+		"The death benefit payout on a life insurance policy following the death of a specific person."^^xsd:string
+		;  
 	.
 
 gist:isUnderJurisdictionOf
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -892,28 +892,18 @@ gist:EquipmentType
 
 gist:Event
 	a owl:Class ;
-	owl:equivalentClass [
-		a owl:Class ;
-		owl:intersectionOf (
-			[
-				a owl:Restriction ;
-				owl:onProperty gist:isCharacterizedAs ;
-				owl:someValuesFrom gist:Behavior ;
-			]
-			[
-				a owl:Restriction ;
-				owl:onProperty gist:endDateTime ;
-				owl:cardinality "1"^^xsd:nonNegativeInteger ;
-			]
-			[
-				a owl:Restriction ;
-				owl:onProperty gist:startDateTime ;
-				owl:cardinality "1"^^xsd:nonNegativeInteger ;
-			]
-		) ;
+	rdfs:subClassOf [
+		a owl:Restriction ;
+		owl:onProperty gist:isCharacterizedAs ;
+		owl:someValuesFrom gist:Behavior ;
 	] ;
-	skos:definition "Something that occurs over some period of time, often characterized as an activity being carried out by some person, organization, or software application."^^xsd:string ;
+	skos:definition "Something that occurs over some period of time, often characterized as an activity being carried out by some person, organization, or software application or brought about by natural forces."^^xsd:string ;
+	skos:example "Transaction, concert, baseball game, hurricane, earthquake."^^xsd:string ;
 	skos:prefLabel "Event"^^xsd:string ;
+	skos:scopeNote
+		"An event does not necessarily have either planned or actual start or end date.s For example, a wedding can be in the planning phase without any dates selected, but is nevertheless an event (though not a ScheduledEvent according to the formal definition of that class). The subclasses state particular restrictions on planned and actual start and end dates." ,
+		"Roughly equivalent to the Basic Formal Ontology (BFO) Process class."^^xsd:string
+		;
 	.
 
 gist:Extent
@@ -1919,28 +1909,6 @@ gist:Place
 	skos:prefLabel "Place"^^xsd:string ;
 	.
 
-gist:PlannedEvent
-	a owl:Class ;
-	owl:equivalentClass [
-		a owl:Class ;
-		owl:intersectionOf (
-			gist:Event
-			[
-				a owl:Restriction ;
-				owl:onProperty gist:plannedEndDateTime ;
-				owl:cardinality "1"^^xsd:nonNegativeInteger ;
-			]
-			[
-				a owl:Restriction ;
-				owl:onProperty gist:plannedStartDateTime ;
-				owl:cardinality "1"^^xsd:nonNegativeInteger ;
-			]
-		) ;
-	] ;
-	skos:definition "An event which, at the time it is created, is to occur in the future."^^xsd:string ;
-	skos:prefLabel "Planned Event"^^xsd:string ;
-	.
-
 gist:PostalAddress
 	a owl:Class ;
 	rdfs:subClassOf gist:Address ;
@@ -2014,21 +1982,22 @@ gist:ProductUnit
 	skos:scopeNote "A ProductUnit is intended have a value for conversionFactor."^^xsd:string ;
 	.
 
-gist:ProjectExecution
+gist:Project
 	a owl:Class ;
 	owl:equivalentClass [
 		a owl:Class ;
 		owl:intersectionOf (
-			gist:TaskExecution
+			gist:Task
 			[
 				a owl:Restriction ;
 				owl:onProperty gist:hasPart ;
-				owl:someValuesFrom gist:TaskExecution ;
+				owl:someValuesFrom gist:Task ;
 			]
 		) ;
 	] ;
-	skos:definition "A project execution is a task execution (usually a longer duration task) made up of other task executions."^^xsd:string ;
-	skos:prefLabel "Project Execution"^^xsd:string ;
+	skos:definition "A task, usually of longer duration, made up of other tasks."^^xsd:string ;
+	skos:example "Renovating a kitchen, writing a book, building an ontology."^^xsd:string ;
+	skos:prefLabel "Project"^^xsd:string ;
 	.
 
 gist:RatioMagnitude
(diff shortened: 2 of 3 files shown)


gistCore.ttl
```diff
@@ -886,28 +886,18 @@ gist:EquipmentType
...
+		owl:onProperty gist:isCharacterizedAs ;
+		owl:someValuesFrom gist:Behavior ;
 	] ;
-	skos:definition "Something that occurs over some period of time, often characterized as an activity being carried out by some person, organization, or software application."^^xsd:string ;
+	skos:definition "Something that occurs over some period of time, often characterized as an activity being carried out by some person, organization, or software application or brought about by natural forces."^^xsd:string ;
+	skos:example "Transaction, concert, baseball game, hurricane, earthquake."^^xsd:string ;
 	skos:prefLabel "Event"^^xsd:string ;
+	skos:scopeNote
+		"An event does not necessarily have either planned or actual start or end date.s For example, a wedding can be in the planning phase without any dates selected, but is nevertheless an event (though not a ScheduledEvent according to the formal definition of that class). The subclasses state particular restrictions on planned and actual start and end dates."^^xsd:string ,
+		"Roughly equivalent to the Basic Formal Ontology (BFO) Process class."^^xsd:string
```
dylan-sa, line 899: FWIW, I think `Event` is even closer to `Process` w/ the start and end date restrictions included. But even w/ the restrictions removed, it's still a very closely related concept. So this statement remains true.

philblackwood:
I reviewed and did not see any errors. 

Not entirely convinced ScheduledEvent is needed. Anyone needing to know if the event is scheduled can look for dates. It seems like including it just makes things more complex with no added value.

rjyounes:
@philblackwood That's a reasonable point. I would like to hang on to `ScheduledTask`, though, so that we can easily make that important distinction between proposed or planned events and those actually scheduled.

Let's see what @uscholdm thinks.

gistCore.ttl
```diff
@@ -886,28 +886,18 @@ gist:EquipmentType
...
-				owl:onProperty gist:endDateTime ;
-				owl:cardinality "1"^^xsd:nonNegativeInteger ;
-			]
-			[
-				a owl:Restriction ;
-				owl:onProperty gist:startDateTime ;
-				owl:cardinality "1"^^xsd:nonNegativeInteger ;
-			]
-		) ;
+	rdfs:subClassOf [
```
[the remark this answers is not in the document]
rjyounes: Good point. I'll also use minCardinality instead of someValuesFrom.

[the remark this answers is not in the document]
rjyounes: On second thought, I'm going to leave this as a subclass assertion. It seems odd to have just one object of an equivalence; we are probably going to remove `isCharacterizedAs` anyway (issue #770), and if not the domain and range need to be reconsidered.

gistCore.ttl
```diff
@@ -2274,8 +2266,9 @@ gist:TaskExecution
 			]
 		) ;
 	] ;
-	skos:definition "An event characterized by a defined piece of work that is either scheduled, accomplished, or both."^^xsd:string ;
-	skos:prefLabel "Task Execution"^^xsd:string ;
+	skos:definition "An event characterized by a defined piece of work that is either proposed, planned, scheduled, underway, or completed."^^xsd:string ;
+	skos:prefLabel "Task"^^xsd:string ;
+	skos:scopeNote "A thing that could potentially be executed which is merely described but not proposed in any specific way, such as a guidebook providing instructions for following hiking trails, is not a task but rather a task template."^^xsd:string ;
```
uscholdm, line 2271: REplace 
> such as a guidebook providing instructions for following hiking trails

with 

- such as a business process for onboarding a new employee or the steps in a recipe for making a batch of Head and Shoulders shampoo or Coco Cola.

HiIlight the fact that this is a negative example.  Maybe replace outright with:

```
skos:scopeNote "Negative examples include:  a business process for onboarding a new employee,  the steps in a recipe for making a batch of Coco Cola or Head and Shoulders shampoo.  These are not events. Instead, they prescribe how events should unfold to produce intended results.  It is important to distinguish is between the thing to be done (TaskTemplate) vs. the doing of the thing (Task/Event). "^^xsd:string ;

```

I wonder what we would call DNA replication. It just happens, like the grand canyon getting carved. We can only make up and backfill an imaginary goal/intention. Like Dawkin's selfish genes that 'want' to stay around.  If there was a conscious being intending of if there was not, there is a set of 'instructions' wound be identical in every case.

gistCore.ttl
```diff
@@ -2292,8 +2285,12 @@ gist:TaskTemplate
 		) ;
 	] ;
 	skos:definition "An outline of a task of a particular type, which is the basis for executing such tasks."^^xsd:string ;
+	skos:example "A recipe for making a cake."^^xsd:string ;
```
uscholdm, line 2288: Not a cake, how about shampoo or coca cola or a business process.

gistCore.ttl
```diff
@@ -2292,8 +2285,12 @@ gist:TaskTemplate
...
 	skos:prefLabel "Task Template"^^xsd:string ;
-	skos:scopeNote "The class TaskExecution is used to represent the execution of the task. Use the property isBasedOn to link the TaskExecution back to the TaskTemplate."^^xsd:string ;
+	skos:scopeNote
+		"A task template may define a single task or a series of tasks. For example, a recipe for baking a particular type of chocolate cake may consist of sub-templates such as cracking eggs, turning on the oven, mixing flour and sugar, etc. The level of granularity is variable according to use case: e.g., mixing flour and sugar can be further broken down into measuring each into a bowl and stirring."^^xsd:string ,
```
uscholdm, line 2291: The juxtapositon of 'sugar' and 'granularity' is amusing.

gistCore.ttl
```diff
@@ -3642,8 +3639,9 @@ gist:isSubTaskOf
 
 gist:isTriggeredBy
 	a owl:ObjectProperty ;
-	skos:definition "General causal relation.  For obligations a property that describes what would happen to trigger the contingent obligation.  In most cases, before the Contingent becomes an Obligation, the triggered by event is a planned event (that is it hasn't happened yet -- if it had happened the contingency would no longer be contingent.  In most cases it will be  a ContingentEvent .  Other uses include controls, processes etc"^^xsd:string ;
+	skos:definition "Relates a contingency, such as an event or obligation, to the event that gives rise to it."^^xsd:string ;
```
uscholdm, line 3642: This does not sound right - what exactly is a contingent event?

gistCore.ttl
```diff
@@ -3642,8 +3639,9 @@ gist:isSubTaskOf
...
 	skos:prefLabel "is triggered by"^^xsd:string ;
+	skos:scopeNote "For obligations, this property describes what must happen to trigger the contingent obligation.  In most cases, before the contingent obligation becomes an actual obligation, the triggering event is scheduled to occur in the future, since if it had already happened, the contingency would no longer be contingent. Other uses include controls, processes, etc."^^xsd:string ;
```
uscholdm, line 3644: > the triggering event is scheduled to occur in the future

Not true in the case of car theft insurance, or a house that burned down from lightnings strike. Those are triggering events for insurance obligations that are never scheduled (well maybe the crook planned to steal her neighbor's car, but that's not really a concern here.)

gistDeprecated.ttl
```diff
@@ -35,39 +35,62 @@ gist:Group
...
 	owl:deprecated "true"^^xsd:boolean ;
 	owl:equivalentClass [
 		a owl:Class ;
 		owl:intersectionOf (
-			gist:Task
+			gist:Event
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:plannedEndDateTime ;
+				owl:cardinality "1"^^xsd:nonNegativeInteger ;
```
uscholdm, line 48: End date not necessary something to be plan

gistDeprecated.ttl
```diff
@@ -35,39 +35,62 @@ gist:Group
 	skos:prefLabel "Group"^^xsd:string ;
 	.
 
-gist:Project
+gist:PlannedEvent
```
uscholdm, line 38: WHy is this here? The version of gistCore.ttl for this branch has no such class. Something is amiss.

uscholdm (changes requested):

gistCore.ttl
```diff
@@ -2274,8 +2266,9 @@ gist:TaskExecution
...
 		) ;
 	] ;
-	skos:definition "An event characterized by a defined piece of work that is either scheduled, accomplished, or both."^^xsd:string ;
-	skos:prefLabel "Task Execution"^^xsd:string ;
+	skos:definition "An event characterized by a defined piece of work that is either proposed, planned, scheduled, underway, or completed."^^xsd:string ;
+	skos:prefLabel "Task"^^xsd:string ;
+	skos:scopeNote "A thing that could potentially be executed which is merely described but not proposed in any specific way, such as a guidebook providing instructions for following hiking trails, is not a task but rather a task template."^^xsd:string ;
 	.
 
 gist:TaskTemplate
```
uscholdm, line 2274: It is important to distinguish is between the thing to be done (TaskTemplate) vs. the doing of the thing (Task/Event).  So a recipe is a task template, it has many steps. But eventually the steps bottom out into atomic steps that are not templates in any common-sense sense.  In the cake example, that might be 'sift flour'. Is 'sift flour' a task template?  There are countless process modeling formalisms out there that have names for these things. A common one  for 'the thing to do' is 'activity' (E.g. BPMN). An atomic activity in BPMN is called a Task (just to be confusing). A Business PRocess is a set/network of linked activities, each of which decomposes finally into Tasks. 

I do not yet have a good alternative term or concept to address my concern. A sleight of hand way to proceed is to simply say that an atomic template that no one would call a template is just the degenerate template, just like a point is a degenerate circle with radius = 0. Noone would call a point a circle, but it can be correctly viewed as one. So maybe 'sift flour' can be correctly viewed as a task template?

We already have something in gist that closely resembles this: `gist:Behavior`

mkumba:
To me it isn't any slight of hand, sifting flour is a template for the ten times I do it including last week during 1/2 cup of King Arthur and next week sifting 3/4 cup of Bob's red Mill. First one took me 2 minutes, second one 4.

Sifting is a behaviour





Sent from my iPhone

On Dec 1, 2022, at 11:58, Michael Uschold ***@***.***> wrote:

﻿

@uscholdm commented on this pull request.

________________________________

In gistCore.ttl<https://github.com/semanticarts/gist/pull/777#discussion_r1037597121>:

> @@ -2274,8 +2266,9 @@ gist:TaskExecution
                        ]
                ) ;
        ] ;
-       skos:definition "An event characterized by a defined piece of work that is either scheduled, accomplished, or both."^^xsd:string ;
-       skos:prefLabel "Task Execution"^^xsd:string ;
+       skos:definition "An event characterized by a defined piece of work that is either proposed, planned, scheduled, underway, or completed."^^xsd:string ;
+       skos:prefLabel "Task"^^xsd:string ;
+       skos:scopeNote "A thing that could potentially be executed which is merely described but not proposed in any specific way, such as a guidebook providing instructions for following hiking trails, is not a task but rather a task template."^^xsd:string ;
        .

 gist:TaskTemplate


It is important to distinguish is between the thing to be done (TaskTemplate) vs. the doing of the thing (Task/Event). So a recipe is a task template, it has many steps. But eventually the steps bottom out into atomic steps that are not templates in any common-sense sense. In the cake example, that might be 'sift flour'. Is 'sift flour' a task template? There are countless process modeling formalisms out there that have names for these things. A common one for 'the thing to do' is 'activity' (E.g. BPMN). An atomic activity in BPMN is called a Task (just to be confusing). A Business PRocess is a set/network of linked activities, each of which decomposes finally into Tasks.

I do not yet have a good alternative term or concept to address my concern. A sleight of hand way to proceed is to simply say that an atomic template that no one would call a template is just the degenerate template, just like a point is a degenerate circle with radius = 0. Noone would call a point a circle, but it can be correctly viewed as one. So maybe 'sift flour' can be correctly viewed as a task template?

We already have something in gist that closely resembles this: gist:Behavior

rjyounes:
Yes, echoing @mkumba, my thought was that "sifting flour" is a task template. There is no constraint on `TaskTemplate` regarding the level of granularity of the defined activity.

uscholdm:
> Yes, echoing @mkumba, my thought was that "sifting flour" is a task template. There is no constraint on `TaskTemplate` regarding the level of granularity of the defined activity.

Ok, that makes logical sense, just like it makes logical sense that a point is a circle of 0 radius - let's go with that for the moment.   I accept that 'sleight of hand' is not quite accurate.  Ignoring the possibility that it might cause confusion, there is still a question of what the difference is between an atomic `TaskTemplate` and a `Behavior`.  'Sifting' could be an instance of either or both.

gistCore.ttl
```diff
@@ -886,28 +886,18 @@ gist:EquipmentType
...
+		owl:onProperty gist:isCharacterizedAs ;
+		owl:minCardinality "1"^^xsd:integer ;
 	] ;
-	skos:definition "Something that occurs over some period of time, often characterized as an activity being carried out by some person, organization, or software application."^^xsd:string ;
+	skos:definition "Something that occurs over some period of time, often characterized as an activity being carried out by some person, organization, or software application or brought about by natural forces."^^xsd:string ;
+	skos:example "Transaction, concert, baseball game, hurricane, earthquake."^^xsd:string ;
 	skos:prefLabel "Event"^^xsd:string ;
+	skos:scopeNote
+		"An event does not necessarily have either planned or actual start or end dates. For example, a wedding can be in the planning phase without any dates selected, but is nevertheless an event (though not a ScheduledEvent according to the formal definition of that class). The subclasses state particular restrictions on planned and actual start and end dates."^^xsd:string ,
+		"Roughly equivalent to the Basic Formal Ontology (BFO) Process class."^^xsd:string
```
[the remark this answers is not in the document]
rjyounes: Removed.

gistCore.ttl
```diff
@@ -2117,17 +2086,40 @@ gist:Restriction
...
-gist:ScheduledTaskExecution
+gist:ScheduledEvent
+	a owl:Class ;
+	owl:equivalentClass [
+		a owl:Class ;
+		owl:intersectionOf (
+			gist:Event
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:plannedEndDateTime ;
```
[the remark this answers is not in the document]
rjyounes: Agreed.

gistCore.ttl
rjyounes: The only thing I did here was break up the definition into definition + scope note. If you want to revise them, you can add another issue.

gistDeprecated.ttl
rjyounes: Right, but when deprecating we don't change the definition of the term.

gistCore.ttl
rjyounes: I've made a small modification which I think addresses this.

gistDeprecated.ttl
rjyounes: We've replaced `PlannedEvent` with `ScheduledEvent`.

gistCore.ttl
```diff
@@ -2004,21 +1972,22 @@ gist:ProductUnit
...
 				owl:onProperty gist:hasSubTask ;
-				owl:someValuesFrom gist:TaskExecution ;
+				owl:someValuesFrom gist:Task ;
 			]
 		) ;
 	] ;
-	skos:definition "A project execution is a task execution (usually a longer duration task) made up of other task executions."^^xsd:string ;
-	skos:prefLabel "Project Execution"^^xsd:string ;
+	skos:definition "A task, usually of longer duration, made up of other tasks."^^xsd:string ;
+	skos:example "Renovating a kitchen, writing a book, building an ontology."^^xsd:string ;
```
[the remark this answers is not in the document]
rjyounes: Done.

gistCore.ttl
```diff
@@ -2117,17 +2086,40 @@ gist:Restriction
...
+gist:ScheduledEvent
+	a owl:Class ;
+	owl:equivalentClass [
+		a owl:Class ;
+		owl:intersectionOf (
+			gist:Event
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:plannedEndDateTime ;
+				owl:cardinality "1"^^xsd:nonNegativeInteger ;
```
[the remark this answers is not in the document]
rjyounes: When there is an exact cardinality, we lose precision with someValuesFrom. I agree in the case of minCardinality.

gistCore.ttl
```diff
@@ -2117,17 +2086,40 @@ gist:Restriction
...
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:plannedStartDateTime ;
+				owl:cardinality "1"^^xsd:nonNegativeInteger ;
+			]
+		) ;
+	] ;
+	skos:definition "An event which, at the time it is created, is to occur in the future. It remains a scheduled event after it has started or ended, but then is also either a historical or contemporary event as well."^^xsd:string ;
+	skos:prefLabel "Scheduled Event"^^xsd:string ;
+	skos:scopeNote "Not all planned events are scheduled events. For example, a wedding can be in an early planning phase without any dates yet selected."^^xsd:string ;
```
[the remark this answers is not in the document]
rjyounes: Replaced with conference.

gistCore.ttl
rjyounes: The only thing I did here was break up the definition into definition + scope note. If you want to revise them, you can add another issue.

rjyounes:
This PR has gone back into draft stage pending a decision about the next release.

gistCore.ttl
```diff
@@ -3501,8 +3493,9 @@ gist:isSupersededBy
 
 gist:isTriggeredBy
 	a owl:ObjectProperty ;
-	skos:definition "General causal relation.  For obligations a property that describes what would happen to trigger the contingent obligation.  In most cases, before the Contingent becomes an Obligation, the triggered by event is a planned event (that is it hasn't happened yet -- if it had happened the contingency would no longer be contingent.  In most cases it will be  a ContingentEvent .  Other uses include controls, processes etc"^^xsd:string ;
+	skos:definition "Relates a contingency, such as an event or obligation, to the event that gives rise to it."^^xsd:string ;
```
uscholdm, line 3496: Confusing. Suggest:
```
skos:definition "Relates something that is contingent on an event happening to that event. ^^xsd:string ;
```

gistCore.ttl
```diff
@@ -3501,8 +3493,9 @@ gist:isSupersededBy
...
 	skos:prefLabel "is triggered by"^^xsd:string ;
+	skos:scopeNote "For obligations, this property describes what must happen to trigger the contingent obligation.  In most cases, before the contingent obligation becomes an actual obligation, the triggering event is scheduled to occur in the future, since if it had already happened, the contingency would no longer be contingent. Other uses include controls, processes, etc."^^xsd:string ;
```
uscholdm, line 3498: > the triggering event is scheduled to occur in the future

Does not make sense to say the house burning down is a scheduled event, or even a planned event. It's really a specification of a hypothetical event. SKiip all the future stuff and replace with:
```
skos:scopeNote "For obligations, this property describes what must happen to trigger the contingent obligation. Other uses include controls, processes, etc."^^xsd:string ;
skos:example "A contingent obligation such as a fire insurance payout is triggered by one's house burning down."^^xsd:string ;  
```

gistCore.ttl
```diff
@@ -893,27 +893,14 @@ gist:EquipmentType
...
-		) ;
+		a owl:Restriction ;
+		owl:onProperty gist:isCharacterizedAs ;
+		owl:someValuesFrom gist:Behavior ;
 	] ;
-	skos:definition "Something that occurs over some period of time, often characterized as an activity being carried out by some person, organization, or software application."^^xsd:string ;
+	skos:definition "Something that occurs over a period of time, often characterized as an activity being carried out by some person, organization, or software application or brought about by natural forces."^^xsd:string ;
+	skos:example "A transaction, conference, baseball game, earthquake."^^xsd:string ;
 	skos:prefLabel "Event"^^xsd:string ;
+	skos:scopeNote "An event does not necessarily have either planned or actual start or end dates. For example, a conference can be in the planning phase without any dates selected, but is nevertheless an (unscheduled) event. The subclasses of Event state particular restrictions on planned and actual start and end dates."^^xsd:string ;
```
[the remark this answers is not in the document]
uscholdm: This also means that the restrictions on ContingentEvent are wrong.

uscholdm (changes requested):

gistCore.ttl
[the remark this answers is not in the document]
rjyounes: Agree with the first proposal.

The `ContingentEvent` restrictions have been fixed in another PR: https://github.com/semanticarts/gist/pull/826/.

gistCore.ttl
```diff
@@ -2127,17 +2093,35 @@ gist:Restriction
...
+		owl:intersectionOf (
+			gist:Event
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:plannedStartDateTime ;
+				owl:cardinality "1"^^xsd:nonNegativeInteger ;
+			]
+		) ;
+	] ;
+	skos:definition "An event which, at the time it is created, is to occur in the future. It remains a scheduled event after it has started or ended, but then is also either a historical or contemporary event as well."^^xsd:string ;
```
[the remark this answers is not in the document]
rjyounes: I like this. Adopted.

gistCore.ttl
```diff
@@ -2127,17 +2093,35 @@ gist:Restriction
...
 		owl:intersectionOf (
-			gist:PlannedEvent
-			gist:TaskExecution
+			gist:ScheduledEvent
+			gist:Task
 		) ;
 	] ;
-	skos:definition "The execution of a task that was planned to occur at a given time. When it was scheduled, it would have been in the future, but now might be in the past."^^xsd:string ;
-	skos:prefLabel "Scheduled Task Execution"^^xsd:string ;
+	skos:definition "A task that was or is planned to occur at a given time. When it was scheduled, it would have been in the future, but might now be in the past."^^xsd:string ;
```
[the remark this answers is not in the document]
rjyounes: I liked your language for `ScheduledEvent` so have adapted it for this term. See what you think.

gistCore.ttl
rjyounes: I agree that this is awkward. I've used the examples for `ContingentEvent`.

docs/release_notes/PR777.md
```diff
@@ -0,0 +1,8 @@
+### Minor Updates
+
+- Changes related to tasks, projects, and events. Issue [#760](https://github.com/semanticarts/gist/issues/760).
+  - Renamed `gist:TaskExecution`, `gist:ScheduledTaskExecution`, and `gist:ProjectExecution` and renamed to `gist:Task`, `gist:ScheduledTask`, and `gist:Project`, respectively.
```
dylan-sa, line 4: Should the first 'renamed' at the beginning of the sentence be 'removed'? Alternatively, could omit 'and renamed' from the middle of the sentence.

dylan-sa (changes requested):

docs/release_notes/PR777.md
rjyounes: Yes, thanks for catching that!

gistCore.ttl
uscholdm: I spend a lot of time on a good wording, not easy. Keep it as it is for now, make sure there is a good example, e.g. house burning down triggers insurance payout.

gistCore.ttl
```diff
@@ -2127,17 +2093,35 @@ gist:Restriction
...
+			[
+				a owl:Restriction ;
+				owl:onProperty gist:plannedStartDateTime ;
+				owl:cardinality "1"^^xsd:nonNegativeInteger ;
+			]
+		) ;
+	] ;
+	skos:definition "An event which, at the time it is created, is to occur in the future. It remains a scheduled event after it has started or ended, but then is also either a historical or contemporary event as well."^^xsd:string ;
+	skos:prefLabel "Scheduled Event"^^xsd:string ;
+	skos:scopeNote "Not all planned events are scheduled events. For example, a conference can be in an early planning phase without any dates yet selected."^^xsd:string ;
```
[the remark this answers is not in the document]
rjyounes: We had `PlannedEvent`, but based on the fact that events are no longer required to have dates, and that, as in the examples, something can be planned but not yet scheduled, we changed it to `ScheduledEvent`. I don't think planned would be the right name for such an event, because  it is too easily confused with scheduled event. We have a [new issue to add `HypotheticalEvent`](https://github.com/semanticarts/gist/issues/831), which is an event that cannot have dates.

PR #778, opened 2022-11-22
Add shacl and skos imports; gitignore catalog files

rjyounes:
Add SKOS and SHACL imports so Protege does not create stub definitions in exports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *_webDownload/
 .DS_Store
 *.orig
+/catalog-v001.xml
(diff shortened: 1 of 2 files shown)


rjyounes:
Closing due to @marksem's comments: Importing SHACL violates the OWL RL, DL, EL, and QL profiles because SHACL uses rdfs:Property  instead of OWL properties.  SKOS does the same, because it uses dc:creator and other dc properties without declaring them.

uscholdm:
Hmm - I routinely import SKOS for our client work, pertly to avoid Protege making up annotation properties.  Do I need to stop doing this?

rjyounes:
Apparently so.

Issue #105, opened 2019-09-08
Use either weight or mass for magnitudes and units


rjyounes:
Gist defines MassUnit and Weight magnitude. Granted that they are interchangeable on the earth's surface, as the notes indicate, it would make more sense if they both referenced mass or both referenced weight, unless there is a reason not to. If there's a reason not to, add it to the notes on the terms so people like me don't get confused.

uscholdm:
Yes, gist:Mass would be a better name for gsit:Weight. Consistency between the unit and the magnitude is handy.

rjyounes:
Deprecate gist:Weight and add an equivalence to gist:Mass, even though not technically correct. Put in gistDeprecated file.

uscholdm:
> Deprecate gist:Weight and add an equivalence to gist:Mass to the gist:Weight definition, even though not technically correct.

Quibble: The equivalence is technically correct. What is wrong is using the term Weight to mean Mass, but that's now deprecated history.

PR #783, opened 2022-12-08
Add owl:Thing assertions to duration units. Fixes #775.

### LINKED ISSUE #775: 3 terms missing `rdfs:isDefinedBy` triples
I'm writing some basic Best Practice-checking code, and in the process I noticed that there are just 3 terms (from the 289 terms in gist) that are missing `rdfs:isDefinedBy` triples, namely:
- `gist:_day`
- `gist:_millisecond`
- `gist:_minute`

I assume this is just a minor oversight (as the other 14 `owl:NamedIndividual` instances do include that triple). And it's easy to fix up by just adding the missing triples.

dylan-sa:
Fixes #775.

### DIFF
--- a/docs/release_notes/pr783.md
+++ b/docs/release_notes/pr783.md
@@ -1,0 +1,3 @@
+### Patch Updates
+
+- Added missing `owl:Thing` assertions to duration units to ensure `rdfs:isDefinedBy` assertions are generated in release package. Issue [#775](https://github.com/semanticarts/gist/issues/775).
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -2615,6 +2615,7 @@ gist:_candela
 gist:_day
 	a
 		owl:NamedIndividual ,
+		owl:Thing ,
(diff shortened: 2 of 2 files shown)


rjyounes (commented):
I'm not convinced this is the right solution. We shouldn't have to add `owl:Thing` assertions to our named individuals; a better approach is to modify onto_tool: see [issue #105](https://github.com/semanticarts/ontology-toolkit/issues/105). 

@uscholdm  What do you think?

rjyounes:
The other option is just to change the yaml file to use `all` rather than `strict`. Is there any reason not to do this?

dylan-sa:
I offer this as a quick fix, but that is a fair point about using `all` rather than `strict`. I agree that using `all` would be the better option in the long term, assuming all of the `NamedIndividual` and `Thing` assertions are removed.

rjyounes:
I don't think we need a quick fix, so I'm closing this.

PR #793, opened 2023-02-21
Define unit of measure for percent. Fixes #785.

### LINKED ISSUE #785: Representing Percents
1) How should percentages be stored when users want to see hundredths vs fractions (75 vs 0.75)
2) Should percentage be considered a unit of measure?
3) Should percentage be considered multiple units of measure depending on underlying units (e.g. number of correct test questions out of total test questions vs. body fat as a percentage of total weight vs. interest rate)

rjyounes:
Fixes #785.

* Define new unit of measure for percent.
* Deprecate class Percentage.


docs/release_notes/pr793.md
```diff
@@ -0,0 +1,5 @@
+### Minor Updates
+
+* Percents, as per issue [#785](https://github.com/semanticarts/gist/issues/785).
+  * Added unit of measure for percents.
+  * Deprecated class `gist:Percentage`.
```
uscholdm, line 5: Add that NamedIndividual's are now gone. That might have been a separate issue and PR, ideally.

uscholdm (changes requested):
Looks good, but for adding a not about NamedIndividual

docs/release_notes/pr793.md
rjyounes: That isn't relevant to this issue. It came in with the upgrade to rdf-toolkit, and I've added the note there.

PR #795, opened 2023-02-22
Changes to pre commit hook. Fixes #794

### LINKED ISSUE #794: pre-commit hook not running
PR #788 may have introduced a bug whereby the pre-commit is no longer running. I thought I had tested the change to rdf-toolkit.jar both manually, by running `serialize.sh`, and as a pre-commit hook, and everything was working at the time I merged to develop.

@stevenchalem Would you mind having a look by trying to commit a file on a branch that has a syntax error? This is how I found just now that it wasn't running.

Jamie-SA:
Fixes #794

- Fix serialize pre-commit hook
  - Make sure the script starts in the root directory of the repository
  - Correct the path to serializer
  - Add file to the commit after the sed command
  - Preserve file permissions during processing steps
  - Return a failure code (non zero return value) when the serializer fails
- Add new "root" pre-commit hook
  - Prevent commits to these branches: develop, main, master
  - Then run the serializer pre-commit hook
- Add a setup.sh file to install the pre-commit hook
- Update the documentation a little

Everyone should run the script `./tools/setup.sh` after pulling in these changes.

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -53,12 +53,19 @@ Submitting an Issue
 
 - External contributors should not specify assignees, projects, or milestones. These are assigned during the internal review process.
 
-Implementation
+Contributing to gist
 -----
+
+After cloning the gist repository, run this script `./tools/setup.sh`.
 
 ### Pre-commit Hook
 
-- The `./tools` directory contains a pre-commit hook that you should copy into `./git/hooks`. This ensures that the serializer is run before each commit, converting files into a standard Turtle format in order to prevent noise in the diffs. As the comments in the file indicate, you should use the version of `rdf-toolkit.jar` in this directory, rather than another version that you may have on your local drive.
+The `./tools` directory contains a pre-commit hook that you should copy into `./git/hooks`. The setup.sh script mentioned above will do this for you.
+
+The pre-commit hook does a several things when you do a `git commit`:
+- Prevent commits to the following branchs: develop, main, and master.
+- Run the serializer before each commit. This converts files into a standard Turtle format in order to prevent noise in the diffs. As the comments in the file indicate, you should use the version of `rdf-toolkit.jar` in this directory, rather than another version that you may have on your local drive.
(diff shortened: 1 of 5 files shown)


rjyounes:
@Jamie-SA We have branch protection rules on develop and main. Why do we need to replicate them in the pre-commit hook?

Jamie-SA:
> @Jamie-SA We have branch protection rules on develop and main. Why do we need to replicate them in the pre-commit hook?

Because it prevents someone from doing what you did yesterday. It stops you from making the commit to the branch in the first place, as opposed to stopping you from pushing the branch to github. I'd personally much prefer to be told up front that I forgot to create a feature branch, instead of having to fix the problem afterwards.

rjyounes:
OK, makes sense. I love the fact that my stupidity was the driver of an improvement. :)

Jamie-SA:
Well, I've done it several times myself which is why I already had done this for the dca repo and figured it would be easy to add to gist.

rjyounes (approved):
Everything looks good here, and I've tested it out.

rjyounes:
@Jamie-SA Hm, it just occurred to me - do `.sh` scripts run on Windows? Do they have to be using a bash shell rather than the Windows PowerShell?

Jamie-SA:
Good point. I suppose we could just update the docs for now and tell people on Windows to copy the pre-commit file, or we could write the equivalent `.bat` file. I went with the script because we may want to add other things to it in the future and it is just easier for people that are not that good with a command line.

Jamie-SA:
I looked into it, there might be a way to write a script that will work on both. I will see if I can get it to work.

rjyounes:
We could also tell windows users they can run the sh file in a bash shell - which people will have if they use VSCode, for example. But if you can write a single script that would be fine. Two is not ideal, obviously, for maintenance reasons. 

You need to open a new PR since I already merged this one.

rjyounes:
@Jamie-SA Added issue #797 to track this change.

PR #798, opened 2023-02-25
Feature 797 add windows support to setup

### LINKED ISSUE #797: Adapt tools/setup.sh for Windows users
The script `<root>/tools/setup.sh` is to be the first time you pull from develop since the merge of PR #795 (on 2/22), due to changes made in the pre-commit setup.

The script runs natively on Mac and Linux. Windows users must either:

* Manually copy `<root>/tools/pre-commit` to `<root>/.git/hooks`, or
* Use a bash shell available through VSCode, Git bash, etc.

@Jamie-SA is working on a solution that will provide a single script that will run natively on all three OSs.

Jamie-SA:
Fixes #797
Modified the setup script so that it might work in the windows command window.
I haven't tested it yet on Windows.


dylan-sa (commented):
I tested the `setup.cmd` script in CMD and PowerShell, and it works perfectly. The unix folks won't be able to run this one, though, right? For example, I don't think I can run `setup.cmd` in ZSH on my Mac. Maybe we want to retain both scripts?

Jamie-SA:
> The unix folks won't be able to run this one, though, right? For example, I don't think I can run `setup.cmd` in ZSH on my Mac.

This worked on linux and I think it will work on Mac. Try to run it in ZSH: `./setup.cmd`

dylan-sa:
Ah, you're absolutely right. User error on my part. 🙂

dylan-sa (approved):
Looking good to me. Tested on Windows and Mac.

rjyounes:
@Jamie-SA Does this require a change to the documentation on setting up the project? Do you need to update the release note (you can include one note for the two issues).

Jamie-SA:
Documentation is now updated.

rjyounes (changes requested):
@Jamie-SA I made a couple of tweaks to the documentation and tried to commit. I am getting the following error when the pre-commit runs: 

line 32: return: can only `return' from a function or sourced script

Line 32 does indeed read "return 0"

I tried the following:

```
if [ ! ${rc} -eq 0 ] ; then
  return 1
else
  return 0
fi
```
But I get a similar error. I don't know why it accepts the "return 1" within the if statement.

Jamie-SA:
I guess some shells are more strict. I've pushed an update that will hopefully fix it.
The old version and the new version are working on my setup so I can't test it.

rjyounes (approved):

rjyounes:
Your fix works, thanks!

PR #811, opened 2023-03-29
Update annotations on gist:isAllocatedBy. Fixes #530.

### LINKED ISSUE #530: Broaden text definition of gist:isAllocatedBy
The definition is: "Connection between an ID and the thing that minted the ID.  It may be a person or organization, or could be an algorithm (next available or random number generator)."

The domain is not specified.

There are other things that can be allocated - e.g., certifications, ratings, default units of measure for an aspect, etc.

wslooney:
…AllocatedBy

### DIFF
--- a/docs/release_notes/pr811.md
+++ b/docs/release_notes/pr811.md
@@ -1,6 +1,4 @@
 ### Minor Updates
 
-* Updated `gist:isAllocatedBy`, as per issue [#530](https://github.com/semanticarts/gist/issues/530).
-    * Updated `skos:definition`.
-    * Added `skos:example`, `skos:scopeNote`, and `gist:domainIncludes` annotations.
+* Broadened definition of `gist:isAllocatedBy` to include entities other than IDs, and added additional annotations, as per [#530](https://github.com/semanticarts/gist/issues/530).
     
--- a/docs/release_notes/pr811.md
+++ b/docs/release_notes/pr811.md
@@ -1,0 +1,6 @@
+### Minor Updates
+
+* Updated `gist:isAllocatedBy`, as per issue [#530](https://github.com/semanticarts/gist/issues/530).
+    * Updated `skos:definition`.
+    * Added `skos:example`, `skos:scopeNote`, and `gist:domainIncludes` annotations.
+    
--- a/docs/release_notes/pr811.md
+++ b/docs/release_notes/pr811.md
@@ -1,4 +1,3 @@
 ### Minor Updates
 
-* Broadened definition of `gist:isAllocatedBy` to include entities other than IDs, and added additional annotations, as per [#530](https://github.com/semanticarts/gist/issues/530).
-    
+* Broadened definition of `gist:isAllocatedBy` to include entities other than IDs, and added clarifying annotations, as per [#530](https://github.com/semanticarts/gist/issues/530).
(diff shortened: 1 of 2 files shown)


gistCore.ttl
```diff
@@ -3450,8 +3450,10 @@ gist:isAllocatedBy
 			gist:Person
 		) ;
 	] ;
-	skos:definition "Connection between an ID and the thing that minted the ID.  It may be a person or organization, or could be an algorithm (next available or random number generator)"^^xsd:string ;
+	skos:definition "Relates the subject to whoever/whatever assigns or distributes it."^^xsd:string ;
+	skos:example "An employee's email address is allocated by the employer. A license plate is allocated by a state's Department of Motor Vehicles. An ID is allocated by whoever minted the ID."^^xsd:string ;
 	skos:prefLabel "is allocated by"^^xsd:string ;
+	gist:domainIncludes gist:ID ;
```
dylan-sa, line 3456: Worth adding `gist:Category` here? Seems similar to`gist:ID`--another gist term that has `isAllocatedBy` in its formal def.

dylan-sa (approved):
One suggestion for the `domainIncludes` statement, but looks good even if you decide against it. Can add 'Fixes #n' to the title to ensure automated closing of the issue.

gistCore.ttl
rjyounes: Dylan's suggestion makes sense to me.

For the ID example, let's think of something more specific, else it's not much of an example. E.g., a student ID, an SSN, etc. An example for Category would also be good. In this case it's allocated by the body governing the taxonomy.

rjyounes (changes requested):

dylan-sa (changes requested):
@wslooney - Just noticed we need a release note for this one too, following our new process of creating a separate release note file for each PR.

gistCore.ttl
```diff
@@ -17,6 +17,26 @@
 	gist:license "https://creativecommons.org/licenses/by-sa/3.0/"^^xsd:string ;
 	.
 
+skos:scopeNote
```
rjyounes, line 20: @DanielHadad  These are Protege artifacts that need to be removed from the file before commit.

docs/release_notes/pr811.md
```diff
@@ -0,0 +1,6 @@
+### Minor Updates
+
+* Updated `gist:isAllocatedBy`, as per issue [#530](https://github.com/semanticarts/gist/issues/530).
```
rjyounes, line 3: I don't think these items are worth calling out separately. You could say, "Broadened definition of `gist:isAllocatedBy` to include entities other than IDs, and updated other annotations."

gistCore.ttl
```diff
@@ -3450,8 +3470,14 @@ gist:isAllocatedBy
 			gist:Person
 		) ;
 	] ;
-	skos:definition "Connection between an ID and the thing that minted the ID.  It may be a person or organization, or could be an algorithm (next available or random number generator)"^^xsd:string ;
+	skos:definition "Relates the subject to whoever/whatever assigns or distributes it."^^xsd:string ;
+	skos:example "A U.S. Social Security number is allocated by the U.S. Social Security Administration. The gist:MediaType 'CVS' is allocated by the gist Development Team."^^xsd:string ;
```
rjyounes, line 3474: The csv example isn't good because, as you can see from the namespaces, these are IANA terms, which we have simply assigned a type `gist:MediaType`. How about  "The media type https://www.iana.org/assignments/media-types/text/csv" is allocated by the Internet Assigned Numbers Authority (IANA)."?

rjyounes (changes requested):

rjyounes (approved):

Issue #590, opened 2021-10-18
Address issues with gist:Task and gist:ScheduledTask


turingtest37:
gist 10.0.0 defines Task as 

> A task which has been defined and either scheduled or accomplished, or both.

It should read "An **Event** which has been..."

uscholdm:
> In the current hierarchy I don't see a difference between Task and ScheduledTask, defined as "A task planned to occur. When it was scheduled, it would have been in the future, but now might be in the past." Isn't that the same as "A task which has been defined and either scheduled or accomplished, or both."?

The difference is that "A task which has been defined and either scheduled or accomplished, or both." might not have been scheduled. A `ScheduledTask` by definition is or was scheduled.

rjyounes:
OK, but is that really a useful distinction? Do we need to know that a task now completed was or wasn't scheduled? Can you describe a context in which that distinction would be important?

rjyounes:
`skos:definition` of `Task` has been changed in PR #597. Remaining issues are:
- Should `Task` be a subclass of `Event`? 
- Is the distinction between `ScheduledTask` and `Task` useful?

dylan-sa:
Following discussion in the 3/24 issue review, I'd like to propose the following changes.

**(1) A new definition for `gist:Task`:**

```
gist:Task
	a owl:Class ;
	owl:equivalentClass [
		a owl:Class ;
		owl:intersectionOf (
			gist:Requirement
			[
				a owl:Restriction ;
				owl:onProperty gist:hasGoal ;
				owl:someValuesFrom gist:Goal ;
			]
		) ;
	] ;
	skos:definition "A piece of work to be done to achieve a goal."^^xsd:string ;
	skos:prefLabel "Task"^^xsd:string ;
	skos:scopeNote "A task itself is distinct from the event of completing that task. Use gist:isBasisFor to relate a gist:Task to a gist:Event. The event of carrying out the task may have already occurred, or it may be scheduled to occur in the future."^^xsd:string ;
	.
```

**Notable changes:**

- `gist:Event` changes to `gist:Requirement`
- Range in property restriction goes from `gist:Intention` to `gist:Goal`
- `skos:definition` changes to align more closely with common usage / dictionary definitions
- `skos:scopeNote` added to provide direction on connecting tasks with events (I believe @rjyounes floated `gist:isBasisFor` as a potential link in 3/24 issue review)

**Rationale:**

[...]

PR #815, opened 2023-04-13
Remove TimeZone & etc. Fixes #650

### LINKED ISSUE #650: Issues relating to `TimeZone` and `TimeZoneStandard`
Thread copied from Teams on 2022-04-04:

_Dan:_

In writing up the instructions for handling the new gist time paradigm, I ran into what I believe to be a problem with the TimeZone and TimeZoneStandard classes.  
The current draft formal definition for TimeZone is "gist:GeoRegion and (gist:hasOffsetToUniversal some gist:Duration)". The definition for TimeZoneStandard is "gist:Specification and (gist:isBasedOn some gist:TimeZone)".
I believe we have these a bit backwards because I cannot figure out how to create meaningful instances of these.  
Instead, I think that TimeZoneStandard should be "gist:Specification and (gist:timeOffset some xsd:duration)", and TimeZone should be "gist:GeoRegion and (gist:isAffectedBy some gist:TimeZoneStandard)".

 This requires bringing back the gist:timeOffset property I had in my original proposal.
    gist:timeOffset rdf:type owl:DatatypeProperty ;
        rdfs:range xsd:duration .

Example usage
```
    :_TimeZoneStandard_easternStandard gist:TimeZoneStandard ;
        gist:timeOffset "-PT5H"^^xsd:duration .
    :_TimeZoneStandard_easternDaylight gist:TimeZoneStandard ;
        gist:timeOffset "-PT4H"^^xsd:duration .
    :_TimeZone_usEastern a gist:GeoRegion ; 
        gist:usesTimeZoneStandard :_TimeZoneStandard_easternDaylight .
    :_Event_x gist:plannedDateTime "2022-06-08T09:00:00"^^xsd:dateTime ;
        gist:usesTimeZoneStandard :_TimeZoneStandard_easternDaylight .
```

_Rebecca:_

Given the definition of TimeZoneStandard  as "The algorithm for getting from Greenwich Mean Time to local time, which includes the time zone offset and rules about daylight savings time," it seems to me that there are not two standards, eastern standard and eastern daylight. Rather, this is one standard that has rules for when it switches from daylight to standard and back.
 
The definition does call for a time zone offset, however, but given that different time zones switch from daylight to standard on different dates, is this possible? If it is, then I'm in favor of it.
 
I prefer the semantics of gist:isBasedOn to gist:isAffectedBy, which is not specific enough, but I agree that we have it backwards - the time zone is based on the time zone started rather than the other way around.
 
I have a somewhat different idea of how this should go - I think an event takes place in a time zone, and is only indirectly related to the time zone standard through the time zone. This gives us triples such as:

```
:_TimeZoneStandard_eastern a gist:TimeZoneStandard .
:_GeoRegion_usEasternTimeZone a gist:TimeZone ;
    gist:isBasedOn :_TimeZoneStandard_eastern .
:_Event_x gist:plannedEndMinute "2022-06-08T09:00:00"^^xsd:datetime ;
    gist:isInTimeZone :_GeoRegion_usEasternTimeZone .
```

Jamie-SA:
Fixes #650

Removes
- gist:TimeZone
- gist:TimeZoneStandard
- gist:hasOffsetToUniversal
- gist:usesTimeZoneStandard

### DIFF
--- a/docs/release_notes/issue650.md
+++ b/docs/release_notes/issue650.md
@@ -1,0 +1,10 @@
+### Major Updates
+
+Addresses issue #650.
+
+Removes
+- gist:TimeZone
+- gist:TimeZoneStandard
+- gist:hasOffsetToUniversal
+- gist:usesTimeZoneStandard
+
--- a/docs/release_notes/issue650.md
+++ b/docs/release_notes/issue650.md
@@ -1,10 +1,10 @@
 ### Major Updates
 
-Addresses issue #650.
+Addresses issue [#650](https://github.com/semanticarts/gist/issues/650).
 
 Removes
-- gist:TimeZone
-- gist:TimeZoneStandard
-- gist:hasOffsetToUniversal
-- gist:usesTimeZoneStandard
+- `gist:TimeZone`
+- `gist:TimeZoneStandard`
+- `gist:hasOffsetToUniversal`
+- `gist:usesTimeZoneStandard`
 
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -2820,7 +2820,7 @@ The subproperties allow the ontologist to do three things:
 
 All datetimes are of the same format: '2021-06-01T08:03:27.12324-6:00'^^xsd:dateTime. This is compatible with and a subset of ISO 8601.
 
-Time zone offset, such as -6:00 (of which there are a few dozen) is recognized in the date itself, as shown. The actual time zone standard (of which there are 131) may optionally be attached to the event or other object itself.
+A time zone offset, such as -6:00, can be included as part of the datetime itself, as shown. The offset is not the same as a time zone. There are roughly a few dozen time zone offsets and around 130 named time zones. If needed, a time zone may be recorded by attaching it to the event or other object.
 
 There will be many historical dates that do not have a time zone offset (e.g., Lincoln's birthday, as well as about 75% of all legacy systems), and in that case the offset can be omitted. 
 
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -2439,40 +2439,6 @@ gist:Text
 	skos:prefLabel "Text"^^xsd:string ;
 	.
 
-gist:TimeZone
-	a owl:Class ;
-	owl:equivalentClass [
-		a owl:Class ;
-		owl:intersectionOf (
-			gist:GeoRegion
-			[
-				a owl:Restriction ;
-				owl:onProperty gist:hasOffsetToUniversal ;
-				owl:someValuesFrom gist:Duration ;
-			]
-		) ;
-	] ;
-	skos:definition "A region that observes a uniform standard time for legal, commercial, and social purposes.  A typical time zone averages 15? of longitude in width and typically observes a clock time one hour earlier than the zone immediately to the east."^^xsd:string ;
-	skos:prefLabel "Time Zone"^^xsd:string ;
-	.
-
-gist:TimeZoneStandard
-	a owl:Class ;
-	owl:equivalentClass [
-		a owl:Class ;
-		owl:intersectionOf (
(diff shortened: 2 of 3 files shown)


docs/release_notes/issue650.md
```diff
@@ -0,0 +1,10 @@
+### Major Updates
+
+Addresses issue #650.
```
dylan-sa, line 3: Could add the link here. https://github.com/semanticarts/gist/issues/650

dylan-sa (changes requested):
Just a couple minor things on the release notes; otherwise looking good to me.

docs/release_notes/issue650.md
rjyounes: I would like to see some descriptive text here, as we typically do with our release notes. For example:  

- Renamed `gist:Task`, `gist:ScheduledTask` and `gist:Project` to `gist:TaskExecution`, `gist:ScheduledTaskExecution` and  `gist:ProjectExecution`, respectively. Issue [#590](https://github.com/semanticarts/gist/issues/590).

This includes the link and backticks that Dylan has suggested.

gistCore.ttl
```diff
@@ -2854,7 +2820,7 @@ The subproperties allow the ontologist to do three things:
 
 All datetimes are of the same format: '2021-06-01T08:03:27.12324-6:00'^^xsd:dateTime. This is compatible with and a subset of ISO 8601.
 
-Time zone offset, such as -6:00 (of which there are a few dozen) is recognized in the date itself, as shown. The actual time zone standard (of which there are 131) may optionally be attached to the event or other object itself; see property gist:usesTimeZoneStandard.
+Time zone offset, such as -6:00 (of which there are a few dozen) is recognized in the date itself, as shown. The actual time zone standard (of which there are 131) may optionally be attached to the event or other object itself.
```
rjyounes, line 2823: I would like to get rid of the second sentence about time zone standard, since we couldn't agree yesterday about what a time zone standard is. Since we deleted the property the comment isn't very relevant anyway - I.e., we are not providing a way to attach the standard to the event.

rjyounes (changes requested):

docs/release_notes/issue650.md
Jamie-SA: @rjyounes I'm not sure what more information you would like to see. Your example shows saying those things were renamed. Below I've already said what things were removed. I think what I have is similar to your example.

gistCore.ttl
```diff
@@ -2854,7 +2820,7 @@ The subproperties allow the ontologist to do three things:
 
 All datetimes are of the same format: '2021-06-01T08:03:27.12324-6:00'^^xsd:dateTime. This is compatible with and a subset of ISO 8601.
 
-Time zone offset, such as -6:00 (of which there are a few dozen) is recognized in the date itself, as shown. The actual time zone standard (of which there are 131) may optionally be attached to the event or other object itself; see property gist:usesTimeZoneStandard.
+A time zone offset, such as -6:00, can be included as part of the datetime itself, as shown. The offset is not the same as a time zone. There are roughly a few dozen time zone offsets and around 130 named time zones. If needed, a time zone may be recorded by attaching it to the event or other object.
```
rjyounes, line 2823: > If needed, a time zone may be recorded by attaching it to the event or other object.

Since we are removing the classes, this should also be removed.

rjyounes (changes requested):
One requested change, otherwise looks good.

rjyounes (approved):

rjyounes:
@dylan-sa Can you review Jamie's changes in response to your review comments, and approve it or submit a new request for changes? This is close to being ready to merge to develop.

dylan-sa (approved):
Looks good.

PR #822, opened 2023-04-20
Fix gist logo in docs

### LINKED ISSUE #750: Fix issue with gist logo
Get the path in the README right so that it works both on GitHub and the README in the release package. In previous versions it was `gist logo` but now that doesn't work, so I've used the GitHub upload path instead. That means that `README.html` in the release package points to the image on GitHub, which isn't ideal. Need to look at previous versions to see how it was done.

Jamie-SA:
Fixes #750

This should fix the gist logo in `README.md` and `docs/README.*`


Jamie-SA:
I don't know why the tests are failing again. I've tried several things to fix it and none have worked.

I did run the bundle to test my changes before submitting this PR and it seemed to work correctly. However, now I've broken my onto_tool install on my machine while trying to track down the problem and can't get it to run.

.github/workflows/validate.yaml
```diff
@@ -15,8 +15,9 @@ jobs:
           python-version: 3.8
       - name: Install onto-tool
         run: |
-          # python -m pip install --upgrade pip
-          pip install onto-tool
+          python -m pip install guess-language-spirit
```
rjyounes, line 18: Do you still want to keep these changes?

rjyounes (commented):
Everything looks good here, but wondering if you want to preserve the changes you made to validate.yaml or remove them.

rjyounes (approved):

PR #824, opened 2023-04-24
Modify setup script to make pre-commit hooks executable (needed on Mac). Fixes #823

### LINKED ISSUE #823: setup.cmd should make pre-commit hooks executable
At least on Mac, when setup.cmd runs and copies the pre-commit hook into ./git/hooks, it needs to be made executable - _after_ the copy. If there isn't already an executable file in the hooks directory, it gets copied as non-executable, regardless of the status of the original file in tools. 

We should also make the pre-commit in the serializer executable, in case it hasn't come in that way from the repo.

rjyounes:
Fixes #823.

### DIFF
--- a/tools/pre-commit
+++ b/tools/pre-commit
@@ -20,6 +20,9 @@ base_dir=$(git rev-parse --show-toplevel)
 
 # Run the serializer
 set -x
+
+# Ensure that the serializer pre-commit is executable
+chmod +x "${base_dir}/tools/serializer/pre-commit"
 "$base_dir/tools/serializer/pre-commit"
 rc=$?
 set +x
@@ -30,4 +33,3 @@ if [ ! ${rc} -eq 0 ] ; then
 fi
 
 exit 0
-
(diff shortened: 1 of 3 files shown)


.github/workflows/validate.yaml
```diff
@@ -16,6 +16,7 @@ jobs:
       - name: Install onto-tool
         run: |
           # python -m pip install --upgrade pip
+          pip install --upgrade pip
```
Jamie-SA, line 19: This has a merge conflict. Probably best to remove the change in the validate.yaml file, and leave the changes that I made in the logo PR.
```suggestion
```

Jamie-SA (changes requested):

tools/pre-commit
```diff
@@ -30,4 +33,3 @@ if [ ! ${rc} -eq 0 ] ; then
 fi
 
 exit 0
-
```
Jamie-SA, line 33: General good practice is that files should end with a newline.

Jamie-SA (approved):

Jamie-SA:
@rjyounes is there a reason this PR is marked as a draft? I think it is ready to merge.

rjyounes:
I had some trouble with it and put it in draft state. I'll have another look at it.

PR #837, opened 2023-05-08
Add supplementary subclass assertions. Fixes #714.

### LINKED ISSUE #714: Explicitly assert subClassOf in gist where applicable
**Request**: Explicitly assert subClassOf in gist where applicable.

**Example**:  `gist:Commitment` is defined as
```
  (Requirement or Restriction)
 and (hasGiver some 
    (Organization or Person))
 and (isCategorizedBy some DegreeOfCommitment)
```
but it is not explicitly defined as subClassOf `gist:Intention`, even though it must be. 
Another example is `gist:CoherentUnit` (not explicitly declared as subClassOf `gist:UnitOfMeasure`)

**Rationale**:  We want gist to be as usable as possible.  Many more triple stores and visualization tools support class subsumption but do not support OWL to the level where they can "figure out" the subsumption if implied.

dylan-sa:
Fixes #714.

This PR contains:
- A new ontology file, `gistSubClassOfAssertions.ttl`, which contains supplementary `rdfs:subClassOf` assertions that cannot be inferred by some reasoners. (For completeness, it also contains the already-explicit `rdfs:subClassOf` assertions from gistCore.)
- Tweaks to the `bundle.yaml` file to ensure that the new ontology file will be bundled into the release. Instead of adding the file name to the growing list of files, I have used a wildcard `*.ttl`, which seems to be generating the package correctly, so that we get three file types for the subClassOf supplement.

Note: I have included the subClassOf assertions ontology here mainly for review. If it's not something we want to store in the gist git repository, I can remove it before merging and then we can add it back during the v12 release process.


gistSubClassOfAssertions.ttl
```diff
@@ -0,0 +1,514 @@
+# imports: https://ontologies.semanticarts.com/o/gistCore
+
+@prefix gist: <https://ontologies.semanticarts.com/gist/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+<https://ontologies.semanticarts.com/o/gistSubClassOfAssertions.ttl>
```
rjyounes, line 10: Can we simplify this to "gistSubClassAssertions" - I.e., delete "of"? Same with the filename.

gistSubClassOfAssertions.ttl
```diff
@@ -0,0 +1,514 @@
...
+	a owl:Ontology ;
+	owl:imports <https://ontologies.semanticarts.com/o/gistCore> ;
+	skos:definition "Supplementary subClassOf assertions for gistCore."^^xsd:string ;
```
rjyounes, line 13: I would either say "subclass assertions," using non-technical language, or "rdfs:subClassOf assertions," using technical language, but not a mix of both.

gistSubClassOfAssertions.ttl
```diff
@@ -0,0 +1,514 @@
...
+	skos:prefLabel "gist subClassOf Assertions"^^xsd:string ;
```
rjyounes, line 14: Ditto

gistSubClassOfAssertions.ttl
```diff
@@ -0,0 +1,514 @@
...
+	skos:scopeNote "This ontology file contains supplementary subClassOf assertions that are implied by gistCore but some automated reasoners will be unable to infer. More precisely, it contains (1) subClassOf assertions derived using an OWL DL reasoner and (2) the subClassOf assertions that are already explicit in gistCore."^^xsd:string ;
```
rjyounes, line 15: Ditto about subClassOf. I would recommend a more technical term than "implied," though. Perhaps "entailed by" or "can be inferred from."

rjyounes (changes requested):
In addition to the embedded comments:

- You need a release note.
- I would like the file to remain in the repo so that we can use it during development as well as in the release package. In fact once we have included the script in the repository, we can run it in our pre-commit hook so that the file is always up-to-date.

bundle.yaml
```diff
@@ -94,9 +91,7 @@ actions:
     from: "(.*)\\.ttl"
     to: "\\g<1>{version}.ttl"
   includes:
-    - gistCore.ttl
-    - gistMediaTypes.ttl
-    - gistPrefixDeclarations.ttl
+    - "*.ttl"
 - action: "definedBy"
   message: "Adding rdfs:definedBy."
```
rjyounes, line 96: Though not part of this PR - could you correct this to `rdfs:isDefinedBy`?

uscholdm:
> * which contains supplementary `rdfs:subClassOf` assertions that cannot be inferred by some reasoners

How about give an example of one that will not infer them and why. E.g. RL reasoners ignore cannont handlye equivalentClass axioms.

bundle.yaml
dylan-sa: Fixed

dylan-sa:
Thank you for comments, @rjyounes & @uscholdm. I've added a release note, tweaked the yaml file, and adjusted the annotations in the new file. Ready for another look when you get a chance.

docs/release_notes/pr837.md
```diff
@@ -0,0 +1,7 @@
+### Minor Updates
+
+- Added `gistSubClassAssertions.ttl` file with supplementary sublcass assertions that are logically entailed but not inferred by some reasoners. Issue [#714](https://github.com/semanticarts/gist/issues/714).
```
rjyounes, line 3: "...that are logically entailed but not inferred by some reasoners..."

There's a small amount of ambiguity here which would be solved by adding a comma after "entailed."

gistSubClassAssertions.ttl
```diff
@@ -0,0 +1,514 @@
...
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+<https://ontologies.semanticarts.com/o/gistSubClassAssertions.ttl>
+	a owl:Ontology ;
+	owl:imports <https://ontologies.semanticarts.com/o/gistCore> ;
+	skos:definition "Supplementary subclass assertions for gistCore."^^xsd:string ;
+	skos:prefLabel "gist Subclass Assertions"^^xsd:string ;
+	skos:scopeNote "This ontology file contains supplementary subclass assertions that are logically entailed by gistCore but are unable to be inferred by some automated reasoners. (For example, an RL reasoner would not infer that gist:Commitment is a subclass of gist:Intention, even though it must be.) More precisely, it contains (1) subclass assertions derived using an OWL DL reasoner and (2) the subclass assertions that are already explicit in gistCore."^^xsd:string ;
```
rjyounes, line 15: "...are unable to be inferred by some automated reasoners."

This is awkward. How about just "...are not inferred..."?

I would remove the parens - this is an important point that doesn't need to be parenthesized.

I would say "gist" instead of "gistCore" so it doesn't have to be modified if we add other ontology files.

rjyounes (commented):
I've suggested a couple of minor wording tweaks, otherwise everything looks good.

tools/gistSubClassAssertions.py
```diff
@@ -0,0 +1,49 @@
+import os
```
rjyounes, line 1: Can you change the filename to generateSubClassAssertions.py? I prefer a verbal name for a script since it does something.

tools/gistSubClassAssertions.py
```diff
@@ -0,0 +1,49 @@
...
+import tempfile
+from rdflib import Graph
+import rdflib
+from owlready2 import get_ontology, default_world, sync_reasoner
+from rdflib.namespace import RDFS
+
+# This script can be used during the gist release process to generate gistSubClassAssertions.ttl. After running it, the release manager should manually add the ontology metadata (the owl:imports statement, the skos:prefLabel, etc.) before generating the release package.
```
rjyounes, line 8: Let's leave this comment out. As I said, I would like to add it to the pre-commit hook. We can add the current process to the Confluence documentation.

docs/release_notes/pr837.md
```diff
@@ -0,0 +1,9 @@
+### Minor Updates
+
+- Added `gistSubClassAssertions.ttl` file with supplementary sublcass assertions that are logically entailed, but not inferred by some reasoners. Issue [#714](https://github.com/semanticarts/gist/issues/714).
```
rjyounes, line 3: typo - subclass

docs/release_notes/pr837.md
```diff
@@ -0,0 +1,9 @@
...
+
+### Patch Updates
+
+- Amended `bundle.yaml` file to include `gistSubClassAssertions.ttl` in the release package. Issue [#714](https://github.com/semanticarts/gist/issues/714).
```
rjyounes, line 7: This isn't quite true - you generalized bundle.yaml to include any ttl files in the directory. You could either leave this out or phrase it that way.

docs/release_notes/pr837.md
```diff
@@ -0,0 +1,9 @@
...
+
+- Added `materialize_subclass_inferences.py`, which generates the subclass assertions for `gistSubClassAssertions.ttl`. Issue [#714](https://github.com/semanticarts/gist/issues/714).
```
rjyounes, line 9: "Added Python script ..."

rjyounes (changes requested):
Just a couple more small wording changes...

rjyounes (approved):

Issue #733, opened 2022-09-25
Remove properties hasSubTask and hasDirectSubTask


rjyounes:
This is covered by `hasPart` and `hasDirectPart` and object `Task`.

rjyounes:
Dave defined these predicates as an initial effort to distinguish various part-of relationships. We agree as a group to remove these predicates, but will run it by Dave first. The part-of issue is larger, and needs to be tackled as a separate topic.

hmoore-sa:
I was a little delayed in getting started on this due to client work, but I have an email out to Dave and will move ahead based on his feedback. WIP.

mkumba:
Yeah I guess its ok to remove hasSubTask and replace it with isDirectPartOf

As I think of it, what we really need is a way to distinguish the part of relationship that must be exclusive and therefore is a candidate for cascading delete.  My current thinking is it is the difference between a physical part and a digital part.  A physical part (with a serial number say) can only be directly part of one other physical part (a Vin number say), and if the vin number is destroyed so is the part.  But the design of that part (caburettor) could be part of many higher level designs (car, truck etc) and is not deleted when its parent is deleted. 

On the task side a subtask is exclusive like that (I think ) but a template task isn't, we could share a template task across many template projects.  

So now we need a name for this distinction.

rjyounes:
Our thinking here is that we need to think through the complexities of mereology and implement it in a consistent way. Until we do that, it doesn't make much sense to differentiate one.

mkumba:
Ugghhh that feels like an infinite regress.  I just went back and sped re-read Peter Simons Parts and in the 300+ pages there are one or two pretty good ideas.  (Its pretty old (1987) and predates DL and OWL and everything but I get the sense the world has not moved on, that the quote on the cover came to pass “Parts could easily be the standard book on mereology for the next twenty or thirty years”)

Ok

Let’s sort it into things we can deal with.

1) Real physical countable things (PhysicallyIdentifialItems).  With the exception of his hi-way example (two roads share an intersection) pretty much nothing else physical shares physical parts.  Systems share parts.  But systems aren’t PIIs. And the hi ways are probably better thought of as networks, not PIIs.
2) Real Physical Substance (mass nouns) Two beers share alcohol, but it isn’t the same alcohol, it’s the same type of alcohol.
3) Physical Items can contain substance (my car might contain some fuel)
4) Temporal things (events) can have temporal parts.  The play last night had three acts.
[...]

PR #839, opened 2023-05-10
issue 733 - gist:hasSubTask and gist:hasDirectSubTask - all changes

hmoore-sa:
Hi Rebecca,

This PR covers all changes for 733:
1) removed gist:hasSubTask and gist:hasDirectSubTask (since this is a major change coming up, removing them entirely per our deletion policy).
2) updated restriction on gist:ProjectExecution to reference gist:hasPart instead of gist:hasSubTask.
3) added release note file with change list.

Please let me know if I've missed anything. Thanks!
-Heather

### DIFF
--- a/docs/release_notes/issue733.md
+++ b/docs/release_notes/issue733.md
@@ -1,4 +1,3 @@
 ## Major Updates
 
-- Removed gist:hasSubTask and gist:hasDirectSubTask. In future, gist:hasPart and gist:hasDirectPart will respectively be used.
-- Restriction on gist:ProjectExecution updated to use gist:hasPart instead of gist:hasSubTask.
+- Removed `gist:hasSubTask` and `gist:hasDirectSubTask`. In future, `gist:hasPart` and `gist:hasDirectPart` will respectively be used. Updated the restriction on `gist:Project` to use `gist:hasPart` instead of `gist:hasSubTask`. Issue [#733](https://github.com/semanticarts/gist/issues/733).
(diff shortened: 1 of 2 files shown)


docs/release_notes/issue733.md
```diff
@@ -0,0 +1,4 @@
+## Major Updates
+
+- Removed gist:hasSubTask and gist:hasDirectSubTask. In future, gist:hasPart and gist:hasDirectPart will respectively be used.
```
rjyounes, line 3: We use backticks around term names.

docs/release_notes/issue733.md
```diff
@@ -0,0 +1,4 @@
...
+- Restriction on gist:ProjectExecution updated to use gist:hasPart instead of gist:hasSubTask.
```
rjyounes, line 4: Let's group these as a single item:

- Removed ... The restriction on...

Then we have a link to the issue:
```
Issue [#733](https://github.com/semanticarts/gist/issues/733).
```

So all in all, something like:

Removed `gist:hasSubTask` and `gist:hasDirectSubTask`. In future, `gist:hasPart` and `gist:hasDirectPart` will respectively be used. Updated the restriction on `gist:Project` to use `gist:hasPart` rather than `gist:hasSubTask`. Issue [#733](https://github.com/semanticarts/gist/issues/733).


I've changed `gist:ProjectExecution` to `gist:Project` because in another PR we have replaced the former with the latter.

rjyounes (changes requested):
Just requesting a couple of tweaks to the release note. The OWL is fine.

docs/release_notes/issue733.md
hmoore-sa: Thank you! Updated.

rjyounes (approved):

Issue #774, opened 2022-11-17
Integrate new rdf-toolkit option to remove NamedIndividual assertions into gist workflow


rjyounes:
`rdf-toolkit` provides a new option to suppress named individual assertions. These are added by Protege and we would like to eliminate them from gist. The option is:

```
-sni,--suppress-named-individuals       suppresses all instances of owl:NamedIndividual
```

Integrating this option into our workflow involves:
* Download the new version of `rdf-toolkit` into the `tools` directory
* Update `serialize` action in `bundle.yaml`
* Update `serialize` shell script

I'm also going to move all the serializer files into a subdirectory `tools/serializer`.

rjyounes:
In progress, waiting for EDM Council to release the change and distribute a new jar file.

rjyounes:
EDM Council has released a new version 1.14.1, but I still cannot get it to work and have sent test files to them. [See rdf-toolkit issue #28](https://github.com/edmcouncil/rdf-toolkit/issues/28).

Jamie-SA:
@stevenchalem as the release manager has offered to help look into this with some help from a more experienced person.

rjyounes:
I've already made some progress with this so I'll continue with it. The issue so far is that EDM Council released a version that is supposed to remove the named individuals but so far hasn't worked. I can consult with Steve to see if he can give it a try, as maybe I've missed something.

rjyounes:
EDM Council has released a new version that they believe fixes the problem. I will do some testing and hopefully be able to close this issue out. @stevenchalem I think I'm all set with this one. I'll let you know if I could use some help.

PR #843, opened 2023-05-16
Add sni serializer option to bundler to suppress named individuals. 

rjyounes:
This option is used by the serializer during development, and the bundler should mirror those options (granted that it's technically unnecessary since it will have been done on commit). 

No release note needed, goes under umbrella of #774.


rjyounes:
For some reason the bundler fails when I add the `-sni` option. This makes no sense to me. I don't know if it has something to do with the fact that the default arguments in `onto_tool` don't include it. But then what's the  point of defining values in the yaml file?

```
def serialize_to_output_dir(tools, output, version, file):
    """Serialize ontology file using standard options."""
    base, ext = splitext(basename(file))
    output_file = join(output, f"{base}{version}{ext}")
    logging.debug(f"Serializing {file} to {output}")
    serialize_args = [
        "java",
        "-jar", join(tools, "rdf-toolkit.jar"),
        "-tfmt", "rdf-xml",
        "-sdt", "explicit",
        "-dtd",
        "-ibn",
        "-s", file,
        "-t", output_file]
    subprocess.run(serialize_args)
    return output_file
```

rjyounes:
Closing - need to create new PR from a different branch.

docs/release_notes/pr943.md
```diff
@@ -0,0 +1,3 @@
+### Patch Updates
+
+- Added option to release bundler configuration to suppress named individual type assertions.
```
dylan-sa, line 3: Should we cite the original issue for this one? https://github.com/semanticarts/gist/issues/774

Issue #683, opened 2022-06-09
Shouldn't gist use a PURL for its namespace IRI?


pmcb55:
Currently the namespace IRI for gist Core is: `https://ontologies.semanticarts.com/gist/`, but this IRI is problematic for two big reasons:
 - It tightly couples the ontology to the commercial company Semantic Arts, which might make non-USA governments, or non-USA-based corporations, wary of reusing it.
 - If Semantic Arts goes away (e.g., is bought out by another company, or it goes bust), then the domain name `https://semanticarts.com` may no longer exist or be resolvable, thereby preventing lookups. This could make any organization, USA-based or not, wary of having their data contain IRIs that may become unresolvable in future.

Instead it would be advisable (as is very common practice for ontologies generally) to define the namespace IRI using a PURL service, such as `https://w3id.org`. (Other PURL services could be considered perhaps (but the only other 'major' PURL services that I'm aware of at the moment are European ones like the new `https://purl.eu`, or `https://data.europa.eu/`)).

If `https://w3id.org` were chosen, this would just require the registration of 'gist' and a simple `.htaccess` file to redirect from a new gist namespace IRI (e.g., something like `https://w3id.org/gist/core/`) to the current location of the gist ontology, i.e., `https://ontologies.semanticarts.com/o/gistCore`.

uscholdm:
This is a good idea. However, we are not exactly sure how this will work. Say we adopt this approach. We create a redirect rule so that when someone types the W3ID URI, it goes to Semantic Arts servers, where we control what is returned. Say Semantic Arts goes bust, so the redirect rule points to a non-functioning domain. How does using permanent URIs help? Who/what will serve up anything useful? How is it any different than what we do now, if we go bust, the domain stops working in both cases.

rjyounes:
@pmcb55 A couple of additional points:

* We would like the w3id namespace to include Semantic Arts to identify us as the ontology authors and maintainers, even if the company were to cease to exist. So we would propose `https://w3id.org/semanticarts/gist`. 
* Unlike common practice, we use two different IRIs for the namespace and the ontology IRI. The former is `https://ontologies.semanticarts.com/gist/`, while the latter is `https://ontologies.semanticarts.com/o/gistCore`. How will that square with the use of the w3id URI?

rivettp:
[...]

PR #852, opened 2023-05-18
Add docs/Namespace.md (from Dave)

Jamie-SA:
Part of #683

This came from Dave. I only slightly edited it, mainly to turn it into markdown.


docs/Namespace.md
```diff
@@ -0,0 +1,182 @@
...
+ 5  @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+ 6  @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+ 7  @prefix xml: <http://www.w3.org/XML/1998/namespace> .
+ 8  @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+ 9  
+10  <https://data.semanticarts.com/example> rdf:type owl:Ontology .
+11  sa:mccomb gist:owns dbr:semanticarts .
+```
+
+The first 8 lines declare namespace prefixes. Strictly speaking they are
```
uscholdm, line 102: I wonder if a more realistic example is possible. This is a dataset rather than an ontology in the usual sense, ABox rather than TBox. Much or even most of the time such ABox triples are not wrapped in an `owl:Ontology`.

uscholdm (commented):
This is excellent, I have one comment.  I wonder if a more realistic example  showing the separation between ontology and namespace is possible. The examplel 'ontology' is a dataset rather than an ontology in the usual sense, (i.e. ABox rather than TBox). Much or even most of the time such ABox triples are not wrapped in an `owl:Ontology`. I cnnot think of a good example at the moment.

Jamie-SA:
I personally don't think it needs to be modified. It is just another file in the docs folder. If someone reads it in the bundled package they will get the gist of it.

docs/Namespace.md
rjyounes: Strictly speaking this is not a DL-compliant ontology. I haven't run this one through HermiT, but HermiT requires that any term used in an "ontology" is defined.

rjyounes:
I've made the change to `bundle.yaml`. Now we just need to agree on the example. See comments from @uscholdm and @rjyounes above.

PR #854, opened 2023-05-18
Updated Gist namespace in all turtle files. Fixes #853.

### LINKED ISSUE #853: Replace namespaces in all files in gist repo
A sub-issue of #683

Change the gist namespace from `https://ontologies.semanticarts.com/gist/` to `https://w3id.org/semanticarts/gist/`.

coltonglasgow:
Fixes #853.

- Updated the Gist namespace in all turtle files. 
- Did not update the Gist prefix in the `.rq` files (SPARQL queries) in the `migration/` directory. 
  > If you are migrating from Gist 9 -> 10 or 10 -> 11 then your data will have the old Gist prefix.


rjyounes (changes requested):
The migration scripts should not be changed, since v11 used the old namespace, so when migrating up to 11 you need to keep the same namespace. The migration from 11 to 12 will need to manage the namespace change. Users are told that migrations need to be applied in sequence, so that if you are on version 10, you need to first apply the 11 migration and then the 12 on top of it.

The branch is out-of-date with develop. When you merge, please be sure to get the `-sni` rdf-toolkit.jar option in the list of arguments.

rjyounes:
@coltonglasgow Please remove changes to the migration scripts, as requested above.

coltonglasgow:
> @coltonglasgow Please remove changes to the migration scripts, as requested above.

My bad, I purposefully did not update the migration queries (noted in my first comment) but I missed the migration input data which is in turtle. I reverted the gist namespace change for those migration input files in my latest commit.

rjyounes (approved):

rjyounes:
@coltonglasgow Ready to merge once you've merged/rebased develop.

PR #847, opened 2023-05-17
Add MonetaryPerDuration and CurrencyPerDurationUnit classes. Fixes #817

rjyounes:
Fixes #817.

PR #856, opened 2023-05-19
Add MonetaryPerDuration and CurrencyPerDurationUnit classes. Fixes #846

### LINKED ISSUE #846: Add magnitude and unit classes for monetary-per-duration
gistAcct, gistHR, and gistProfSrv all require the classes `MonetaryPerDuration` (subclass of `Magnitude`) and `CurrencyPerDurationUnit` (subclass of `UnitOfMeasure`). Since gistAcct and gistHR are independent - neither imports the other - these classes need to be promoted to gistCore. They are needed now rather than waiting for the post-12.0.0 release.

Specific units, such as USD_per_hour or US_per_year can be defined in client ontologies rather than gistCore.

rjyounes:
Fixes #846. 

Due to some technical issues regarding branch names, I had to recreate the former PR #847. You can refer to comments there.

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -178,6 +178,7 @@ The pre-commit hook does several things when you run `git commit`:
   - Ontology and other code terms are enclosed in backticks and use the `gist:` prefix; e.g., `gist:Weight`.
   - Each note includes a link to the issue(s) - but not PRs - addressed.
   - Each note ends in a period.
+  - Bullets are symbolized with dash rather than asterisk.
   - Refer to the most recent release notes for examples.
   - Follow the heading conventions shown, as the [markdown configuration file](.markdownlint.json) dictates the mixed heading style.
   - Release note example:
--- a/docs/release_notes/issue817.md
+++ b/docs/release_notes/issue817.md
@@ -1,0 +1,3 @@
+### Minor Updates
+
+- Added classes `gist:MonetaryPerDuration` and `gist:CurrencyPerDurationUnit`. Issue [#817](https://github.com/semanticarts/gist/issues/817).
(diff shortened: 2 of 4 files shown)


dylan-sa:
Continuing the discussion from #847:
> Are these definitions correct in that they follow current practice in defining magnitude and unit classes? This is what needs to be reviewed. 

Yes, makes sense to keep that as the focus for now. I think the relevant analogs would be subclasses of `ProductUnit` and `RatioUnit`, because `CurrencyPerDurationUnit` falls under the latter and product units are structurally similar w/ multipliers and multiplicands rather than numerators and denominators. Here's what we have currently:

- `VolumeUnit` and `AreaUnit` use qualified cardinality restrictions. (Not quite the pattern I suggested in 847, but close.)

- `CoherentProductUnit` and `CoherentRatioUnit` use `allValuesFrom`. (Part of what I suggested with 847 but w/o the cardinality restrictions.)

To more closely align with current practice, we might go w/ one of these patterns.

rjyounes:
@dylan-sa OK, the samples I looked at must have been only the ones with someValuesFrom. Here's what I propose:

* For this issue, change someValuesFrom to allValuesFrom
* hasNumerator and hasDenominator are functional properties. That will achieve the cardinality constraint you propose, but it's more generally applicable than here so I wouldn't want to add it here. Same for hasMultiplier and hasMultiplicand. Add an issue for this. 
* The qualified cardinality and someValuesFrom cases seem to me just wrong. Add an issue to use allValuesFrom instead. (However, this is going to be subsumed by the units and magnitudes group, so we may not want to actually implement this.

Will this satisfy all of your concerns?

dylan-sa:
Sounds good to me. 👍

rjyounes:
@dylan-sa Change made. Please give a quick review.

rjyounes:
Issues added: #857 (functional properties) and #858 (allValuesFrom).

PR #868, opened 2023-06-06
Update Namespace.md to replace quote characters

Jamie-SA:
Change quote characters


stevenchalem (approved):
Onto_tool bundler ran successfully. Also visually confirmed changes in file.

PR #871, opened 2023-06-09
Release/12.0.0

stevenchalem:
Merging release note edits back into Develop


Jamie-SA (changes requested):
@stevenchalem it looks like your changes replaced the release notes for v11.1.0. I think we want to keep the v11.1.0 notes and add the v12.0.0 notes above them.

stevenchalem (commented):
Please check again.

Jamie-SA (approved):

PR #825, opened 2023-04-25
Remove `TreatyOrganization` and modify `IntergovernmentalOrganization`. Fixes #766, #756.

hmoore-sa:
Hi Rebecca, Dylan, 

This is a small PR that deprecates the TreatyOrganization class in favor of IntergovernmentalOrganization, adjusts the definition and example of the latter, and makes a pref-label fix to remove a '-'. It also adds the relevant information to the release notes.

Let me know if you don't have time and I can switch reviewers. Aiming for the end of the month.

Thanks,
Heather

Fixes #766, fixes #756.

Issue #766, opened 2022-10-21
Remove distinction between Intergovernmental Organization and Treaty Organization


rjyounes:
According to [Harvard Law School](https://hls.harvard.edu/bernard-koteen-office-of-public-interest-advising/about-opia/what-is-public-interest-law/public-service-practice-settings/international-public-interest-law-practice-setting/intergovernmental-organizations-igos/):

The term intergovernmental organization (IGO) refers to an entity _created by treaty_, involving two or more nations, to work in good faith, on issues of common interest. In the absence of a treaty an IGO does not exist in the legal sense. For example, the G8 is a group of eight nations that have annual economic and political summits. IGOs that are formed by treaties are more advantageous than a mere grouping of nations because they are subject to international law and have the ability to enter into enforceable agreements among themselves or with states...Among the oldest IGOs are the [United Nations](http://www.un.org/en/), which replaced the League of Nations, the [Universal Postal Union](http://www.upu.int/), and the [North Atlantic Treaty Organization](http://www.nato.int/cps/en/natolive/index.htm) (NATO).

Other sources agree:

[Duke Law School](https://law.duke.edu/ilrt/int_orgs_1.htm): "International governmental organizations (or IGOs) are _associations of states established by a treaty_ to pursue the common aims of their member states. An intergovernmental organization has a legal personality separate from its member states, and can enter into legally binding agreements with other IGOs or with other states. The [United Nations](http://www.un.org/) and the [Organization of American States](http://www.oas.org/) are examples of international governmental organizations.
The other category of international organization is the non-governmental organization (or NGO). NGOs are made up of private citizens. NGOs do not enter into treaties or other international agreements. The [World Wildlife Fund](http://www.worldwildlife.org/), [Greenpeace](http://www.greenpeace.org/international/) and [Amnesty International](http://www.amnesty.org/) are examples of non-governmental organizations."

(Note the interesting distinction between IGOs and NGOs.)

[Wikipedia](https://en.wikipedia.org/wiki/Intergovernmental_organization): "An intergovernmental organization or intergovernmental organisation, also known as an international institution, is an organization composed primarily of sovereign states, or of other organizations _through formal treaties_ for handling/serving common interests and governed by international laws."

(All emphases mine.)

It appears that "International Governmental Organization" is the most commonly used term, as opposed to "Treaty Organization."

rjyounes:
Fix wtih #756.

Jamie-SA:
[...]

PR #828, opened 2023-04-28
added TimeInterval class

kchastain703:
Adding the TimeInterval class back to gistCore

PR #877, opened 2023-06-16
Remove TreatyOrganization (accidentally left in)

stevenchalem:
Peter found at the last minute that TreatyOrganization was still present in gistCore.ttl. This PR is to remove it consistent with the intention of issue #766 and pr #825

### DIFF
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -1261,9 +1261,9 @@ gist:IntergovernmentalOrganization
 			]
 		) ;
 	] ;
-	skos:definition "An organization whose members are government organizations."^^xsd:string ;
-	skos:example "The United Nations, the European Union"^^xsd:string ;
-	skos:prefLabel "Inter-Governmental Organization"^^xsd:string ;
+	skos:definition "An organization whose members are government organizations. This can comprise regional, municipal, state/province, or national level entities."^^xsd:string ;
+	skos:example "The United Nations, the European Union, the MTA (Metropolitan Transit Authority)"^^xsd:string ;
+	skos:prefLabel "Intergovernmental Organization"^^xsd:string ;
 	.
 
 gist:Landmark
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -1140,10 +1140,7 @@ gist:GovernedGeoRegion
 gist:GovernmentOrganization
 	a owl:Class ;
 	rdfs:subClassOf gist:Organization ;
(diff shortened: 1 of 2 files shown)


stevenchalem:
@dylan-sa does this look right?

dylan-sa:
Getting close to what we need, but looks like Heather's changes to the annotations for `IntergovernmentalOrganization` didn't make it in, either: https://github.com/semanticarts/gist/pull/825/files

If you change the prefLable, definition, and example for `IntergovernmentalOrganization`, I think we'll have what we need.

dylan-sa (changes requested):
Just need to change the annotations for `IntergovernmentalOrganization`

stevenchalem (commented):
requested changes are complete

dylan-sa:
I did some more digging here and believe I found the root of the issue.

It looks like one of my [commits](https://github.com/semanticarts/gist/commit/b492805ada640d9169b7be4dbe15344585a036cb) on the inverses branch was not caught up with some of the work that came right before it. (Surprising to me since I had made a point to merge the develop branch into my inverses branch, but it appears that not everything was updated as expected.) So, I've looked to see what exactly got lost, and I believe this is what we need to get caught up:

- [x] Remove `TreatyOrganization` (@stevenchalem has addressed in this PR)
- [x] Update annotations for `IntergovernmentalOrganization` (already addressed in this PR)
- [ ] Remove `TimeZone`, `TimeZoneStandard`, `hasOffsetToUniversal`, and `usesTimeZoneStandard` (per #815)
- [ ] Update annotation on `atDateTime` as done in #815
- [ ] Add `TimeInterval` class as done in #828 

As best I can tell, this would fill in the gaps, but I would appreciate some more eyes on this just to make sure I've got it right. (@stevenchalem @rjyounes @Jamie-SA, if you could please take a look.) If this is looking right, I can make the remaining changes since my branch appears to be the culprit.

Jamie-SA:
I retract my last statement... do to something I was working on, I was not looking at the correct version of gistCore.ttl.

stevenchalem:
@dylan-sa I think you've caught everything.

Jamie-SA:
I submitted PR #879 to re-apply the TimeZone changes. Steven approved, and I merged it.
I have submitted PR #880 to re-apply the lost TimeInterval commit.

Jamie-SA:
Hopefully everything is fixed now...

PR #879, opened 2023-06-19
Re-apply previous commit: Remove TimeZone & etc

Jamie-SA:
One scopeNote change from this PR was lost. This is an attempt to fix that with git magic.


Jamie-SA:
I used git to re-apply my previous change. It shows the complete diff but I am hopeful it will work correctly and just get the scopeNote change that was lost.

stevenchalem (approved):

Jamie-SA:
I am going to merge this and then I will double check to verify it did what I was hoping.

PR #881, opened 2023-06-20
Release/12.0.0

stevenchalem:


gistCore.ttl
```diff
@@ -1,17 +1,19 @@
-@prefix : <https://ontologies.semanticarts.com/o/gistCore#> .
-@prefix gist: <https://ontologies.semanticarts.com/gist/> .
+@prefix : <https://w3id.org/semanticarts/gistCore#> .
```
dylan-sa, line 1: Should this blank prefix for the ontology IRI be in here? Similar question for the xml one, since there are no xml IRIs being used in gistCore.

dylan-sa (changes requested):

Jamie-SA:
I guess a review from me is dependent upon whether we decide to make any changes to the IRIs as a result of the discussion this morning.

stevenchalem:
I'm putting this on hold until after Thursday's gist dev meeting.

PR #911, opened 2023-06-30
Update IRIs in Namespace.md

stevenchalem:


stevenchalem:
incorrect PR. will re-create

PR #941, opened 2023-07-27
Deprecated properties related to occupying a location.

### LINKED ISSUE #809: Remove `occupiesGeographically` and `occupiesGeographicallyPermanently`?
During the inverse working group meeting, we noticed a potential redundancy with the following predicates.

- Currently, we have `isGeographicallyOccupiedBy` and its inverse, `occupiesGeographically`. We also have `isGeographicallyPermanentlyOccupiedBy` and its inverse, `occupiesGeographicallyPermanently`.
- While not formally specified, these predicates are likely intended to relate objects to spatial regions. (`isGeographicallyContainedIn`, a further property not on the list above, is likely intended to relate spatial regions to spatial regions.)
- Arguably, `hasPhysicalLocation` could do the work that `occupiesGeographically` does, and for this reason we may want to remove `occupiesGeographically` and its inverse to avoid redundancy.
- If we were to remove `occupiesGeographicallyPermanently` and its inverse, there is a question as to whether we would want to add an analogous `hasPermanentPhysicalLocation`.

uscholdm:
Fixes #809.

See https://github.com/semanticarts/gist/issues/947 for what to do on the next major release.

### Minor Updates

- Deprecated properties related to occupying a location. Issue [#809](https://github.com/semanticarts/gist/issues/809). Changes:
  - Deprecated properties: `gist:occupiesGeographically` and `gist:occupiesGeographicallyPermanently`.
  - Updated annotations for `gist:hasPhysicalLocation` regarding location temporality.
  - Added editorialNote on `gist:Landmark` to update the restriction on the next major release.


gistCore.ttl
```diff
@@ -3213,8 +3213,9 @@ gist:hasPhysicalLocation
 		owl:TransitiveProperty
 		;
 	rdfs:range gist:Place ;
-	skos:definition "Where something is located"^^xsd:string ;
+	skos:definition "Relates something to where it is physically located."^^xsd:string ;
 	skos:prefLabel "has physical location"^^xsd:string ;
+	skos:scopeNote "This property does not distinguish between things whose locations vary over different time periods. E.g. a fire hydrant vs. a car."^^xsd:string ;
```
rjyounes, line 3218:
```suggestion
	skos:scopeNote "This property does not distinguish between things whose locations are relatively stable and those that vary over different time periods; e.g., a fire hydrant vs. a car. Use a temporal relation if you need to explicitly model a time-delimited location."^^xsd:string ;
```

The first sentence is a grammar adjustment. The second sentence is added content which you may or may not agree with - I don't have strong feelings so don't mind if you drop it.

gistCore.ttl
```diff
@@ -3555,27 +3556,6 @@ gist:numericValue
 	skos:prefLabel "numeric value"^^xsd:string ;
 	.
 
-gist:occupiesGeographically
```
rjyounes, line 3558: I apologize for not thinking through this when we discussed the implementation. We have a new strategy for deprecation: https://github.com/semanticarts/gist/blob/develop/docs/DeprecationAndDeletionPolicy.md. The upshot is:
1. Rather than deleting the terms, we mark them as `owl:deprecated true` but keep them in the gistCore file (no gistDeprecated.ttl).
2. The restrictions stay the same for now in order to maintain backward compatibility. 
3. On the next major release, we will delete the terms and adjust the restriction.

docs/release_notes/Issue-809-occupiesGeographically.md
```diff
@@ -0,0 +1,6 @@
+### Major Updates
+
+- Removed properties related to occupying a location. Issue [#809](https://github.com/semanticarts/gist/issues/809). Changes:
```
rjyounes, line 3:
```suggestion
- Deprecated properties related to occupying a location. Issue [#809](https://github.com/semanticarts/gist/issues/809). Changes:
```

docs/release_notes/Issue-809-occupiesGeographically.md
```diff
@@ -0,0 +1,6 @@
...
+  - Removed properties: `gist:occupiesGeographically` and `gist:occupiesGeographicallyPermanently`.
+  - Updated restriction on `gist:Landmark` due to property being removed.
```
rjyounes, line 5:
```suggestion

```

rjyounes (changes requested):
Sorry about this, I forgot the current protocol for deprecation. I'm going to update the issue with the new implementation decision.

gistCore.ttl
```diff
@@ -1287,6 +1287,7 @@ gist:Landmark
 		) ;
 	] ;
 	skos:definition "Something permanently attached to the Earth."^^xsd:string ;
+	skos:editorialNote "On the next major release gist:occupiesGeographicallyPermanently will be gone. The restriction above will use gist:hasPhysicalLocation. Use subClassOf instead of equivalentClass."^^xsd:string ;
```
dylan-sa, line 1290: I'm refreshing my memory on the new deprecation policy. One part of the policy is:

> If the term slated for removal is referenced by other term definitions, as in a domain axiom or property restriction, such that a change to that reference constitutes a major change, the term cannot be deprecated in a minor release. The change or removal must wait until the next major release, when the references to it can be appropriately modified or removed.

In this case, the term slated for removal (`occupiesGeographicallyPermanently`) is used in a property restriction. If we were to use `hasPhysicalLocation` in the restriction instead, would that rise to the level of a major change, meaning that `occupiesGeographicallyPermanently` should not be deprecated in a minor release? Am I interpreting this part of the policy correctly? @rjyounes

dylan-sa (commented):
I just had one clarification question on the new deprecation policy. Otherwise looking good to me!

rjyounes:
I'm inclined to hold hard to the line that any change in inferencing outcomes constitutes a major change; otherwise, the criteria are too subjective. In software development, one would not say "Yeah, changing this method name breaks backward compatibility, but no one uses the method anyway, so let's call it a minor change." I recommend closing this PR and tagging the issue as "deferred to major release." Let's bring this up at a meeting because it's a significant issue that should be resolved as a group.

gistCore.ttl
rjyounes: Now I'm not sure I correctly captured our intent in the documentation. I think we might have said that we can deprecate the term, but not change the restrictions or other references. @marksem Do you recall?

rjyounes:
MarkW: Policy should be to deprecate (with OWL annotation) but keep any references as is.

Given this policy, the PR should not represent a major change. Michael will check.

Rebecca: check documentation.

docs/release_notes/Issue-809-occupiesGeographically.md
```diff
@@ -0,0 +1,6 @@
+### Minor Updates
+
+- Deprecated properties related to occupying a location. Issue [#809](https://github.com/semanticarts/gist/issues/809). Changes:
+  - Deprecated properties: `gist:occupiesGeographically` and `gist:occupiesGeographicallyPermanently`.
+  - Updated annotations for `gist:hasPhysicalLocation` regarding location temporality.
+  - Added editorialNote on `gist:Landmark` to update the restriction on the next major release.
```
rjyounes, line 6:
```suggestion

```
This doesn't need to be in the release note.

gistCore.ttl
rjyounes, line 1290:
```suggestion
	skos:editorialNote "See guidance on removing the term in the next major release at https://github.com/semanticarts/gist/issues/947#issuecomment-1679566885."^^xsd:string ;
```

This is the format we agreed on for the editorial notes. It goes without saying, according to our deprecation policy, that it will be removed on the next release, and I don't think the ontology should be littered with such comments, nor the documentation of exactly what changes will be made.

rjyounes (changes requested):

uscholdm:
> @uscholdm Please remember to add "Fixes #nnn" to the description of the PR so that when the PR is merged, the issue is automatically closed. This also helps readers easily find the issue.

Done

rjyounes (approved):

Issue #758, opened 2022-10-13
Update guidance on adding release notes to a PR


rjyounes:
New strategy for release notes. We need a new strategy for handling release notes, as there are many merge conflicts that can be time-consuming to resolve. 

Proposal: 
- Create a directory for release notes in the repository
- Each PR includes a new Markdown file in this directory named  `prXXX.md`, where `XXX` is the number of the PR. The note should indicate whether it's a major, minor, or patch change.
- During the release process, on the release branch, the release manager will copy these notes to the relevant section of `ReleaseNotes.md` and delete the files.

TO DO:
* Update documentation on creating a PR in`Contributing.md`.
* Update Semantic Arts internal documentation on creating a release. 
* Create the `releaseNotes` folder.

Issue #753, opened 2022-10-09
Reword definition of gist:nonConformingLabel


rjyounes:
Current: "The annotated ontology entity is excluded from label validation rules."

Change to follow our verb-first convention: "Indicates that the annotated entity is excluded from label validation constraints."

Issue #734, opened 2022-09-25
Additions to Contributing.md


rjyounes:
If not already included:

* When you start working on an issue, move to "In Progress" in the project board. There is no automation of this change in state.
* When you have submitted the PR for an issue, move the issue to "In Review" in the project board. There is no automation of this change in state.
* When you have submitted your PR, check that there are no conflicts with develop. If there are, you should resolve them.

Jamie-SA:
@Jamie-SA will take a first whack for Rebecca to edit.

Jamie-SA:
I don't see anything about how we expect release notes are expected to be added to the /docs/ directory and then combined for the release.

Jamie-SA:
Having to name the release notes with the PR number means you can't have the right filename until after you create the PR. Is there a better way to get around this?

rjyounes:
> I don't see anything about how we expect release notes are expected to be added to the /docs/ directory and then combined for the release.

I have been lax in documenting this. I will try to get that done soon, if not I can tell Steven how to do it.

> Having to name the release notes with the PR number means you can't have the right filename until after you create the PR. Is there a better way to get around this?

Yes, that's true. You could use the issue number. The filename is inconsequential, so you can really name it whatever you want.

PR #971, opened 2023-09-14
Fix typo in scope note for `_percent`

### LINKED ISSUE #961: Typo in scope note for `_percent`
Current scope note for `gist:_percent`: "Because the units of the numerator and denominator are the same, they cancel out, resulting in a dimensional quantity."

Should be 'dimensionless' rather than 'dimensional'.

dylan-sa:
Closes #961

### DIFF
--- a/docs/release_notes/issue-961-percent-typo.md
+++ b/docs/release_notes/issue-961-percent-typo.md
@@ -1,0 +1,3 @@
+### Patch Updates
+
+- Fixed typo in scope note for `gist:_percent`: changed 'dimensional' to 'dimensionless'. Issue [#961](https://github.com/semanticarts/gist/issues/961).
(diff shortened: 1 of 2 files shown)


docs/release_notes/issue-961-percent-typo.md
```diff
@@ -0,0 +1,3 @@
+### Patch Updates
+
+- Fixed typo in scope note for `gist:_percent`: changed 'dimensional' to 'dimensionless'. Issue [#961](https://github.com/semanticarts/gist/issues/961).
```
rjyounes, line 3: I think this is overkill. We should probably say something like "Made various minor wording corrections to annotations." In the 12.0.0 release notes we said

- Miscellaneous non-semantic annotation updates. Issues [#734](https://github.com/semanticarts/gist/issues/734),[#753](https://github.com/semanticarts/gist/issues/753), [#758](https://github.com/semanticarts/gist/issues/758).

I'm not sure how to do this with our new approach of using separate files for release notes. Perhaps you could add a note for the release manager on adding similar items to the list of issues without specifying the particular fix in each case.

rjyounes (changes requested):
The actual fix is fine; I recommend changes to the release note.

rjyounes (approved):

Issue #947, opened 2023-08-10
Address deprecated concepts for release 13.0.0


uscholdm:
This issue spells out changes that need to be made regarding deprecated items in preparation for the next major release: 13.0.0

uscholdm:
## PR https://github.com/semanticarts/gist/pull/941, Deprecated properties related to occupying a location

1. Remove the two deprecated properties
  - occupiesGeographically
  - occupiesGeographicallyPermanently
2. Change the restriction on Landmark to use `hasPhysicalLocation` rather than `occupiesGeographicallyPermanently`. Also, use subClassOf instead of equivalentClass.

```
gist:Landmark
	a owl:Class ;
	rdfs:subClassOf
		gist:PhysicalIdentifiableItem ,
		gist:Place ,
		[
			a owl:Restriction ;
			owl:onProperty gist:hasPhysicalLocation ;
			owl:someValuesFrom [
				a owl:Class ;
				owl:unionOf (
					gist:GeoRegion
					gist:GeoVolume
				) ;
			] ;
		]
		;
	skos:definition "Something permanently attached to the Earth."^^xsd:string ;
	skos:prefLabel "Landmark"^^xsd:string ;
	.
```

uscholdm:
##  PR https://github.com/semanticarts/gist/pull/951 Remove isCharacterizedAs

Remove the deprecated property, `isCharacterizedAs`
Remove the restriction on `Event` that uses `isCharacterizedAs`

The new version of `Event` will look like this:

```
gist:Event
	a owl:Class ;
	skos:definition "Something that occurs over a period of time, often characterized as an activity being carried out by some person, organization, or software application or brought about by natural forces."^^xsd:string ;
	skos:example "A transaction, conference, baseball game, earthquake."^^xsd:string ;
	skos:prefLabel "Event"^^xsd:string ;
	skos:scopeNote "An event does not necessarily have either planned or actual start or end datetimes. For example, a conference can be in the planning phase without any dates selected, but is nevertheless an (unscheduled) event. The subclasses of Event state particular restrictions on planned and actual start and end dates."^^xsd:string ;
	.
```

Consider whether anything else is needed/desired for Event.

uscholdm:
##  PR https://github.com/semanticarts/gist/pull/951 Remove tagText

- `Tag`: change restriction
  - to use `containedText` instead of `tagText`
  - to use subClass rather than equivalentClass
- Make  `uniqueText` a subproperty of `containedText`
[...]

PR #989, opened 2023-10-26
Issue 372 text datatype properties

### LINKED ISSUE #372: Review and refactor :hasTag and other text datatype properties, and consider making uniqueText and tagText subproperties of containedText.
In conformance to our naming standard, we should refactor the data property :hasTag as :tagText or something similar.

uscholdm:
Fixes #372 

- Deprecated `tagText`, to be replaced by `containedText`.
- `containedText` 
  - Added domainIncludes `Tag` and `Text` 
  - Removed reference to Text class in the definition
  - Added examples

### DIFF
--- a/docs/release_notes/372-text-datatype-properties.md
+++ b/docs/release_notes/372-text-datatype-properties.md
@@ -1,0 +1,8 @@
+### Minor Updates
+Issue [#372](https://github.com/semanticarts/gist/issues/372). Changes:
+  
+- Deprecated `tagText`, to be replaced by `containedText`.
+- `containedText` 
+  - Added domainIncludes `Tag` and `Text` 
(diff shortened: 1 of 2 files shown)


rjyounes (changes requested):
As discussed, this PR should not include the deprecation. Please add it as a new issue for triage.

uscholdm:
> As discussed, this PR should not include the deprecation. Please add it as a new issue for triage.

That is not my understanding.  This deprecation signals that the property will be removed on next major release.  It is a minor update with instructions for what to do on the release ([here]( https://github.com/semanticarts/gist/issues/947#issuecomment-1780220876)).  I think it's ok as it is.

rjyounes (approved):

PR #1005, opened 2023-11-08
Add scopeNote to uniqueText

uscholdm:
Add scopeNote to uniqueText explaining that uniqueness only goes in one direction.

### DIFF
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -3909,7 +3909,7 @@ gist:uniqueText
 	skos:definition "The unique string value of some content object; i.e., there is no possibility of having more than one value."^^xsd:string ;
 	skos:example "The unique string for a vehicle identification number."^^xsd:string ;
 	skos:prefLabel "unique text"^^xsd:string ;
-	skos:scopeNote "The uniqueness only goes in one direction. A case may arise where two different objects have the same unique text."^^xsd:string ;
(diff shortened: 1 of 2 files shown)


gistCore.ttl
```diff
@@ -3906,6 +3909,7 @@ gist:uniqueText
 	skos:definition "The unique string value of some content object; i.e., there is no possibility of having more than one value."^^xsd:string ;
 	skos:example "The unique string for a vehicle identification number."^^xsd:string ;
 	skos:prefLabel "unique text"^^xsd:string ;
+	skos:scopeNote "The uniqueness only goes in one direction. A case may arise where two different objects have the same unique text."^^xsd:string ;
```
rjyounes, line 3912: This sounds like it is a remote case, but I don't think it is necessarily, given that IDs can represent all sorts of things.

rjyounes (changes requested):

gistCore.ttl
uscholdm: Can you give a specific and realistic example of two different objects that have the same unique text? I could not think of any. The wording reflected this.

rjyounes: Two organizations issue employee IDs. Both employee A at org 1 and employee B at org 2 have ID "12345." I consider this to be a potentially common occurrence.

uscholdm:
Thanks, that's a good example, in principle, but  Im not sure how often o

rjyounes (approved):

Issue #891, opened 2023-06-23
Refactor pre-commit hook to always use the latest version


Jamie-SA:
Original Title: Warn user if the pre-commit hook is not the latest version

Original Description: Have the pre-commit hook check that it is the same as the one at `tools/pre-commit`. If it isn't, tell the user to run the setup.cmd and abort.

New Description - see PR #1359

rjyounes:
Good idea.

Jamie-SA:
Also need to do the same thing for `${base_dir}/.git/hooks/pre-commit-serializer`.

rjyounes:
In PR #1012 I've undone this change and the serializer pre-commit is now left in place; the problem was solved in another way. So you don't have to do the check for latest version on the serializer pre-commit.

Jamie-SA:
I just submitted a new PR for this Issue.

PR #1007, opened 2023-11-09
Fix serializer pre-commit permissions problem

### LINKED ISSUE #1006: Fix pre-commit permissions problem
Perhaps a Mac-only problem: Serializer pre-commit hook keeps changing permissions, which git considers a change and wants to commit. This happens repeatedly and is a real nuisance. The solution is to copy the serializer pre-commit to the hooks directory and invoke that one from the main pre-commit; the one in the repo is then not being executed.

rjyounes:
Fixes #1006

### DIFF
--- a/docs/release_notes/1007-pre-commit
+++ b/docs/release_notes/1007-pre-commit
@@ -1,0 +1,3 @@
+### Patch Updates
+
+- Fixed serializer pre-commit hook permissions problem on MacOS.
--- a/tools/pre-commit
+++ b/tools/pre-commit
@@ -21,9 +21,8 @@ base_dir=$(git rev-parse --show-toplevel)
 # Run the serializer
 set -x
 
-# Ensure that the serializer pre-commit is executable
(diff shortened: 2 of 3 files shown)


Jamie-SA (approved):
I am not real thrilled with this change. This means that the hook will not automatically be updated when changed in the git repo.

However, there is already issue #891 about having to deal with this for the main hook, so it could be dealt with in the same way.

rjyounes:
> This means that the hook will not automatically be updated when changed in the git repo.

If the main pre-commit changes in the repo you have to rerun `/tools/setup.cmd`. This handles changes to the serializer pre-commit the same way, unless I'm missing something.

Jamie-SA:
Yes, but see the issue I referenced: #891 
You don't currently have to run setup.cmd if the serialize/pre-commit changes, but you will after this change.

rjyounes:
I see. I see you added a comment to that effect on #891. It seems to me to be in principle the same issue.

PR #1008, opened 2023-11-09
Fixed definition of Magnitude.

### LINKED ISSUE #965: Fix definition of gist:Magnitude
Definition is now "Base class for units which can be converted.  The primitive units can be converted from one measurement system to another; the complex units (ratio or product) have to decompose to their primitives."

Seems to be a copy-paste error. I think we can fix this before the major units and magnitudes refactoring, because unless I'm missing something it's an error.

uscholdm:
Fixes #965 
It was likely a copy/paste error.

### DIFF
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -1403,7 +1403,8 @@ gist:Magnitude
 			]
 		) ;
 	] ;
-	skos:definition "Base class for units which can be converted.  The primitive units can be converted from one measurement system to another; the complex units (ratio or product) have to decompose to their primitives."^^xsd:string ;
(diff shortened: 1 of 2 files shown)


gistCore.ttl
```diff
@@ -3194,7 +3195,10 @@ gist:hasMember
 	a owl:ObjectProperty ;
 	skos:definition "Relates something, such as a collection or organization, to its member individuals."^^xsd:string ;
 	skos:prefLabel "has member"^^xsd:string ;
-	gist:domainIncludes gist:Collection , gist:Organization
+	gist:domainIncludes
```
rjyounes, line 3198: Funny, looks like the serializer hadn't run previously.

rjyounes (approved):

Issue #934, opened 2023-07-26
Add skeletal definitions of SKOS and SHACL terms used in gist in order to achieve DL-compliance


rjyounes:
gist is not DL-compliant because it uses undeclared terms in the SHACL and SKOS namespaces. Add skeletal definitions (à la Protégé) to achieve DL compliance.

rjyounes:
need to remove this from the pre-commit hook:

```
function remove_skos_decls () {
  for fileToBeCommitted in $(git diff --cached --name-only) ; do
    permissions=$(stat -c "%a" "${fileToBeCommitted}")
    sed -e '/^skos/,+3d' "$fileToBeCommitted" > tmp$$
    mv tmp$$ "$fileToBeCommitted"
    chmod $permissions "${fileToBeCommitted}"
    git add --update "${fileToBeCommitted}"
  done

  return 0
}
```

dylan-sa:
I like the idea of adding the skeletal definitions to achieve DL compliance, but it's worth considering what they should look like. As @marksem has pointed out, Protege automatically adds the following definitions for gist:

dylan-sa:
I like the idea of adding skeletal definitions to be DL-compatible, but it's worth thinking more about where the defs should come from and what they should look like. As @marksem has pointed out, Protege automatically generates the following definitions for gist:

```
#################################################################
#    Annotation properties
#################################################################
###  http://www.w3.org/2004/02/skos/core#definition
skos:definition rdf:type owl:AnnotationProperty .

###  http://www.w3.org/2004/02/skos/core#example
skos:example rdf:type owl:AnnotationProperty .

###  http://www.w3.org/2004/02/skos/core#prefLabel
skos:prefLabel rdf:type owl:AnnotationProperty .

###  http://www.w3.org/2004/02/skos/core#scopeNote
skos:scopeNote rdf:type owl:AnnotationProperty .

###  http://www.w3.org/ns/shacl#declare
sh:declare rdf:type owl:AnnotationProperty .

###  http://www.w3.org/ns/shacl#namespace
sh:namespace rdf:type owl:AnnotationProperty .

###  http://www.w3.org/ns/shacl#prefix
sh:prefix rdf:type owl:AnnotationProperty .
```
[...]

PR #1011, opened 2023-11-20
Remove remove_skos_decls function from serializer pre-commit

### LINKED ISSUE #1010: Some files get corrupted after running through the pre-commit
When committing files like`.pdf` or `.png`, the file will become corrupted. If you commit the same files with `--no-verify`, skipping the pre-commit, the file will not be corrupted. Something in the pre-commit is causing these files to become corrupt.

rjyounes:
Fixes #1010.

The function corrupted binary files like PDFs and PNGs because it was called on any file, even those skipped by the serializer. Removing this function rather than modifying the call since we are going to remove it anyway as part of #934.

### DIFF
--- a/docs/release_notes/1007-pre-commit
+++ b/docs/release_notes/1007-pre-commit
@@ -1,0 +1,3 @@
+### Patch Updates
+
+- Fixed serializer pre-commit hook permissions problem on MacOS.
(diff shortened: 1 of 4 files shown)


rjyounes:
Closing - updates from an earlier PR were inadvertently included here. I'll merge from develop and submit a new PR.

PR #1012, opened 2023-11-21
Ignore executable flag on files

rjyounes:
@Jamie-SA On Mac files become non-executable when committed. This means that you have to make setup.cmd executable to run it, at which point git records a change and wants to re-commit. Setting the flag in setup.cmd means git will ignore this change. 

See https://git-scm.com/docs/git-config#Documentation/git-config.txt-corefileMode and
https://stackoverflow.com/questions/1580596/how-do-i-make-git-ignore-file-mode-chmod-changes

### DIFF
--- a/tools/setup.cmd
+++ b/tools/setup.cmd
@@ -27,7 +27,7 @@ cp "${base_dir}/tools/serializer/pre-commit-serializer" "${base_dir}/.git/hooks/
 # Ensure that the serializer pre-commit hook is executable.
 chmod +x "${base_dir}/.git/hooks/pre-commit-serializer"
 
-# Don't track executable flag on files. Needed on Mac, which makes files non-executable on commit. 
+# Don't track executable flag on files; this is a repository-specific setting. Needed on Mac, which makes files non-executable on commit. 
 git config core.filemode false
 
 # Exit linux shell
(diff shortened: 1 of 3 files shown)


tools/setup.cmd
```diff
@@ -19,15 +19,15 @@ set -x
 # Copy pre-commit to the git hooks directory
 cp "${base_dir}/tools/pre-commit" "${base_dir}/.git/hooks/"
 
-# Make pre-commit hook executable. Needed at least on Mac, where the copied file is non-executable unless there's an existing executable file in the hooks directory, regardless of the status of the original file.
+# Make pre-commit hook executable. 
 chmod +x "${base_dir}/.git/hooks/pre-commit"
 
 # Copy serializer pre-commit to hooks directory.
 cp "${base_dir}/tools/serializer/pre-commit-serializer" "${base_dir}/.git/hooks/pre-commit-serializer"
```
dylan-sa:
```suggestion
cp "${base_dir}/tools/serializer/pre-commit" "${base_dir}/.git/hooks/pre-commit-serializer"
```
This is what we need, right?

coltonglasgow (approved):

tools/setup.cmd
rjyounes: No, we're not doing the copy, but running it in place - see the main pre-commit.

PR #1026, opened 2024-01-25
sorted typo

pwin:


rjyounes (changes requested):
Technically we need a release note even though this is a tiny change. I usually combine these into a single patch note along the lines of "Corrected miscellaneous typos." No need to cite an issue number. Then the release manager should combine similar release notes into a single one.

rjyounes (approved):

PR #1032, opened 2024-02-05
Changed restriction on gist:LivingThing

### LINKED ISSUE #806: Change minCardinality 1 restriction on LivingThing to someValuesFrom
minCardinality of 1 is equivalent to someValuesFrom, and the latter has computational advantages (@uscholdm can explain what these are). This is the last remaining restriction of this sort in gist. minCardinality 2 and exact cardinality should be retained, as they provide more precision than someValuesFrom.

rjyounes:
Closes #806

### DIFF
--- a/gistCore.ttl
+++ b/gistCore.ttl
@@ -12,9 +12,9 @@
 	a owl:Ontology ;
 	owl:versionIRI <https://w3id.org/semanticarts/ontology/gistCoreX.x.x> ;
 	skos:definition "gist is a minimalist upper ontology created by Semantic Arts."^^xsd:string ;
+	skos:editorialNote "Information about the standard namespace prefix is in gistNamespaceInformation.ttl."^^xsd:string ;
 	skos:prefLabel "gist"^^xsd:string ;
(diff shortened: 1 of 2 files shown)


gistCore.ttl
```diff
@@ -12,9 +12,9 @@
 	a owl:Ontology ;
 	owl:versionIRI <https://w3id.org/semanticarts/ontology/gistCoreX.x.x> ;
 	skos:definition "gist is a minimalist upper ontology created by Semantic Arts."^^xsd:string ;
+	skos:editorialNote "Information about the standard namespace prefix is in gistNamespaceInformation.ttl."^^xsd:string ;
```
uscholdm, line 15: I cannot find any file called "gistNamespaceInformation.ttl"

uscholdm (commented):
Fine except that I cannot find any file called "gistNamespaceInformation.ttl".

gistCore.ttl
rjyounes: There's a story here that you don't want to know about...I will remove the line.

PR #1040, opened 2024-02-12
post 531 actions - deprecating gist:symbolUnicode and gist:symbolHtml

### LINKED ISSUE #992: Remove symbolUnicode and symbolHtml? (post-531 implementation)
Note: these predicates do not currently exist, but will exist after the implementation of issue #531. #531 involves renaming to make them applicable to things other than units. This is about whether we need the HTML and Unicode versions of the symbol predicate.

pwin:
Fixes #992.


uscholdm (changes requested):
These should be removed, not deprecated. Here is why.

This is very confusing.  I'm not exactly sure what is going on. @rjyounes  It looks like the net effect of what we ended up doing was to rename `unitSymbol` to `symbol` and to remove `unitSymbolUnicode` and `unitSymbolHtml`, but we did it in two steps:
1. First we 'renamed' `unitSymbol`, and `unitSymbolUnicode` and `unitSymbolHtml`, to `unitSymbol`, `symbol` and `symbolUnicode` and `symbolHtml`,.  By rename, I mean we deprecated the original and created new properties.
2. We removed `symbolUnicode` and `symbolHtml`

Normally we do not remove things except on a major release, but in this case we are not removing anything we added during this release, we are removing something we created during this release that we then removed.

uscholdm (approved):

docs/release_notes/531-extend-symbol-predicates.md
```diff
@@ -4,3 +4,4 @@
   - Deprecated `gist:unitSymbol`, `gist:unitSymbolHtml` and `gist:unitSymbolUnicode`. 
   - Added predicates: `gist:symbol`, `gist:symbolHtml` and `gist:symbolUnicode`.
   - Added symbol triples for existing units, anticipating the removal of `gist:unitSymbol`.
+  - Deleted  `gist:symbolUnicode` and `gist:symbolHtml` since they were added within this change - see above
```
rjyounes, line 7:
```suggestion
  - Added predicate `gist:symbol`.
  - Added symbol triples for existing units, anticipating the removal of `gist:unitSymbol`.
```

Note: you don't need to mention deletion when they were added in this release. From the user's perspective they simply never existed. The actual changes are the deprecation of 3 predicates and the addition of one.

rjyounes (changes requested):

docs/release_notes/531-extend-symbol-predicates.md
rjyounes: Note also that line 3 should have "Extended" rather than "Extending."

rjyounes (approved):

PR #1052, opened 2024-02-23
Run subclass inference during bundling

### LINKED ISSUE #1051: Add generating subclass assertion functionality into bundle.yaml
Don't need a separate script.

sa-bpelakh:
Fixes #1051 

When this merges, we should remove the instructions from https://semarts.atlassian.net/wiki/spaces/OF/pages/1126760539/gist+Release+Management+Ontologists#Materialize-Subclass-Inferences


dylan-sa (changes requested):
Working perfectly on my end. Thanks for helping with this @bpelakh!

tools/subclass_inferences/materialize_subclass_inferences.py
```diff
@@ -1,70 +1,104 @@
-import os
+"""Generate subClassOf inferences to help OWL-RL reasoners."""
```
rjyounes, line 1:
```suggestion
"""Materialize rdfs:subClassOf inferences to help OWL-RL reasoners."""
```

tools/subclass_inferences/materialize_subclass_inferences.py
```diff
@@ -1,70 +1,104 @@
...
+        (SKOS.prefLabel, Literal("gist Subclass Assertions",
+                                 datatype=XSD.string)),
+        (SKOS.scopeNote, Literal("This ontology contains supplementary subclass "
+                                 "assertions that are logically entailed by gistCore but "
+                                 "are not inferred by some automated reasoners. For example, "
+                                 "an OWL RL reasoner would not infer that gist:Commitment is a "
+                                 "subclass of gist:Intention, although it follows from the "
+                                 "ontology axioms. More precisely, it contains (1) subclass "
+                                 "assertions derived using an OWL DL reasoner and (2) the "
+                                 "subclass assertions that are already explicit in gistCore.",
```
rjyounes, line 40:
```suggestion
        (SKOS.scopeNote, Literal("This ontology contains supplementary subclass "
                                 "assertions to aid some automated reasoners.   For example, "
                                 "while an OWL DL reasoner infers from the class equivalence that "
                                 "gist:Commitment is a subclass of gist:Intention, an RL "
                                 "reasoner does not. More precisely, the ontology contains (1) "
                                 "direct subclass assertions derived using an OWL DL reasoner and "
                                 "(2) the subclass assertions that are already explicit in gistCore.",
```

@sa-bpelakh Can you confirm the following:
* (2) is still true (if not, remove it)
* The indirect subclass triples are filtered out (if not, please remove the word "direct")

tools/subclass_inferences/materialize_subclass_inferences.py
```diff
@@ -1,70 +1,104 @@
...
 
-    for prefix, namespace in input_graph.namespaces():
-        output_graph.bind(prefix, namespace)
+        output_graph.serialize(destination=output_ttl, format='turtle')
 
-    output_graph.serialize(destination=output_ttl, format='turtle')
+def _materialize_subclasses(argv: list[str]):
+    parser = ArgumentParser(
+        "Subclass Materializer",
+         description="Use OWL-DL reasoner to generate inferred rdfs:subClassOf triples. "
```
rjyounes, line 91:
```suggestion
         description="Use OWL DL reasoner to generate inferred rdfs:subClassOf triples. "
```

rjyounes (changes requested):

rjyounes (changes requested):
We need a release note. This may have changed since you were last here - consult the documentation in docs/Contributing.md.

Should we settle the question of whether to include the ontology definition or not? Or leave that until you move this over to the submodule? I'm fine with the latter.

rjyounes:
@sa-bpelakh I will approve this once the release note has been added. See comment above. Thanks.

rjyounes:
@sa-bpelakh I believe there is one more issue that hasn't been addressed: when running the script on any ontology that imports gistCore (or any ontology, for that matter), we would like to add gist to the graph but output only the subclass assertions for the classes defined in the lower ontology. This should be generalizable to any number of imports. 

One thought I had was to load the ontology you want to generate inferences for into a different graph (possibly in addition to the main graph that will be inferred over) in order to reference the classes defined there when generating the output file - I.e., only output inferences with subject in that second graph. We can't rely on namespace because gistCore shares a namespace with all the sub-gists.

rjyounes:
Adding myself as assignee to write a release note.

rjyounes (approved):

Issue #980, opened 2023-10-02
Magnitude has an infinite regress of precisions


uscholdm:
Currently every `Magnitude` has a precision that is also a `Magnitude`. So if I weigh myself as 80kg on my scale at home that is accurate to 500g, then the precision is the `Magnitude` that represents 500g. But according to gist, that `Magnitude` also has a precision. I cannot think what it might mean for a precision to have a precision. 

The problem is that the precision characterizes the weighing event and the accuracy of the scale, not the magnitude itself.  The magnitude itself is just a number and a unit; it's a mathematical construct used to describe the physical world.  If 10 different people weigh in at 80kg on 10 different scales, with different accuracies, there need only be a single magnitude instance for 80kg, and the 10 measuring events would each have a precision based on the scale and possibly other factors. 

Proposal:
- remove the `hasPrecision` restriction from `Magnitude`
- keep the property `hasPrecision` and give it a scope note indicating how it might be used. For example.

``` 
skos:scopeNote "the event of measuring a person's weight at the doctor's office may have an associated precision of 1 pound"^^xsd:string ; 
```

dylan-sa:
Yes, I agree that precisions aren't inherent to magnitudes; it's probably particular measurements (like weighing events) that have precisions, as you suggest. I'd be in favor of removing the restriction.

I see some discussion of a `gist:Measurement` class here: https://github.com/semanticarts/gist/issues/18. I'm not advocating that we add the class, but may be worth considering.

sa-bpelakh:
Agree w/ @dylan-sa , precision is an aspect of a measurement (or even a sensor, in an IoT world), and does not belong on every Magnitude. In case of use case where measurements vary widely in precision, a sub-class `PreciseMagnitude` can be defined which bakes precision into every instance.

philblackwood:
Precision is actually precision of the measurement method, but can also be interpreted as likelihood of error in the value of the measurement. 

I think this is resolved by the interpretation given with hasPrecision in the new Units of Measure:

skos:scopeNote "A typical way to use precision is to have it represent 2 standard deviations of the distribution of measurement errors. With this convention, when the measurement method is well-calibrated (has an average error of zero), there is a 95% chance that the actual error in measurement, in either direction, is less than the precision.

(also implies the measurement method results in a normal distribution of errors)

uscholdm:
> Precision is actually precision of the measurement method, but can also be interpreted as likelihood of error in the value of the measurement.

[...]

PR #1064, opened 2024-03-05
gistCore changes to support Units of Measure and Magnitudes

### LINKED ISSUE #1033: Turtle implementation of new units and measures model
Open questions:
* Is gist:Magnitude the extended or "clean" version, and what do we call the other one?
* How many and which namespaces? (gist:, gistd: for reference data?)

### LINKED ISSUE #1065: Options for expressing the broader relationship among aspects
@uscholdm @philblackwood 

In the new uom model, aspects are organized into a hierarchy.  Options for the linking relationship, including pros and cons:

1. `skos:broader`
  * Note: `skos:broader` has been proposed, but we actually want the transitive version, `skos:broaderTransitive`.
  * Broadly disliked because it doesn't indicate direction.
  * Up to now we have not wanted to import another ontology into gist. gist is an upper ontology - should it therefore be self-contained?

2.  New property `gist:hasBroader`
  * Con: Easily confused with `skos:broader`? - especially if the gist one is transitive.

3.  New property with some other name
  * Pro: Addresses problems with both 1 and 2.

4. Reuse `gist:hasSuperCategory`
  * Pro: Doesn't involve using SKOS or defining a new predicate.
  * Con: Although no domain and range are specified, the name suggests it only applies to categories, and therefore will be confusing to use with aspects. 
  
5. Rename `gist:hasSuperCategory` (major change)
  * Use a name to indicate hierarchical, broader/narrower relationships that does not suggest categories.
  * Pro: Gets the advantages of 4 and solves the issues with all those above.
  * This may bring us back to `gist:hasBroader`.
  * Con: Changing the name of a long-standing property.

philblackwood:
Adding two classes and one property. Fixes #1033.
Closes #1065.


ontologies/gistCore.ttl
```diff
@@ -2529,6 +2537,26 @@ gist:Transaction
...
+				owl:someValuesFrom gist:UnitOfMeasure ;
+			]
+		) ;
+	] ;
+	owl:disjointWith gist:UnitOfMeasure ;
+	skos:definition "A collection of Units of Measure that can all be used to measure the same Aspects."^^xsd:string ;
+	skos:example "Bit, byte, kilobit, kilobyte, etc. can all be used to measure an amount of data."^^xsd:string ;
+	skos:prefLabel "Unit Group"^^xsd:string ;
+	skos:scopeNote "When adding a unit of measure to a Unit Group, make sure it has the same exponents as the other members of the Unit Group. For example, miles per gallon and gallons per mile could both be used to measure vehicle efficiency, but they need to be in different unit groups."^^xsd:string ;
+	.
```
uscholdm, line 2558: As I understand it, the idea of a UnitGroup is a good idea serving as a convenience to the user, it can contain any set of units, depending on the desires of the user.  Suppose there is a user that has no particular need to group units together - can the idea of a UnitGroup be regarded as optional?   If not, how is a user to know what unit group (or groups) they must create?

ontologies/gistCore.ttl
```diff
@@ -3326,6 +3354,14 @@ gist:hasUniqueSuperCategory
 	skos:prefLabel "has unique supercategory"^^xsd:string ;
 	.
 
+gist:hasUnitGroup
+	a owl:ObjectProperty ;
+	rdfs:domain gist:Aspect ;
+	rdfs:range gist:UnitGroup ;
+	skos:definition "Relates an Aspect to a Unit Group whose members are Units of Measure. The Aspect can be measured using any of the members of the Unit Group."^^xsd:string ;
```
uscholdm, line 3636: What if an aspect can be measured by any number of unit groups? Is it useful, or necessary or good practice to link to all of them?

uscholdm, line 3636: A unit group by definition has units of measure as members, you could leave that out of the definition.

ontologies/gistCore.ttl
```diff
@@ -3326,6 +3354,14 @@ gist:hasUniqueSuperCategory
 	skos:prefLabel "has unique supercategory"^^xsd:string ;
 	.
 
+gist:hasUnitGroup
```
uscholdm, line 3632: I preferred the term `isMeasuredBy`, from a prior proposal.

uscholdm (commented):
A few suggestions.

philblackwood:
Working on release notes. A brief guide to review the new ontology is attached: 
[Review Guide for Units of Measure.docx](https://github.com/semanticarts/gist/files/14563977/Review.Guide.for.Units.of.Measure.docx)

ontologies/gistCore.ttl
```diff
@@ -2529,6 +2537,26 @@ gist:Transaction
...
+				owl:onProperty gist:hasMember ;
+				owl:someValuesFrom gist:UnitOfMeasure ;
+			]
+		) ;
+	] ;
+	owl:disjointWith gist:UnitOfMeasure ;
+	skos:definition "A collection of Units of Measure that can all be used to measure the same Aspects."^^xsd:string ;
+	skos:example "Bit, byte, kilobit, kilobyte, etc. can all be used to measure an amount of data."^^xsd:string ;
+	skos:prefLabel "Unit Group"^^xsd:string ;
+	skos:scopeNote "When adding a unit of measure to a Unit Group, make sure it has the same exponents as the other members of the Unit Group. For example, miles per gallon and gallons per mile could both be used to measure vehicle efficiency, but they need to be in different unit groups."^^xsd:string ;
```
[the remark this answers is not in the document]
philblackwood: The exponents are with the Unit Group (and the entire data set is not large), and the conversionFactor can be used to sort the Units of Measure within a group so that ones with the scale the user cares about are easy to identify.

ontologies/gistCore.ttl
philblackwood: The contents of the Unit Group should be all units that can be used to measure the same Aspect. The Unit Group serves to identify which units can be used to measure a given Aspect, so it should always be there.

ontologies/gistCore.ttl
philblackwood: True. The redundancy is intended to make it easier for someone new to gist to understand how the concepts tie together. Not sure it is needed, but we could put the redundant part in parentheses to make the meaning of the sentence clearer.

ontologies/gistCore.ttl
philblackwood: Originally I liked the "measuredBy" term but several people said they found it ambiguous, e.g. temperature is actually measured by a thermometer. 

An advantage of "hasUnitGroup" is that it implies the intended range, and is consistent with hasUnitOfMeasure and hasAspect.

ontologies/gistCore.ttl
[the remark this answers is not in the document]
uscholdm: I don't quite follow this. It would be a great help to see exactly how you would represent this exact example so I can assess it by looking at the triples.  That is the only way I can be sure to understand the proposal. Better still, for myself and future gist users, it would be helpful to have the new representation of what has been in gist for many years there as a set of triples ready to use.  I'm not sure at the moment how I would proceed using the new approach with a new client.

ontologies/gistCore.ttl
uscholdm: See above comment.

ontologies/gistCore.ttl
uscholdm: This is a matter of style/prefernce. Not sure what our general policy is. Perhaps @rjyounes can opine?

uscholdm: What about the hypothetical example of astronomical units vs. teeny ones. That would seem to be a good case where two unit groupls would make sense. What is your recommendation for handling that situation?

uscholdm: > "Relates an aspect to a group of units of measure. The aspect can be measuring using any of the units in the unit group."

I think you mean 'can be measured using', not 'measuring using'

uscholdm: > "Relates an aspect to a group of units of measure. The aspect can be measuring using any of the units in the unit group."

This allow for any grouping of units, e..g meter, mile per hour, milligram. But that is not intended. As best as I can tell, the class `UnitGroup` is a standin for a dimension with a standard unit.

Also the definition should link the `relates` to the meaning of the relationship, such as:
`Relates an aspect to a group of units of measure whereby the aspect can be measured using any of the units in the unit group.` or

`Relates an aspect to a group of units of measure all of which can be used to measure the aspect.`

ontologies/gistCore.ttl
```diff
@@ -1414,55 +1045,35 @@ gist:Magnitude
...
 		) ;
 	] ;
-	skos:definition "A quantity expressed with a unit of measure and a number."^^xsd:string ;
-	skos:example "3 millimeters; 70 miles per hour; 110 degrees F"^^xsd:string ;
+	skos:definition "The amount of a measurable characteristic (Aspect)."^^xsd:string ;
+	skos:example "A model of car could have a wheelbase of 113.2 inches. In this example, the Aspect is wheelbase, the Unit of Measure is inch, and the numericValue is 113.2."^^xsd:string ;
 	skos:prefLabel "Magnitude"^^xsd:string ;
-	skos:scopeNote
-		"Magnitudes of the same dimensional type (i.e., duration or electric current) can be compared with a greater-than or less-than operator, but can still differ in their relationToTheWorld type.  (I.e., you can compare actuals to estimates or references, so long as the dimension is the same.)"^^xsd:string ,
-		"The precision should be in the same type of unit as the magnitude."^^xsd:string
```
kchastain703, line 1426: Do we want to keep this scope note about precision in the new Magnitude definition? Or, rather, should the precision at least be in the same Unit Group as the Magnitude?

ontologies/gistCore.ttl
```diff
@@ -32,6 +33,55 @@ skos:scopeNote
 	a owl:AnnotationProperty ;
 	.
 
+gistd:_Aspect_area
+	a gist:Aspect ;
+	skos:definition "The aspect area."^^xsd:string ;
```
rjyounes, line 38: I'd like more substantive definitions. You could take them from the current gist class definitions or grab them off Google. E.g., 

"A measure of two-dimensional space." (gist)
"A measure of a region's size on a surface." (Wikipedia)
"The extent of measurement of a surface."
etc.

ontologies/gistCore.ttl
```diff
@@ -32,6 +33,55 @@ skos:scopeNote
...
+	skos:prefLabel "area"^^xsd:string ;
+	skos:scopeNote "This instance has been duplicated from reference data so it can be used in class equivalence statements."^^xsd:string ;
```
rjyounes, line 40:
```suggestion
	skos:editorialNote "This instance has been duplicated from reference data so it can be used in class equivalence statements."^^xsd:string ;
```

Here and below - these are all editorial notes.

ontologies/gistCore.ttl
```diff
@@ -146,48 +165,9 @@ gist:Artifact
 
 gist:Aspect
 	a owl:Class ;
-	rdfs:subClassOf gist:Category ;
-	skos:definition "A very general term for the characteristic of something that is being measured.  E.g., property (height) or a process (cycle time) or a behavior (loyalty)."^^xsd:string ;
+	skos:definition "A measurable characteristic such as cost, cycle time, or defect rate."^^xsd:string ;
```
rjyounes, line 168:
```suggestion
	skos:definition "A measurable characteristic such as length, weight, cost, cycle time, or defect rate."^^xsd:string ;
```

Proposing a couple of other common examples.

ontologies/gistCore.ttl
```diff
@@ -1414,55 +1045,35 @@ gist:Magnitude
 			[
 				a owl:Restriction ;
 				owl:onProperty gist:numericValue ;
-				owl:someValuesFrom xsd:double ;
+				owl:someValuesFrom [
```
rjyounes, line 1048: It would be more concise not to repeat the entire range of `gist:numericValue`. I'd be in favor of using `owl:minCardinality 1` here.

rjyounes: Alternatively, you could go with `owl:someValuesFrom rdfs:Literal`.

ontologies/gistCore.ttl
rjyounes: @philblackwood Do we still allow for a Magnitude to have a precision? Should we keep it as a restriction (which doesn't have to be stated due to OWA)?

ontologies/gistCore.ttl
```diff
@@ -1414,55 +1045,35 @@ gist:Magnitude
...
+						owl:rational
+						owl:real
+					) ;
+				] ;
 			]
 		) ;
 	] ;
-	skos:definition "A quantity expressed with a unit of measure and a number."^^xsd:string ;
-	skos:example "3 millimeters; 70 miles per hour; 110 degrees F"^^xsd:string ;
+	skos:definition "The amount of a measurable characteristic (Aspect)."^^xsd:string ;
```
rjyounes, line 1074:
```suggestion
	skos:definition "The amount of a measurable characteristic (aspect)."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -2529,231 +1975,30 @@ gist:Transaction
...
-				owl:onProperty gist:hasUnitOfMeasure ;
-				owl:someValuesFrom gist:VolumeUnit ;
+				owl:onProperty gist:hasMember ;
+				owl:someValuesFrom gist:UnitOfMeasure ;
 			]
 		) ;
 	] ;
-	skos:definition "Three-dimensional space, or equivalent fluid measurement."^^xsd:string ;
-	skos:prefLabel "Volume"^^xsd:string ;
+	owl:disjointWith gist:UnitOfMeasure ;
```
rjyounes, line 1991: Shouldn't be needed, since a collection is not a unit of measure (not sure if that gets stated explicitly anywhere in gist 12, but it seems out of place here. Maybe just add the disjoint on `gist:UnitOfMeasure`, since it already has a bunch of them.

ontologies/gistCore.ttl
```diff
@@ -2529,231 +1975,30 @@ gist:Transaction
...
+	skos:definition "A collection of units of measure that can all be used to measure the same aspects."^^xsd:string ;
+	skos:example "The units of measure bit, byte, kilobit, kilobyte, etc. are all in the same unit group because they can all be used to measure an amount of data."^^xsd:string ;
+	skos:prefLabel "Unit Group"^^xsd:string ;
+	skos:scopeNote "Typically there is one UnitGroup per aspect. An example of an aspect with two unit groups is vehicle efficiency, which can be measured by miles per gallon (distance per volume) or by liters per 100 kilometers (volume per distance). These two units of measure need to be in different unit groups because they have different values of exponents. When adding a unit of measure to a unit group, make sure it has the same exponents as the other members of the unit group."^^xsd:string ;
```
rjyounes, line 1995:
```suggestion
	skos:scopeNote "Typically there is one unit group per aspect. An example of an aspect in two unit groups is vehicle efficiency, which can be measured by miles per gallon (distance per volume) or by liters per 100 kilometers (volume per distance). These two units of measure need to be in different unit groups because they have different values of exponents. When adding a unit of measure to a unit group, make sure it has the same exponents as the other members of the unit group."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -2967,23 +2212,48 @@ gist:contributesTo
...
+	rdfs:domain gist:UnitOfMeasure ;
+	rdfs:range [
+		a rdfs:Datatype ;
+		owl:unionOf (
+			xsd:decimal
+			xsd:double
+		) ;
+	] ;
+	rdfs:seeAlso gist:conversionOffset ;
+	skos:definition "A value that relates a unit of measure to units of the International System of Units. For example, in the equation 1 inch = 0.0254 meters, the value 0.0254 is the conversion factor of inch."^^xsd:string ;
```
rjyounes, line 2228:
```suggestion
	skos:definition "A value that relates a unit of measure to units of the International System of Units. For example, in the equation 1 inch = 0.0254 meters, the value 0.0254 is the factor used to convert meters to inches."^^xsd:string ;
```

rjyounes: Should we add a scopeNote to the effect that standard units have a conversion factor of 1?

ontologies/gistCore.ttl
```diff
@@ -3043,6 +2313,211 @@ gist:endDateTime
...
+
+		Every member of a unit group containing candlepower will have a similar decomposition."""^^xsd:string ;
+	skos:prefLabel "exponent of candela"^^xsd:string ;
+	.
+
+gist:exponentOfKelvin
+	a owl:DatatypeProperty ;
+	rdfs:domain gist:UnitGroup ;
+	rdfs:range xsd:decimal ;
+	skos:definition "The exponent of kelvin in a product of powers of base units."^^xsd:string ;
```
rjyounes, line 2366: Is Kelvin capitalized or not? Can you make it consistent either way?

ontologies/gistCore.ttl
```diff
@@ -3326,6 +2809,14 @@ gist:hasUniqueSuperCategory
 	skos:prefLabel "has unique supercategory"^^xsd:string ;
 	.
 
+gist:hasUnitGroup
+	a owl:ObjectProperty ;
+	rdfs:domain gist:Aspect ;
+	rdfs:range gist:UnitGroup ;
+	skos:definition "Relates an aspect to a unit group. The aspect can be measured using any of the members of the unit group."^^xsd:string ;
+	skos:prefLabel "has unit group"^^xsd:string ;
```
rjyounes, line 2817: Add example.

rjyounes (changes requested):
This is amazing work! I've made some comments related to improving annotations.

ontologies/gistCore.ttl
```diff
@@ -3043,6 +2313,211 @@ gist:endDateTime
 		;
 	.
 
+gist:exponentOfAmpere
```
rjyounes, line 2316: For my own understanding - why do we have exactly these exponents?

ontologies/gistCore.ttl
```diff
@@ -3043,6 +2313,211 @@ gist:endDateTime
...
+	rdfs:domain gist:UnitGroup ;
+	rdfs:range xsd:decimal ;
+	skos:definition "The exponent of ampere in a product of powers of base units."^^xsd:string ;
+	skos:example """In the equation 1 milliampere = 0.001 x ampere
+
+		the conversionFactor for milliampere is 0.001
+		the exponent of ampere is 1
+		all other exponents are zero
+
+		Every member of a unit group containing milliampere will have a similar decomposition, although with a different conversionFactor."""^^xsd:string ;
```
rjyounes, line 2327: Here and below:

```suggestion
		Every member of a unit group containing milliampere will have a similar decomposition, although with a different conversion factor."""^^xsd:string ;
```

ontologies/gistCore.ttl
rjyounes: If a unit group has two members with the same conversion factor, do we say these are `owl:sameAs`? Otherwise you can't tell which is the standard unit.

ontologies/gistCore.ttl
```diff
@@ -780,112 +559,12 @@ gist:DegreeOfCommitment
...
-				owl:hasValue gist:_ampere ;
-			]
-		) ;
-	] ;
-	skos:definition "Unit of electrical current, which is charge per unit time.  The SI unit is the ampere.  (Note that electrical current is a composed unit.)"^^xsd:string ;
-	skos:prefLabel "Electrical Current Unit"^^xsd:string ;
+	rdfs:subClassOf gist:Category ;
+	skos:definition "An area of study or practice, such as accounting."^^xsd:string ;
+	skos:example "Finance, Accounting, Project Management, Acoustics, Ballistics, etc."^^xsd:string ;
+	skos:prefLabel "Discipline"^^xsd:string ;
```
rjyounes, line 567: How do we formally relate units to disciplines?

rjyounes (changes requested):
Additional comments and questions.

ontologies/gistCore.ttl
rjyounes: We should also have a restriction relating an aspect to a unit group, shouldn't we?

uscholdm: > We should also have a restriction relating an aspect to a unit group, shouldn't we?

A top level aspect always relates to a unit group, e.g. distance, but the lower level ones e.g. width, outer diameter point to a broader aspect, not a unit group. @philblackwood Is this right?

rjyounes: @uscholdm Yes, you're right.

rjyounes, line 1991: Shouldn't be needed, since a collection is not a unit of measure. Not sure if that gets stated explicitly anywhere in gist 12, but it seems out of place here. Maybe just add the disjoint on gist:UnitOfMeasure, since it already has a bunch of them.

uscholdm (commented):
On the restrictions for say, GeoRegion, the semantics is incomplete.  GeoRegion hasMagnitude a magnitude that has the aspect area. But the area aspect is not tied to any unit group or unit. This would be solved by having a separate module of unit groups, units and aspects that replace the ones in gist 12, rather than removing them.

Rather than duplicating from reference data, have the aspect instances for mass, volume etc be in a separate module that is imported by gist.y has gist:Goal gone away? 

The restriction on magnitude about precision gives infinite regress of magnitudes, there is an issue on this. #980 

The definition of Magnitude could be a bit more elaborate. 
- gist12.1.0 definition: `A quantity expressed with a unit of measure and a number.`
- current definition: `The amount of a measurable characteristic (Aspect).`
I suggest more like: 
- `The amount of a measurable characteristic intended to be used to describe something.`

Magnitude example annotation:
- Current example: `A model of car could have a wheelbase of 113.2 inches. In this example, the Aspect is wheelbase, the Unit of Measure is inch, and the numericValue is 113.2`.
- I suggest: `The magnitude of a wheelbase being equal to 116.7 inches describes the 2024 Tesla Model X. In this example, the Aspect is wheelbase, the Unit of Measure is inch, and the numericValue is 116.7.`

With the exponent properties  having a domain of `UnitGroup`, if you say the logically sensible thing, that the unit for milli-second has exponent of second equal to 1, then second will be inferred into the class `UnitGroup`. I suggest using `domainIncludes` rather than range.  This could be useful in a scenario where materialization is used for compute efficiency.

Problems with the annotations for the exponent properties:
 the example for exponentOfMeter says:
`Every member of a Unit Group containing microgram per milliliter will have a similar decomposition, although with a different conversionFactor.`

- First off, why are we talking about micograms per millilitere for the meter exponent? Probably a cut/paste and forget to edit error.
- Next, units are not given exponents, only unit groups are so it is misleading to speak of decomposition of a unit.
- Finally, term 'decomposition' is used but not explained.
I suggest this alternative:
- Have a superpropety of all the exponent properties and explain what that is all about. 
  - May or may not need/want to use the word decomposition.
  - Have a comment something like: `Every member of the unit group containing meter (e..g inch, kilometer) inherits the exponents from the unit group.`

ontologies/gistCore.ttl
uscholdm: It is a true statement. If it is inferable from other disjoint and subclass statements, then it is redundant.  Can you point that out, I'm not seeing it. Otherwise, it makes sense to keep it.

ontologies/gistCore.ttl
```diff
@@ -146,48 +165,9 @@ gist:Artifact
...
-			gist:_kelvin
-			gist:_kilogram
-			gist:_meter
-			gist:_mole
-			gist:_second
-		) ;
-	] ;
-	skos:definition "A primitive unit that cannot be decomposed into other units. It can be converted from one measurement system to another.  The base units in gist are the seven primitive units from the System Internationale (SI): (meter, second, kilogram, ampere, kelvin, mole, candela), plus three convenience ones: each. bit and usDollar."^^xsd:string ;
-	skos:prefLabel "Base Unit"^^xsd:string ;
+	skos:scopeNote "Every aspect should be related to a broader aspect or to a unit group. For example, angle of incidence should be related to the broader concept of angle, which in turn is related to a unit group."^^xsd:string ;
```
kchastain703, line 170: The scopeNote for UnitGroup mentions that an aspect can potentially have more than one unit group (the vehicle efficiency example). Should we change this note to "... to one or more unit groups" to align with that?

ontologies/gistCore.ttl
```diff
@@ -3326,11 +2891,19 @@ gist:hasUniqueSuperCategory
...
+	skos:definition "Relates an aspect to a unit group. The aspect can be measured using any of the members of the unit group."^^xsd:string ;
+	skos:prefLabel "has unit group"^^xsd:string ;
+	.
+
 gist:hasUnitOfMeasure
 	a owl:ObjectProperty ;
 	rdfs:domain gist:Magnitude ;
 	rdfs:range gist:UnitOfMeasure ;
-	skos:definition "Which unit of measure you are using. All measures are expressed in some unit of measure, even if we don't know what it is initially."^^xsd:string ;
+	skos:definition "Which unit of measure you are using. All measures are expressed in some unit of measure."^^xsd:string ;
```
kchastain703, line 2906: Since we're juggling the word "measure" here a lot, do you think it might be helpful to say instead, "All measurements are expressed in..."?

ontologies/gistCore.ttl
```diff
@@ -3288,12 +2853,12 @@ gist:hasRecipient
...
 
-gist:hasStandardUnit
+gist:hasSubtrahend
 	a owl:ObjectProperty ;
-	rdfs:domain gist:UnitOfMeasure ;
-	rdfs:range gist:CoherentUnit ;
-	skos:definition "For a complex unit refers to a unit that has all the component parts in SI"^^xsd:string ;
-	skos:prefLabel "has standard unit"^^xsd:string ;
+	skos:definition "Relates an aspect to another aspect that is a subtracted component of it."^^xsd:string ;
+	skos:example "In the equation profit = revenue - expenses, revenue is an addend and expenses is a subtrahend."^^xsd:string ;
```
kchastain703, line 2859: For this example and the ones for the other math term properties, do you think it would make it more readable to set off the equation in quotation marks, like a sentence?
as in,
"In the equation 'profit = revenue - expenses', 'revenue' is an addend and 'expenses' is a subtrahend'." ?

ontologies/gistCore.ttl
rjyounes: But the syntax needs to be cleaned up:

"Relates a magnitude to the unit of measure it uses to express a quantity." - or something.

ontologies/gistCore.ttl
rjyounes: How about:

"""In the equation

profit = revenue - expenses

revenue is an addend and expenses is a subtrahend."""

ontologies/gistCore.ttl
```diff
@@ -10,7 +10,7 @@
 
 <https://w3id.org/semanticarts/ontology/gistCore>
 	a owl:Ontology ;
-	owl:versionIRI <https://w3id.org/semanticarts/ontology/gistCoreX.x.x> ;
+	owl:versionIRI <https://w3id.org/semanticarts/ontology/gistCore13.0.0> ;
```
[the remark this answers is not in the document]
philblackwood: Fixed

ontologies/gistCore.ttl
philblackwood: Updated to say:

skos:definition "Relates an aspect to a unit group. The aspect can be measured using any of the members of the unit group."^^xsd:string ;

ontologies/gistCore.ttl
philblackwood: Added a note about consistency. And yes, I think we need to keep precision.

note added: Note that the unit of measure of the precision has to be compatible with the unit of measure of the original magnitude (e.g. something measured in meters could have a precision in terms of millimeters or any other unit that measures distance).

ontologies/gistCore.ttl
philblackwood: Fixed

ontologies/gistCore.ttl
philblackwood: Done

ontologies/gistCore.ttl
philblackwood: Nice. Done.

ontologies/gistCore.ttl
uscholdm: > And yes, I think we need to keep precision.

This results in an infinite regress of precisions (see #980 )

ontologies/gistCore.ttl
```diff
@@ -1045,29 +1045,7 @@ gist:Magnitude
...
-						xsd:short
-						xsd:unsignedByte
-						xsd:unsignedInt
-						xsd:unsignedLong
-						xsd:unsignedShort
-						owl:rational
-						owl:real
-					) ;
-				] ;
+				owl:minCardinality "1"^^xsd:nonNegativeInteger ;
```
uscholdm, line 1048: I like limiting the type to be just nonNegativeInteger, but independently, the someValuesFrom has chagned to min 1 cardinality. This goes against the style guide, which says to use someValuesFrom

ontologies/gistCore.ttl
philblackwood: Yes. In fact, a precision does have a precision, and if anybody wanted to record these they could. Are there any practical negative consequences?

ontologies/gistCore.ttl
philblackwood: fixed

ontologies/gistCore.ttl
philblackwood: I don't think there is a need to suggest there is a concept of "standard unit".

philblackwood: The original definition gives an example of a conversion factor. The proposed change doesn't say that 0.254 is a conversion factor. The scope note explains how to use the conversion factors, so nothing has been omitted.

ontologies/gistCore.ttl
```diff
@@ -3043,6 +2313,211 @@ gist:endDateTime
...
+	skos:definition "The exponent of US Dollar in a product of powers of base units."^^xsd:string ;
+	skos:example """In the equation 1 million dollars per week = 1.65344 x dollars per second
+
+		the conversion factor for million dollars per week is 1.65344
+		the exponent of US Dollar is 1
+		the exponent of second is -1
+		all other exponents are zero
+
+		Every member of a unit group containing million dollars per week will have a similar decomposition, although with a different conversionFactor."""^^xsd:string ;
+	skos:prefLabel "exponent of us dollar"^^xsd:string ;
```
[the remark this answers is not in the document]
philblackwood: done

ontologies/gistCore.ttl
philblackwood: Yes, Kelvin should be capitalized. Fixed.

ontologies/gistCore.ttl
```diff
@@ -3082,19 +2565,19 @@ gist:hasAddress
 gist:hasAltitude
 	a owl:ObjectProperty ;
 	rdfs:domain gist:GeoPoint ;
-	rdfs:range gist:Extent ;
+	rdfs:range gist:Magnitude ;
```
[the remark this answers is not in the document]
philblackwood: I think it's OK as is.   hasAltitude makes sense. We could make it a subProperty of hasMagnitude. Also, the Domain should be expanded, so that something like an airplane can have an altitude (future change).

ontologies/gistCore.ttl
```diff
@@ -3105,6 +2588,15 @@ gist:hasBiologicalParent
...
 	.
 
+gist:hasBroader
+	a
+		owl:ObjectProperty ,
+		owl:TransitiveProperty
+		;
+	skos:definition "Relates a thing to another thing with a broader meaning."^^xsd:string ;
+	skos:prefLabel "has broader"^^xsd:string ;
+	.
```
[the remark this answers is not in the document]
philblackwood: done

ontologies/gistCore.ttl
```diff
@@ -3147,6 +2632,14 @@ gist:hasDirectSuperCategory
 	skos:scopeNote "Unlike its superproperty gist:hasSuperCategory, this property is not transitive. It is essentially the same as the non-transitive skos:broader, using gist:Category rather than skos:Concept."^^xsd:string ;
 	.
 
+gist:hasDivisor
+	a owl:ObjectProperty ;
+	skos:definition "Relates an aspect to another aspect that is a divisor. May also be used with units of measure."^^xsd:string ;
+	skos:example "Miles per hour has miles as a multiplier and hour as a divisor."^^xsd:string ;
```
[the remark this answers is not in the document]
philblackwood: done. speed has distance as multiplier and duration as divisor.

ontologies/gistCore.ttl
```diff
@@ -3183,7 +2676,8 @@ gist:hasIncumbent
 gist:hasMagnitude
 	a owl:ObjectProperty ;
 	rdfs:range gist:Magnitude ;
-	skos:definition "To have a comparable numeric value. Each magnitude has a unit."^^xsd:string ;
+	skos:definition "Relates a Thing to a magnitude."^^xsd:string ;
```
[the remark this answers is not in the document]
philblackwood: A magnitude can be a stand-alone amount, used by multiple things. There might also be cases when design considerations lead to making each magnitude specific to one individual. The ontology should be the same either way.

ontologies/gistCore.ttl
```diff
@@ -3197,20 +2691,12 @@ gist:hasMember
...
-
 gist:hasMultiplier
 	a owl:ObjectProperty ;
-	skos:definition "Relates a product to the factor that is multiplying another factor (i.e. the multiplicand)."^^xsd:string ;
-	skos:example "Newton is the multiplier for the unit: newton-meter"^^xsd:string ;
+	skos:definition "Relates an aspect to another aspect that is a factor. May also be used with units of measure."^^xsd:string ;
+	skos:example "Miles per hour has miles as a multiplier and hour as a divisor."^^xsd:string ;
 	skos:prefLabel "has multiplier"^^xsd:string ;
-	skos:scopeNote "Because multiplication is commutative, the difference between multiplier and multiplicand will not matter for computation, but it may matter if you care about the order for presentation, e.g. newton-meter vs. meter-newton."^^xsd:string ;
+	skos:scopeNote "Provides a supplemental method of decomposing an aspect or UnitOfMeasure into component factors. Enables dimensional analysis such as miles per hour x hours = miles."^^xsd:string ;
```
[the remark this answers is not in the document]
philblackwood: done

ontologies/gistCore.ttl
philblackwood: done

ontologies/gistCore.ttl
philblackwood: The exponents allow math such as creating a new unit of measure from existing units. Without the exponents, parsing is required. Example:

1 watt-hour per mile

= watt x hour x mile-1

= (1 x kilogram x meter2 x second-3) x (3600 x second) x (1609.344 meter)-1

= (1 x 3600 x 1609.344-1)  x kilogram x (meter2 x meter-1) x (second-3 x second)

= 2.237 x kilogram x meter x second-2

ontologies/gistCore.ttl
philblackwood: done

ontologies/gistCore.ttl
philblackwood: Discipline is a subclass of Category, so we can use gist:isCategorizedBy to relate things to disciplines. For ease of maintaining the data, it is probably simpler to relate only aspects to disciplines and then tailor the unit groups to the aspects instead of tailoring them to disciplines as I was originally thinking. For example, diastolic blood pressure would have its own unit group with just a few entries, even though theoretically we could say it has a broader aspect of pressure, and any unit of pressure could be used for blood pressure. Unfortunately, there are over 100 units for pressure and most of them are never used for blood pressure.

ontologies/gistCore.ttl
philblackwood: Volume is no longer a class. I don't see an issue.

ontologies/gistCore.ttl
philblackwood: I think 99.9% of the time an aspect will be related to one unit group. But never say never. So the informal annotation describes the main case, while the ontology allows for the outlier.

ontologies/gistCore.ttl
philblackwood: Added single quotes around equations.

ontologies/gistCore.ttl
philblackwood: As far as I can tell, someValuesFrom would require the complete list of types listed as the range of numericValue, or we can use minCardinality. Is there another option? I like the minCardinality because it is more concise, and also because it eliminates redundant listing of all the types.

ontologies/gistCore.ttl
philblackwood: done

ontologies/gistCore.ttl
uscholdm: > I don't think there is a need to suggest there is a concept of "standard unit".

There is such a concept in physics, it is called 'coherent unit', it is the basis for conversions.  I am in favor of adding a scope note about this, given it is a fundamental concept in physics that the decision to not explicitly include it in the model.

ontologies/gistCore.ttl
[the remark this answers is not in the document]
uscholdm: > I think it's OK as is. hasAltitude makes sense. We could make it a subProperty of hasMagnitude.

@rjyounes makes a good point.  Doing this without an aspect undermines the goal of doing magnitudes in a consistent way.  It would serve as an example of how to use aspect in the new way.

> Also, the Domain should be expanded, so that something like an airplane can have an altitude (future change). [as opposed to just `GeoPoint`s]

Good point, so to speak, which will be moot if we choose to use an aspect instead.

ontologies/gistCore.ttl
```diff
@@ -2652,6 +2652,7 @@ gist:hasBroader
 		owl:TransitiveProperty
 		;
 	skos:definition "Relates a thing to another thing with a broader meaning."^^xsd:string ;
+	skos:example "Height is a context-specific notion of distance, so height has broader distance."^^xsd:string ;
```
uscholdm, line 2655: This is not natural English.  How about: 
- `the aspect, distance is broader than the aspect height`

I cannot think of a good way to say this mentioning the height first and still using the term 'broader'.

ontologies/gistCore.ttl
uscholdm: Perhaps add a scopeNote somewhere to explain this?

ontologies/gistCore.ttl
philblackwood: reworded as requested

ontologies/gistCore.ttl
uscholdm: > Discipline is a subclass of Category, so **we can** use gist:isCategorizedBy to relate things to disciplines.

**we can** or **we do in fact** use gist:isCategorizedBy to relate things to disciplines.

ontologies/gistCore.ttl
uscholdm: In Manchester syntax, the two options are as follows (cut and pasted from Protege)
-  (gist:numericValue some xsd:nonNegativeInteger)
-  (gist:numericValue min 1 xsd:nonNegativeInteger)

Conciseness is the same.

uscholdm: Oops, my mistake.   Direct from Protege, the proper comparison is:
-  (gist:numericValue min 1 rdfs:Literal)
-  (gist:numericValue some rdfs:Literal)

The Turtle will look slightly different.

The are semantically identical and there is no significant difference in conciseness.    
The first follows our style guidelines.

ontologies/gistCore.ttl
philblackwood: Can. Not sure we need to. The reference data is not part of this pull request.

ontologies/gistCore.ttl
philblackwood: If the aspect indicates a particular range of values is needed, create a unit group for the aspect and put the appropriate units of measure in it. 

Example: pressure could have 100 different values, but diastolic blood pressure could have just a few.

ontologies/gistCore.ttl
uscholdm: Ok - im guessing that the evolving reference data does it this way at the moment, it's good to do so.

ontologies/gistCore.ttl
kchastain703: What about something like, 
"Height is specifically a distance between a lower point and a higher point..."

ontologies/tmp.domain
Jamie-SA, line 1: Maybe I missed it somewhere, but what is this file for?

uscholdm: @philblackwood Looks like a mistake,

docs/releaseNote1033.md
```diff
@@ -0,0 +1,116 @@
+## Release 13.0.0
+
+This is a major release which includes several changes which break compatibility with previous versions of gist, most notably an entirely revised model of units and magnitudes. See the [migration guide](./MajorVersionMigration.html) for documentation on updating existing gist-based ontologies and instance data. Migration scripts are included to facilitate the upgrade process.
```
rjyounes, line 3:
```suggestion
This is a major release that includes several changes which break compatibility with previous versions of gist, most notably an entirely revised model of units and magnitudes. See the [migration guide](./MajorVersionMigration.html) for documentation on updating existing gist-based ontologies and instance data. Migration scripts are included to facilitate the upgrade process.
```

docs/releaseNote1033.md
```diff
@@ -0,0 +1,116 @@
...
+  - `gist:LuminousIntensity`
+  - `gist:Mass`
+  - `gist:MolarQuantity`
+  - `gist:Monetary`
+  - `gist:MonetaryPerDuration`
+  - `gist:ProductMagnitude`
+  - `gist:RatioMagnitude`
+  - `gist:Temperature`
+  - `gist:Volume`
+- Deleted all subclass of `gist:UnitOfMeasure`:
```
rjyounes, line 26:
```suggestion
- Deleted all subclasses of `gist:UnitOfMeasure`:
```

ontologies/gistCore.ttl
```diff
@@ -1414,55 +1048,14 @@ gist:Magnitude
 			[
 				a owl:Restriction ;
 				owl:onProperty gist:numericValue ;
-				owl:someValuesFrom xsd:double ;
+				owl:minCardinality "1"^^xsd:nonNegativeInteger ;
```
rjyounes, line 1051:
```suggestion
				owl:someValuesFrom rdfs:Literal ;
```

ontologies/gistCore.ttl
```diff
@@ -1414,55 +1048,14 @@ gist:Magnitude
...
-			[
-				a owl:Restriction ;
-				owl:onProperty gist:hasBaseUnit ;
-				owl:hasValue gist:_kilogram ;
-			]
-		) ;
-	] ;
-	skos:definition "A unit representing the amount of matter in a particle or object.  The SI unit of mass is the kilogram."^^xsd:string ;
-	skos:prefLabel "Mass Unit"^^xsd:string ;
+	skos:scopeNote "A precision can be assigned to a magnitude to describe the probabilistic accuracy of the numeric value."^^xsd:string ;
```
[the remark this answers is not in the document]
philblackwood: The meaning of precision in gist is not the number of decimal places that are significant. It is intended to represent the plus/minus accuracy of a measurement (e.g. accurate to within 1 millimeter 95% of the time).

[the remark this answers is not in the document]
philblackwood: It looks like the concept we want to capture is accuracy, not precision.

I suggest we change the name of the property to gist:hasAccuracy.

From the link Rebecca provided (which agrees with other sources):

"Accuracy is degree to which the measured value agrees with true value. Example-Our objective is to make rod of 25mm And we are able to make it of 25 mm then it is accurate.

Precision is the repeatability of the measuring process. Example-Our objective is to make 10 rods of 25mm and we make all rods of 24mm then we are precise as we make all rods of same size,but it is not accurate as true value is 25 mm."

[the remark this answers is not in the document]
kchastain703: That makes sense to me: re: changing to gist:hasAccuracy.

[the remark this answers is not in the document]
uscholdm: > Precision is the repeatability of the measuring process. Example-Our objective is to make 10 rods of 25mm and we make all rods of 24mm then we are precise as we make all rods of same size,but it is not accurate as true value is 25 mm.

this surprises me. I always thought precision was about number of digits. E.g. 3.14 is less precise than 3.14159.  I would ask Dave  @mkumba what he was thinking.   Note  spurious digits may arise which should be ignored.

ontologies/gistCore.ttl
```diff
@@ -2559,248 +1986,50 @@ gist:TimeInterval
...
-	gist:hasBaseUnit gist:_second ;
-	gist:symbol "min"^^xsd:string ;
-	gist:unitSymbol "min"^^xsd:string ;
+	skos:definition "A span of time with a known start time, end time, and duration. As long as two of the three are known, the third can be inferred."^^xsd:string ;
+	skos:example "7pm to 9pm on Jan 1, 2001; fiscal year 2023; the week starting at midnight of January 12, 2023 and lasting exactly 168 hours."^^xsd:string ;
+	skos:prefLabel "Time Interval"^^xsd:string ;
+	skos:scopeNote
+		"An ongoing state of affairs with an unknown end time in the future cannot be a time interval; e.g. the lifespan of a living person cannot be a time interval, as the end time is unknown."^^xsd:string ,
+		"This is distinct from a gist:Duration, which describes how long a time interval lasts (e.g., one hour; 3 days; 22 minutes)."^^xsd:string
+		;
```
uscholdm, line 1995: gist:Duration is no more.

ontologies/gistCore.ttl
```diff
@@ -1425,55 +1058,14 @@ gist:Magnitude
 			[
 				a owl:Restriction ;
 				owl:onProperty gist:numericValue ;
-				owl:someValuesFrom xsd:double ;
+				owl:someValuesFrom rdfs:Literal ;
```
uscholdm, line 1061: Would it make sense to use the following instead of rdfs:LIteral?
```
[
		a rdfs:Datatype ;
		owl:unionOf (
			xsd:decimal
			xsd:double
		) ;
	]
```

ontologies/gistCore.ttl
```diff
@@ -3012,23 +2241,48 @@ gist:contributesTo
...
 gist:conversionOffset
-	a owl:DatatypeProperty ;
+	a
+		owl:DatatypeProperty ,
+		owl:FunctionalProperty
+		;
 	rdfs:domain gist:UnitOfMeasure ;
-	rdfs:range xsd:double ;
-	skos:definition "Add this number to get to the zero point.  On the Celsius scale, the conversionOffset is -273.15 degrees C. On the Fahrenheit scale it is -459.67 degrees.  Is equal to 0 when the unit has the same zero point as the base unit. e.g. inch, meter."^^xsd:string ;
+	rdfs:range xsd:decimal ;
```
uscholdm, line 2277: Why use a different range than for conversionFactor and numericValue?

ontologies/gistCore.ttl
```diff
@@ -3275,11 +2824,17 @@ gist:hasPhysicalLocation
...
 
 gist:hasPrecision
-	a owl:ObjectProperty ;
+	a
+		owl:ObjectProperty ,
+		owl:FunctionalProperty
+		;
 	rdfs:range gist:Magnitude ;
-	skos:definition "Links a Magnitude to the degree of accuracy of the numeric value.   This allows for fuzzy numbers.  All magnitudes have a precision.  Usually we don't record them.  When we do this, it will be a value whose extent covers 2 standard deviations around the stated magnitude"^^xsd:string ;
+	skos:definition "Relates a magnitude to the accuracy of the numeric value."^^xsd:string ;
```
uscholdm, line 2832: per above comments,  I think Dave originally had in mind precision in the sense of finer granularity (e.g. more digits) rather than accuracy.

uscholdm (changes requested):
Nearly there.

docs/releaseNote1033.md
```diff
@@ -0,0 +1,115 @@
...
+  - `gist:UnitOfMeasure`
+  - `gist:hasMagnitude`
+  - `gist:hasMultiplier`
+  - `gist:hasPrecision`
+  - `gist:conversionFactor`
+  - `gist:conversionOffset`
+- Added scripts for the migration of existing client ontologies and instance data. [TODO - add pointer]
+- Added complete documentation on understanding and implementing the new model. [TODO - add pointer]
+
+Import URL: <https://w3id.org/semanticarts/ontology/gistCore13.0.0>.
```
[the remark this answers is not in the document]
kchastain703: I believe these are the scripts that @philblackwood wrote in order to identify and amend UoM triples in gist users' stores using the new patterns.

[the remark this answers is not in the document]
uscholdm: > I believe these are the scripts that @philblackwood wrote in order to identify and amend UoM triples in gist users' stores using the new patterns.

Yes, I agree. There is a separate PR for the data conversion scripts,it would be appropriate to have these comments in the other PR

ontologies/gistCore.ttl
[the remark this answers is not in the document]
rjyounes: Phil: hasAccuracy = how close is the measurement to the actual measurement (probability distribution of error). This is needed for magnitudes. 

Decision: Change `hasPrecision` to `hasAccuracy`.

rjyounes (approved):

ontologies/gistCore.ttl
philblackwood: In gist12, the conversion factor is a double. Most can be decimals (simpler to read), but some have to be double because they have so many decimal places.

So I added decimal as an option. This is generally curated data, and we can say decimal where possible, double where necessary. Keeps things simpler.

ontologies/gistCore.ttl
philblackwood: It says "degree of _accuracy_" in the original definition.

ontologies/gistCore.ttl
philblackwood: changed gist:Duration to "duration"

ontologies/gistCore.ttl
```diff
@@ -1019,7 +687,17 @@ gist:GeoPoint
...
+					a owl:Class ;
+					owl:intersectionOf (
+						gist:Magnitude
+						[
+							a owl:Restriction ;
+							owl:onProperty gist:hasAspect ;
+							owl:hasValue gistd:_Aspect_distance ;
+						]
+					) ;
+				] ;
```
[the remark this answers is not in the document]
philblackwood: Currently follows the same usage as gist12. 

Would be better to use gistd:_Aspect_altitude in the definition. Changed.

docs/releaseNote1033.md
```diff
@@ -0,0 +1,115 @@
...
+- Modified the following classes and predicates:
+  - `gist:Aspect`
+  - `gist:Magnitude`
+  - `gist:UnitOfMeasure`
+  - `gist:hasMagnitude`
+  - `gist:hasMultiplier`
+  - `gist:hasPrecision`
+  - `gist:conversionFactor`
+  - `gist:conversionOffset`
+- Added scripts for the migration of existing client ontologies and instance data. [TODO - add pointer]
```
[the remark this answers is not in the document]
philblackwood: pointer to what?

docs/release_notes/issue1033-units-and-magnitudes.md
```diff
@@ -0,0 +1,115 @@
...
+  - `gist:Offer`
+  - `gist:PhysicalIdentifiableItem`
+  - `gist:PhysicalSubstance`
+- Modified the following classes and predicates:
+  - `gist:Aspect`
+  - `gist:Magnitude`
+  - `gist:UnitOfMeasure`
+  - `gist:hasMagnitude`
+  - `gist:hasMultiplier`
+  - `gist:hasPrecision`
```
kchastain703, line 109: We changed "hasPrecision" to "hasAccuracy", right?
(I'm having a bit of trouble finding the previous discussion on it, but I think we went with "hasAccuracy".)

philblackwood: Yes, I will update the release notes (and the data conversion).

rjyounes (changes requested):
@philblackwood I think the `hasAltitude` property should be removed, don't you?

docs/release_notes/issue1033-units-and-magnitudes.md
```diff
@@ -50,6 +50,9 @@ This is a major release that includes several changes which break compatibility
   - `gist:hasPrecision`
   - `gist:hasStandardUnit`
   - `gist:isAspectOf`
+  - `gist:unitSymbol`
+  - `gist:unitSymbolHtml`
+  - `gist:unitSymbolUnicode`
```
uscholdm, line 55: @rjyounes I think these were removed as part of another PR/Issue,  not sure how best to rationalize the release notes, maybe it does not matter.

uscholdm (dismissed):
@philblackwood - Im not sure which changes have been made, so this may be ready to approve. Apparently I had a few comments stacked up waiting for me to submit review.

kchastain703 (approved):
This looks good to me.

docs/release_notes/issue1033-units-and-magnitudes.md
rjyounes: I'll look into it when I compile the release notes.

ontologies/gistCore.ttl
rjyounes: No. Have a look at the range of `numericValue`.

ontologies/gistCore.ttl
```diff
@@ -485,9 +374,15 @@ gist:ContingentEvent
...
+					a owl:Class ;
+					owl:intersectionOf (
+						gist:Magnitude
+						[
+							a owl:Restriction ;
+							owl:onProperty gist:hasAspect ;
+							owl:hasValue gistd:_Aspect_probability ;
+						]
+					) ;
 				] ;
```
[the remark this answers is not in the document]
rjyounes: @uscholdm I guess this is an arguable point. My winning the lottery is contingent on my buying a winning ticket? If you think it's a problem, go ahead and add an issue for it. As you say, this is a hangover from the previous implementation.

rjyounes (approved):

ontologies/gistCore.ttl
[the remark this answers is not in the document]
uscholdm: Its a small point with little practical implications - I'll leave it for now.

PR #1077, opened 2024-04-15
Feature/new issue 831 add event specification

### LINKED ISSUE #831: Define a class EventSpecification (was HypotheticalEvent)
This would be a subclass of `Event` for dateless events - I.e., max 0 start and end dates. Note that unlike `PlannedEvent`s, which remain `PlannedEvents` even when they have occurred, a `HypotheticalEvent` must cease to be one as soon as it gets dates, either planned or actual.

Related to the removal of date restrictions in version 12.0.0; see issue #760 and PR #777.

uscholdm:
Fixes #831 
- Added `EventSpecification`.
- Added and updated annotations for `Specification`.

OPEN QUESTION: there is now an incorrect subclass relationship.

`Requirement`: gist 12.0: 
- "A documented physical or functional need that a particular design, product, or process must be able to perform.  Alternately, the obligation of a person or organization to behave in a certain way (i.e., drive on the right side of the road)."

Updated definition of `Specification`:
- "One or more characteristics that specify what it means to be a particular type of thing such as a material, product, service or event. A specification is sufficiently precise to allow conformance of something to the specification to be tested."

A Specification is often not a need to be performed, so Specification cannot be a subclass of Requirement.  It includes the idea of a law or regulation which is outside the scope of a specification.
 
Options include:
1. Broaden the definition of Requirement to include Specification so the subclass relationship can remain. Easier said than done. 
2. Remove the class Requirement, it's never been use in my experience, if anyone want to represent a law or regulation, they can create a separate subclass.

### DIFF
--- a/docs/release_notes/Issue-831-EventSpecification.md
+++ b/docs/release_notes/Issue-831-EventSpecification.md
@@ -1,0 +1,7 @@
+## Release 13.0.0
+
+### Major Updates
+
+### Minor Updates
+- Added EventSpeciifcation [#831](https://github.com/semanticarts/gist/issues/831).
+- Added and updated annotations for Specification Issue [#831](https://github.com/semanticarts/gist/issues/831).
--- a/docs/release_notes/Issue-831-EventSpecification.md
+++ b/docs/release_notes/Issue-831-EventSpecification.md
@@ -1,7 +1,8 @@
+
 ## Release 13.0.0
 
 ### Major Updates
 
 ### Minor Updates
-- Added EventSpeciifcation [#831](https://github.com/semanticarts/gist/issues/831).
-- Added and updated annotations for Specification Issue [#831](https://github.com/semanticarts/gist/issues/831).
+- Added `EventSpecification` [#831](https://github.com/semanticarts/gist/issues/831).
+- Added and updated annotations for `Specification` [#831](https://github.com/semanticarts/gist/issues/831).
--- a/docs/release_notes/Issue-831-EventSpecification.md
+++ b/docs/release_notes/Issue-831-EventSpecification.md
@@ -4,5 +4,5 @@
 ### Major Updates
 
 ### Minor Updates
-- Added `EventSpecification` [#831](https://github.com/semanticarts/gist/issues/831).
-- Added and updated annotations for `Specification` [#831](https://github.com/semanticarts/gist/issues/831).
+- Added `gist:EventSpecification`. Issue [#831](https://github.com/semanticarts/gist/issues/831).
+- Added and updated annotations for `gist:Specification`. Issue [#831](https://github.com/semanticarts/gist/issues/831).
--- a/docs/release_notes/Issue-831-EventSpecification.md
+++ b/docs/release_notes/Issue-831-EventSpecification.md
@@ -2,7 +2,9 @@
 ## Release 13.0.0
 
 ### Major Updates
+- `gist:Specification` is now a direct subclass of `gist:Intention`. Issue [#831](https://github.com/semanticarts/gist/issues/831)
 
 ### Minor Updates
+- Updated definition of `gist:Requirement`. Issue [#831](https://github.com/semanticarts/gist/issues/831)
 - Added `gist:EventSpecification`. Issue [#831](https://github.com/semanticarts/gist/issues/831).
 - Added and updated annotations for `gist:Specification`. Issue [#831](https://github.com/semanticarts/gist/issues/831).
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -2255,8 +2255,10 @@ gist:SimpleUnitOfMeasure
 gist:Specification
 	a owl:Class ;
 	rdfs:subClassOf gist:Requirement ;
-	skos:definition "One or more requirements to be satisfied by a material, design, product, or service."^^xsd:string ;
+	skos:definition "One or more characteristics that specify what it means to be a particular type of thing such as a material, product, service or event. A specification is sufficiently precise to allow conformance of something to the specification to be tested."^^xsd:string ;
+	skos:example "The specification of the iPhone 14; hypothetical events covered by a home owner's insurance policy."^^xsd:string ;
 	skos:prefLabel "Specification"^^xsd:string ;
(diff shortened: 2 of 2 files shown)


rjyounes:
@uscholdm Is this PR ready for review? You've requested my review but it's still in draft state.

rjyounes:
You could tweak the definition of `Requirement` to include the sense of `Specification` that includes `EventSpecification`. This makes perfect semantic sense to me: a certain weather event is required to have certain characteristics in order to be a hail storm and thus covered under insurance policy XYZ. 

What I really don't like is that - as acknowledged in the current definition - `Requirement` merges two entirely separate concepts, presumably because they use the same English word - that's a trap we shouldn't fall into. Is this what you are suggesting? - to remove `Requirement` and promote `Specification` and `Restriction` to direct subclasses of `Intention`? Thus not combining the two different meanings of "requirement" into one class?

That might be nice, but that is too large a change to be rolled into this PR without group consideration. So for right now I would choose the first option I mentioned, and add a new issue for the second, to be considered for version 13.1.0.

uscholdm:
> What I really don't like is that - as acknowledged in the current definition - `Requirement` merges two entirely separate concepts, presumably because they use the same English word - that's a trap we shouldn't fall into. Is this what you are suggesting? - to remove `Requirement` and promote `Specification` and `Restriction` to direct subclasses of `Intention`? Thus not combining the two different meanings of "requirement" into one class?

Yes exactly! 

> That might be the right thing to do, but that is too large a change to be rolled into this PR without group consideration. 

agreed

> So for right now I would choose the first option I mentioned, and add a new issue for the second, to be considered for version 13.1.0.

I tried a bunch of wording tweaks and failed to find something much better.  I propose to just leave it for now and add an issue.  It's only by chance I happened to notice anyway, and its only an annotation for a class that as far as I know rarely gets used.

docs/release_notes/Issue-831-EventSpecification.md
```diff
@@ -0,0 +1,8 @@
+
+## Release 13.0.0
+
+### Major Updates
+
+### Minor Updates
+- Added `EventSpecification` [#831](https://github.com/semanticarts/gist/issues/831).
```
rjyounes, line 7:
```suggestion
- Added `gist:EventSpecification`. Issue [#831](https://github.com/semanticarts/gist/issues/831).
```

docs/release_notes/Issue-831-EventSpecification.md
```diff
@@ -0,0 +1,8 @@
...
+- Added and updated annotations for `Specification` [#831](https://github.com/semanticarts/gist/issues/831).
```
rjyounes, line 8:
```suggestion
- Added and updated annotations for `gist:Specification`. Issue [#831](https://github.com/semanticarts/gist/issues/831).
```

ontologies/gistCore.ttl
```diff
@@ -944,6 +944,15 @@ gist:Event
 		;
 	.
 
+gist:EventSpecification
+	a owl:Class ;
+	rdfs:subClassOf gist:Specification ;
+	skos:definition "A characterization of an event that might happen."^^xsd:string ;
+	skos:example "A hail storm that is covered by an insurance policy.  Someone defaults on a loan."^^xsd:string ;
```
rjyounes, line 951: How about something like:

An insurance company defines the characteristics of a weather event that must be satisfied for it to qualify as a hail storm covered in its homeowner's policy. Defaulting on a loan.

ontologies/gistCore.ttl
```diff
@@ -944,6 +944,15 @@ gist:Event
...
+	skos:prefLabel "Event Specification"^^xsd:string ;
+	skos:scopeNote "It is useful for risk assessment and insurance policies."^^xsd:string ;
```
rjyounes, line 953:
```suggestion
	skos:scopeNote "This concept is useful for risk assessment and insurance policies."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -2246,8 +2255,10 @@ gist:SimpleUnitOfMeasure
 gist:Specification
 	a owl:Class ;
 	rdfs:subClassOf gist:Requirement ;
-	skos:definition "One or more requirements to be satisfied by a material, design, product, or service."^^xsd:string ;
+	skos:definition "One or more characteristics that specify what it means to be a particular type of thing such as a material, product, service or event. A specification is sufficiently precise to allow testing conformance to the specification."^^xsd:string ;
+	skos:example "The specification of the iPhone 14; hypothetical events covered by a home owner's insurance policy."^^xsd:string ;
```
rjyounes, line 2259:
```suggestion
	skos:example "The specification of the iPhone 14; hypothetical events covered by a homeowner's insurance policy."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -2246,8 +2255,10 @@ gist:SimpleUnitOfMeasure
 gist:Specification
 	a owl:Class ;
 	rdfs:subClassOf gist:Requirement ;
-	skos:definition "One or more requirements to be satisfied by a material, design, product, or service."^^xsd:string ;
+	skos:definition "One or more characteristics that specify what it means to be a particular type of thing such as a material, product, service or event. A specification is sufficiently precise to allow testing conformance to the specification."^^xsd:string ;
```
rjyounes, line 2258:
```suggestion
	skos:definition "One or more characteristics that specify what it means to be a particular type of thing such as a material, product, service or event. A specification is sufficiently precise to allow evaluating conformance to the specification."^^xsd:string ;
```

"Evaluating" seems like a more accurate term in the case of event specifications.

rjyounes (changes requested):
Looks good. I've proposed some wording tweaks.

rjyounes:
@uscholdm I've created issue #1087 for the `Requirement` proposal.

uscholdm (commented):
@rjyounes This is ready for review again.

rjyounes (changes requested):

uscholdm (commented):
I don't know why I am being asked to review my own PR

ontologies/gistCore.ttl
```diff
@@ -944,6 +944,15 @@ gist:Event
 		;
 	.
 
+gist:EventSpecification
+	a owl:Class ;
+	rdfs:subClassOf gist:Specification ;
```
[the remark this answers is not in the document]
uscholdm: You could, but Im not sure we want to.  It says if anything conforms to the event spec, then that thing must be an event. Its rather self-evident, and i suspect we could do this sort of thing in many places.

rjyounes (approved):

PR #1083, opened 2024-04-18
Change direction of some inverse pairs

### LINKED ISSUE #1070: Change some inverses
Based on internal discussion, change:
* `hasPart` to `isPartOf` (also `hasDirectPart` to `isDirectPartOf`)
* `hasMember` to `isMemberOf`
* Check for others

Some of us think that we made the wrong decisions in some cases when selecting which member of a pair of inverses to keep and which to remove. In general, we believe that the child-to-parent principle should be the determining factor in all cases, whereas in the initial implementation we brought in other considerations. This will also make our policy uniform and coherent.

The old properties can be deprecated or deleted, depending on whether this gets done in a minor or major version.

rjyounes:
Closes #1070.

### DIFF
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -1294,7 +1294,7 @@ gist:Landmark
 			gist:PhysicalIdentifiableItem
 			[
 				a owl:Restriction ;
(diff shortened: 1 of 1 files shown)


uscholdm (approved):
It's a pity there are so many verbose restrictions now.

PR #1085, opened 2024-04-18
Remove occupies predicates and shorten isGeographicallyContainedIn 

### LINKED ISSUE #809: Remove `occupiesGeographically` and `occupiesGeographicallyPermanently`?

### LINKED ISSUE #812: Shorten geo predicates
Some of us think the shift to more explicit, but longer, predicates in gist 11 was overkill. The proposal is to go back to something like (without inverses, which are eliminated in #506):

* geoContains
* geoOccupies
* geoOccupiesPermanently 

However, the occupies predicates are considered for removal in issue #809.

rjyounes:
Fixes #809 - see https://github.com/semanticarts/gist/issues/947#issuecomment-1679565100
Fixes #812 

NB: I've deliberately kept the full prefLabel "is geographically contained in" for `isGeoContainedIn`. This is a bending of our rule about constructing pref labels, but in this case is warranted, in my opinion. I've updated the style guide to reflect this change.

### DIFF
--- a/docs/gistStyleGuide.md
+++ b/docs/gistStyleGuide.md
@@ -12,6 +12,7 @@
   - [Classes](#classes)
   - [Properties](#properties)
   - [Valid Exceptions](#valid-exceptions)
+  - [gist:nonConformingLabel](#gistnonconforminglabel)
   - [gist Definition of Title Case](#gist-definition-of-title-case)
 - [Annotations](#annotations)
   - [Conventions for Use](#conventions-for-use)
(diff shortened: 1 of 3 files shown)


docs/gistStyleGuide.md
```diff
@@ -130,7 +131,14 @@ Note: As of version 12.0.0, gist itself does not itself follow the infix convent
...
 - Normalized to natural language standards. E.g., hyphens inserted, acronyms in all caps, proper nouns capitalized, etc.
-- Examples: *has unit of measure*, *has SSN*, *unit symbol Unicode*
+- Examples: *has unit of measure*, *has SSN*
+
+### Valid Exceptions
+
+There may occasionally be valid reasons to deviate from the conventions stated here:
+
+- Deviation from wording of the local name. For example, the predicate `gist:isGeoContainedIn` uses a shortened form of "geographically" for conciseness. The `skos:prefLabel` uses the fully spelled out word: "is geographically contained in."
+- Deviation from typographical conventions such as case. For example, a proper name in a property label could appropriately be capitalized. Since this will fail the SHACL validation constraints during the ontology build and release process, add the term to the `gistValidationAnnotations.ttl` file so that label validation will be skipped.
```
kchastain703, line 141: Should we give an example of this, too?

uscholdm (approved):
Looks good.

docs/gistStyleGuide.md
dylan-sa: I agree w/ @kchastain703 that an example would be helpful.

dylan-sa, line 141: In the 'Properties' section above, it already says we can deviate from lower case for proper nouns. So the exception here doesn't seem like a genuine exception to the convention as stated. Maybe we can say the convention is simply lowercase, and then the bit about normalizing to natural language standards can be left to the 'Valid Exceptions' section?

dylan-sa (changes requested):

docs/gistStyleGuide.md
rjyounes: I can't think of a good one. I think the text following "For example..." is pretty self-explanatory, so I'm going to resolve this.

rjyounes: This is a good point. I think I'll just remove that point and add a note about the validation annotations.

PR #1089, opened 2024-04-25
Removes sensor-related terms to fix issue 462

### LINKED ISSUE #462: Remove sensor-related terms
I think modeling of sensors is too domain-specific for an upper ontology.

Related terms (possibly incomplete):

```
hasViableRange
respondsTo
Controller
Sensor
Actuator
```

Look in old IoT module for others.

turingtest37:
Fixes #462


ontologies/Issue-462-Remove-Sensor-Related-Items.md
```diff
@@ -0,0 +1,17 @@
+## Release 13.0.0
+
+### Major Updates
+
+- Removed sensor-related concepts. Issue [#462](https://github.com/semanticarts/gist/issues/462)
+
+- Classes removed:
+  - `gist:Actuator`
+  - `gist:Controller`
```
Jamie-SA, line 9:
```suggestion
  - `gist:Controller`
  - `gist:ControllerType`
```

Jamie-SA (changes requested):

ontologies/gistCore.ttl
```diff
@@ -1518,7 +1477,7 @@ gist:Message
 gist:MessageDefinition
 	a owl:Class ;
 	rdfs:subClassOf gist:SchemaMetaData ;
-	skos:definition "Each pulse from a Sensor is reflected in a message, as well as each instruction to an Actuator"^^xsd:string ;
+	skos:definition "Each communication to and from a remote device or application programming interface is reflected in a message."^^xsd:string ;
```
uscholdm, line 1480: This looks to be specific to sensors. I propose to remove the class MessageDefinition.

uscholdm (commented):
- `gist:MessageDefinition` looks to be sensors-specific. I propose that we remove it unless there is justification for having it independently from sensors.
- `gist:directs` looks to be sensors-specific. I propose that we remove it unless there is justification for having it independently from sensors.
- `gist:accepts` looks to be sensors-specific. I propose that we remove it unless there is justification for having it independently from sensors.

rjyounes:
DECISION:

* Remove `MessageDefinition` and `directs`
* Rewrite definition of `accepts` as "Relates something to the types of inputs it accepts."

rjyounes (changes requested):
See decision in main comment thread.

ontologies/Issue-462-Remove-Sensor-Related-Items.md
rjyounes: Should remove `PhenomenaType`

ontologies/Issue-462-Remove-Sensor-Related-Items.md
```diff
@@ -0,0 +1,17 @@
...
+  - `gist:Sensor`
+  
+- Properties removed:
+  - `gist:hasViableRange`
+  - `gist:respondsTo`
+
+- Changed `skos:definition` of `gist:directs` to remove sensor references.
```
rjyounes, line 16: We should delete this.

ontologies/gistCore.ttl
```diff
@@ -3010,7 +2955,8 @@ gist:description
 
 gist:directs
 	a owl:ObjectProperty ;
-	skos:definition "The set of actuators that a controller can affect"^^xsd:string ;
+	skos:definition "To issue orders to someone or something by right of authority."^^xsd:string ;
```
rjyounes, line 2958: We have agreed to delete this term: https://github.com/semanticarts/gist/pull/1089#issuecomment-2102900192

rjyounes (changes requested):

ontologies/gistCore.ttl
turingtest37: I'm not sure I agree. A Message Definition is not specific to sensors at all. Any data exchange with a remote service / API will have a message definition. I think it makes sense to leave this class as an example of what SchemaMetaData represents.

rjyounes: Michael and I agree that this should be removed by the same logic as `gist:directs`.

rjyounes:
@uscholdm  to contact @turingtest37 to resolve questions about term deletions.

ontologies/gistCore.ttl
uscholdm: The idea is that, as originally conceived, MessageDefinition was specifically aligned with sensor concepts, and thus should be removed. 

@turingtest37 Separately, if we think that there is a need for something like MessageDefinition that has broader use outside the sensors context, then a new issue to that effect should be created, just like we do for adding a totally new concept to gist.

uscholdm (changes requested):
Remove MessageDefinition

rjyounes (approved):

PR #1096, opened 2024-05-24
Added Determination class.

### LINKED ISSUE #317: New subclass of Event: Determination
There are many examples of events whose central purpose is to determine some kind of  result, value or outcome.  Determinations will always have:
1. Method what was done to make the determination - (e.g. manual process, algorithm)
2. Result: usually a value or a category or some sort
3. Performer: of the determination (a person, a program/application)
4. Date performed (inherited from Event)

The most common  examples in my experience are:
- Measurement
    - the sulphur content of Brent crude
    - blood pressure
- Evaluation - based on quality.  
    - loan approval (from FIBO)
    - how good of a modeler is an interviewee?
- Assessment [quality-neutral]
    - What will the price of Brent crude be next month?
- Classification: what category does something belong in?
    - is a person short, medium height or tall?

There is a a bit of overlap here.  
- Sometimes categories will be in some kind of notional order, which makes it more like (e.g. bad, neutral, good).  
- the word Assessment often is used in the context of quality.

uscholdm:
Closes #317.

I made the class a direct subclass of `Event` because a `Determination` need not be historical, it may be planned but unscheduled, or scheduled, in the process of happening, or already happened.


ontologies/gistCore.ttl
```diff
@@ -787,6 +787,15 @@ gist:DegreeOfCommitment
 	skos:prefLabel "Degree Of Commitment"^^xsd:string ;
 	.
 
+gist:Determination
+	a owl:Class ;
+	rdfs:subClassOf gist:Event ;
+	skos:definition "An event whose central purpose is to obtain some kind of result, value or outcome. The basis for the determination can be qualitative, quantitative or a combination of both. "^^xsd:string ;
```
rjyounes, line 793: This definition seems a little off, perhaps because of the use of "obtain." Many events are held to obtain a result - e.g., a campaign speech has the purpose of gathering support, but this is not a determination. 

Some suggestions based on various dictionary definitions:
* An event whose purpose is to establish something by calculation or research. The basis for the determination can be..."
* The act of measuring, evaluating, or calculating something.
* The process of finding out or calculating something.
* The act of determining the properties of something, usually by research or calculation.

ontologies/gistCore.ttl
```diff
@@ -787,6 +787,15 @@ gist:DegreeOfCommitment
...
+	skos:example "Measuring the sulphur content crude oil. Evaluation of a loan application for approval. Estimating the price of gas for the next three months. Determining whether and by how much an interest rate should change. Classifying something."^^xsd:string ;
+	skos:prefLabel "Determination"^^xsd:string ;
+	skos:scopeNote "This class is meant to exclude cases such as the chemical makeup of food determining its taste. "^^xsd:string ;
```
rjyounes, line 796: This is not needed if you modify the definition as suggested.

rjyounes (changes requested):

ontologies/gistCore.ttl
uscholdm: How about "An event whose purpose is to establish a specific result, value, or outcome usually by research, measuring, evaluating, or calculating."

ontologies/gistCore.ttl
```diff
@@ -787,6 +787,14 @@ gist:DegreeOfCommitment
 	skos:prefLabel "Degree Of Commitment"^^xsd:string ;
 	.
 
+gist:Determination
```
marksem, line 790: I like the work done so far. 

Q: Is there any property that would always be the case on a determination?  Otherwise this is just a named class.

ontologies/gistCore.ttl
rjyounes: That's good, but needs a comma after "outcome."

rjyounes (changes requested):
I like your proposed definition, so will approve this once you've added it and removed the scope note.

ontologies/gistCore.ttl
uscholdm: @marksem In real world terms, a determination 
- will always have an entity that does the determining (maybe inverse of `produces`) with the doer being a person, org or program
- will always have an outcome (in the past I have used `hasOutput` for this)
- will often have a method (min 0).

uscholdm (commented):
Not sure why Im being asked to review my own PR

rjyounes:
@uscholdm I'm leaving this unmerged to consider whether we want to name it `DeterminationEvent` rather than `Determination`. Looks like it also needs to be updated with latest changes on develop.

uscholdm:
> @uscholdm I'm leaving this unmerged to consider whether we want to name it `DeterminationEvent` rather than `Determination`. Looks like it also needs to be updated with latest changes on develop.

Lets leave it as it is. Its a subclass of `Event`.

rjyounes (approved):

Issue #1091, opened 2024-04-25
Migration scripts for major changes in 13.0.0


rjyounes:
Phil has created migration scripts for the units and magnitudes model (PR #1071), so no need to deal with those.

Major changes requiring migration scripts:

* Address model - ask Dan, Dylan, Michael, or me for help 
  * Minimally change `hasCommunicationAddress` to `hasAddress`. Work with Dan to see what else can be automated.
* geo-predicate names - PR #1085
* Inverses - PR #1083
* ~~Remove sensor-related terms - PR #1089 - no migration here, they will just have to stop using them~~

Thanks!

rjyounes:
@DanCarey404  is away until July 5, please consult @rjyounes, @dylan-sa, @kchastain703, or @uscholdm to see what changes, if any, can be automated.

uscholdm:
> @DanCarey404 is away until July 5, please consult @rjyounes, @dylan-sa, @kchastain703, or @uscholdm to see what changes, if any, can be automated.

This would take some time to figure out what is possible and then to write the scripts. I suspect that there may be a few things that are automatable.

rjyounes:
Sample triples:

1. geo-predicate names - PR #1085 


```
ex:_City_minneapolis gist:isGeographicallyContainedIn ex:_State_minnesota => ex:_City_minneapolis gist:isGeoContainedIn ex:_State_minnesota 

ex:_thing1 gist:occupiesGeographicallyPermanently ex:_thing2  => ex:_thing1 gist:hasPhysicalLocation ex:_thing2 

ex:_thing1 gist:occupiesGeographically ex:_thing2  => ex:_thing1 gist:hasPhysicalLocation ex:_thing2 
```


2. Inverses - PR #1083

```
ex:_Collection_1 gist:hasMember ex:_Person_1 => ex:_Person_1 gist:isMemberOf ex:_OrderedCollection_1

ex:_OrderedCollection_1 gist:hasFirstMember ex:_OrderedMember_1 => ex:_OrderedMember_1 gist:isFirstMemberOf ex:_OrderedCollection_1

ex:thing1 gist:hasPart ex:thing2 => ex:thing2 gist:isPartOf ex:thing1 

ex:thing1 gist:hasDirectPart ex:thing2 => ex:thing2 gist:isDirectPartOf ex:thing1 
```

3. Address model - PR #1106 

```
[...]

PR #1116, opened 2024-06-30
non-uom data conversion scripts

philblackwood:
Data conversion scripts for non-uom changes. Closes issue #1091.


sa-bpelakh:
@philblackwood  I will add the bundle drivers directly into this PR, if that's OK. I don't think you need to do the named graph exclusion on the default graph queries because the pattern should already have been transformed there.

migration/v13.0/queries/replaceHasCommunicationAddress1.rq
```diff
@@ -0,0 +1,25 @@
+# replace gist:hasCommunicationAddress in named graphs with gist:hasAddress
```
rjyounes, line 1: Phil, can't we include the `hasAddress/hasCommunicationAddress` replacement in the general `renameProperties` files and eliminate the two additional files?

rjyounes (changes requested):

migration/v13.0/queries/replaceHasCommunicationAddress1.rq
philblackwood: yes! I will update the renameProperties scripts to include the pair:

oldProperty: gist:hasCommunicationAddress

newProperty: gist:hasAddress.

philblackwood:
From Boris:  I don't think you need to do the named graph exclusion on the default graph queries because the pattern should already have been transformed there."

me: I think it's OK to leave it in. The scenario where it helps is:

   - user runs first script to make changes in named graph, then puts the data conversion aside for a while
   -  new changes are made in the named graph
   -  user returns to data conversion, runs second script which results in named graph triples removed and their replacement incorrectly gets added to the default graph.

I looked at making the change, and it is big enough (and my typing bad enough) to warrant a re-test, which I would prefer not to have to do.

Are you OK with leaving it in?

philblackwood:
@sa-bpelakh "I will add the bundle drivers directly into this PR, if that's OK."

phil: yes, certainly.

philblackwood:
Adding data conversion for deprecating hasSuperCategory and its sub properties.

rjyounes:
@philblackwood @sa-bpelakh Is this ready to go? It needs to be updated with `develop`.

migration/v13.0/non-uomDataConversion.md
```diff
@@ -0,0 +1,34 @@
...
+
+Be sure to read the header of each data conversion script.
+
+## Examples:
+- renameProperties replaces `gist:isGeographicallyContainedIn` with `gist:isGeoContainedIn`
+
+- replaceAddressClasses replaces `gist:EmailAddress` with `gist:ElectronicAddress`
+- replaceAddressClasses replaces `gist:StreetAddress` with `gist:PhysicalAddress`
+
+The different types and uses of addresses can be modeled as categories.
```
rjyounes, line 17:
```suggestion
The different types and uses of addresses can be modeled as categories; see [the address migration documentation(AddressMigration.md) as well as the [full documentation of the new address model](/docs/model_documentation/AddressGuidance.md).
```

rjyounes (changes requested):

PR #1118, opened 2024-07-01
Add recent release dates in a skos:historyNote

### LINKED ISSUE #1017: Cannot tell gist ontology release date by looking in the ontology
One can't tell gist ontology release date by looking in the ontology.
It would be nice to be able to do so.
Suggest we add indication of gist release date as an ontology annotation.
Be sure to use a property that will keep gist as OWL-DL.

rjyounes:
Closes #1017.

In addition to the requested changes, I moved a release note out of the ontologies folder into the correct folder. 

We did not discuss whether release dates should be added to the accompanying ontologies - `gistMediaTypes` and `gistPrefixDeclarations`. I did not do so because they are used only as supplements to `gistCore`, but let me know if you have a strong feeling that they should be.


rjyounes:
Date of 13.0.0 might need to change if release slips.

uscholdm (approved):

Issue #1115, opened 2024-06-28
Deprecate gist:hasSuperCategory and use gist:hasBroader instead


rjyounes:
Since we introduce `hasBroader` in gist 13.0.0, it is now confusing that we have it in addition to `gist:hasSuperCategory`. `hasBroader` is meant to apply to more cases, such as aspects, than is `hasSuperCategory`, but it could apply equally well to categories. I propose that we:

- Deprecate `gist:hasSuperCategory` and `gist:hasDirectSuperCategory`
- Define a new property `gist:hasDirectBroader` that serves the same purpose as `gist:hasDirectSuperCategory`

rjyounes:
DECISION:
* Delete `hasSuperCategory`
* Change `hadDirectSuperCategory` to `hasDirectBroader` and make subproperty of `hasBroader`
* Change `hasUniqueSuperCategory` to `hasUniqueBroader` and make subproperty of `hasBroader`
* Rewrite any relevant restrictions (e.g., on `Taxonomy`)

stevenchalem:
Maybe this is overly pedantic, but shouldn't the textual definition of `broader` be something like "Relates a thing to another thing with a broader meaning _that encompasses it_" or "_that subsumes it_"? "Citrus Fruits" is broader than "Limes", and "Vertebrates" is broader than "Mammals". Now, "Limes" is a pretty narrow category and "Vertebrates" is a pretty broad category, so we could say the "Vertebrates" is broader than "Limes".  But we're not meaning to talk just about broadness here; we also mean the the object encompasses the subject.

philblackwood:
Also modify data conversion scripts to rename properties

philblackwood:
@stevenchalem we might want to tune the definition in the future. My first impression is that people will understand the simple definition provided as it is intended (I hope so).

uscholdm:
> Maybe this is overly pedantic, but shouldn't the textual definition of `broader` be something like "Relates a thing to another thing with a broader meaning _that encompasses it_" or "_that subsumes it_"? "Citrus Fruits" is broader than "Limes", and "Vertebrates" is broader than "Mammals". Now, "Limes" is a pretty narrow category and "Vertebrates" is a pretty broad category, so we could say the "Vertebrates" is broader than "Limes". But we're not meaning to talk just about broadness here; we also mean the the object encompasses the subject.

Your point is accurate, and it would not be a bad idea to clarify as you say, for those folk who may wonder what we mean.

[...]

PR #1120, opened 2024-07-02
deprecate hasSuperCategory and rename its sub properties

philblackwood:
hasSuperCategory is deprecated (redundant with gist:hasBroader)

rename hasDirectSuperCategory to hasDirectBroader, a sub property of hasBroader

rename hasUniqueSuperCategory to hasUniqueBroader, a sub property of hasBroader

change class expressions as needed

closes issue #1115

### DIFF
--- a/docs/release_notes/issue1115-deprecate-hasSuperCategory
+++ b/docs/release_notes/issue1115-deprecate-hasSuperCategory
@@ -1,0 +1,8 @@
+
+### Major updates
+
+- Removed `gist:hasSuperCategory` and `gist:hasDirectSuperCategory` and `gist:hasUniqueSuperCategory`
+- Added `gist:hasDirectBroader` and `gist:hasUniqueBroader`
+- `gist:hasSuperCategory` is replaced by `gist:hasBroader`
+- `gist:hasDirectSuperCategory` is replaced by `gist:hasDirectBroader`
+- `gist:hasUniqueSuperCategory` is replaced by `gist:hasUniqueBroader`
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -1861,13 +1861,13 @@ gist:Taxonomy
 								[
 									a owl:Restriction ;
 									owl:onProperty [
-										owl:inverseOf gist:hasSuperCategory ;
+										owl:inverseOf gist:hasBroader ;
 									] ;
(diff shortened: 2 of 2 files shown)


ontologies/gistCore.ttl
```diff
@@ -2689,12 +2689,12 @@ gist:hasBroader
...
-	skos:definition "The subject category is a subcategory of the object category. This property defines the direct links in a category hierarchy; no intermediate categories can exist between the direct links."^^xsd:string ;
-	skos:prefLabel "has direct supercategory"^^xsd:string ;
-	skos:scopeNote "Unlike its superproperty gist:hasSuperCategory, this property is not transitive. It is essentially the same as the non-transitive skos:broader, using gist:Category rather than skos:Concept."^^xsd:string ;
+	rdfs:subPropertyOf gist:hasBroader ;
+	skos:definition "Relates a thing to another thing with a broader meaning, when there is no intermediate between them."^^xsd:string ;
+	skos:prefLabel "has direct broader"^^xsd:string ;
+	skos:scopeNote "Unlike gist:hasBroader, this property is not transitive. It is safest to use this property when semantic directness is inherent in the relationship. Otherwise, there is a risk of making a hasDirectBroader assertion and then later inserting an intermediate part; this will result in making an asserted triple false even though there was no change in the world. When in doubt, use the transitive version gist:isBroader."^^xsd:string ;
```
Jamie-SA, line 2697: Happened to notice this:
```suggestion
	skos:scopeNote "Unlike gist:hasBroader, this property is not transitive. It is safest to use this property when semantic directness is inherent in the relationship. Otherwise, there is a risk of making a hasDirectBroader assertion and then later inserting an intermediate part; this will result in making an asserted triple false even though there was no change in the world. When in doubt, use the transitive version gist:hasBroader."^^xsd:string ;
```

rjyounes (approved):

docs/release_notes/issue1115-deprecate-hasSuperCategory
```diff
@@ -0,0 +1,8 @@
+
+### Major updates
+
+- Removed `gist:hasSuperCategory` and `gist:hasDirectSuperCategory` and `gist:hasUniqueSuperCategory`
```
uscholdm, line 4: @philblackwood @rebecca This is unnecessary and somewhat misleading.

```suggestion

PR #1147, opened 2024-08-15
Patch migration docs

### LINKED ISSUE #1140: Correct migration scripts for inverse properties
Applies to both v13 and v14 migration scripts: they do not handle use of the properties in OWL restrictions.  For example:

```
[
    a owl:Restriction ;
    owl:onProperty gist:hasPart ;
    owl:someValuesFrom ex:myClass ;
]
```
Has to be changed to:
 ```
[
    a owl:Restriction ;
    owl:onProperty [ owl:inverseOf gist:isPartOf ] ;
    owl:someValuesFrom ex:myClass ;
]
 ```

@Jamie-SA Do you know whether we are attempting to cover this case? If not we should at least warn users that our scripts will produce the wrong result in these cases.

### LINKED ISSUE #1142: Issue with migration scripts and local files
One use case for the migration scripts is to run them over local ontology/vocabulary files to capture updates. Our README points out limited support for named graphs when the scripts are run over local files. Since named graphs are used in `queries/action/default`, running these queries over local files generates the following error (as expected):
```
raise Exception(
Exception: You performed a query operation requiring a dataset (i.e. ConjunctiveGraph), but operating currently on a single graph.
```
Commenting out the `minus {named graph pattern}` portion of these queries gets them to work. I think the queries in this directory will still work well when run from a triplestore, but I wonder if we can find a way to better accommodate the local file use case. This might just be adding some additional notes to the README or queries. 

See also #1140.

philblackwood:
closes #1140 
closes #1142 

Text added to describe manual steps needed.

### DIFF
--- a/docs/MajorVersionMigration.md
+++ b/docs/MajorVersionMigration.md
@@ -10,6 +10,11 @@ For changes that are not amenable to automatic migration, queries will be provid
 report the use of deleted or modified classes and properties so that mitigation measures
 can be initiated.
 
+Changes that need to be done manually may include:
+- domains and ranges
+- subproperty assertions
+- restrictions, when a property is replaced with its inverse.
+
 Every major version of `gist` (starting with 10.0.0) will add a sub-directory under the
 `migration` directory containing the queries and two migration scripts:
 
--- a/migration/v13.0/README.md
+++ b/migration/v13.0/README.md
@@ -17,6 +17,12 @@ by the gist team to help migrate from the previous version of gist.
   about changes that were made in gist and you will need to decide how you want
   to handle the changes.
 
+- **Manual updates for inverses:** Migration scripts do not update all class expressions in an ontology. In particular for gist 13, 'owl:onProperty gist:hasPart' needs to be changed to 'owl:onProperty [ owl:inverseOf gist:isPartOf ]'. Class expressions that include `gist:hasMember` need to be modified in a similar way.
+
+- **Update all related artifacts.** Be sure to make updates as needed to artifacts that refer to changed parts of gist, which can include ontologies, taxonomies, data ingestion pipelines, data validations, queries, forms, documention, etc.
+
+- **Patch:** comment out the statements 'minus {named graph pattern}' in migration scripts for local files.
+
 ## Migration Queries
 
 All of our migration tools are SPARQL queries.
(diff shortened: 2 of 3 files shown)


docs/MajorVersionMigration.md
```diff
@@ -8,7 +8,7 @@ the release package will include queries to facilitate the migration of existing
 
 For changes that are not amenable to automatic migration, queries will be provided that
 report the use of deleted or modified classes and properties so that mitigation measures
-can be initiated.
+can be initiated. For example, some may releases require manual intervention when a property is being replaced by its inverse (class expressions, restrictions, and definitions of subproperties).
```
rjyounes, line 11:
```suggestion
can be initiated. For example, some releases require manual intervention when a property is being replaced by its inverse (in class expressions, restrictions, and definitions of subproperties).
```

migration/v13.0/README.md
```diff
@@ -17,6 +17,12 @@ by the gist team to help migrate from the previous version of gist.
   about changes that were made in gist and you will need to decide how you want
   to handle the changes.
 
+- **Manual updates for inverses:** Migration scripts do not update all class expressions in an ontology. In particular for gist 13, `owl:onProperty gist:hasPart` needs to be changed to `owl:onProperty [ owl:inverseOf gist:isPartOf ]`. Class expressions that include `gist:hasMember` need to be modified in a similar way. If you have created subproperties of `gist:hasPart` or `gist:hasMember` they will also need to be updated manually.
```
rjyounes, line 20:
```suggestion
- **Manual updates for inverses:** Migration scripts do not update all class expressions in an ontology. In particular for gist 13, `owl:onProperty gist:hasPart` needs to be changed to `owl:onProperty [ owl:inverseOf gist:isPartOf ]`. Class expressions that include `gist:hasMember` or `gist:hasDirectPart` need to be modified in a similar way. If you have created subproperties of any of these properties, they will also need to be updated manually.
```

rjyounes (commented):
I've made a couple of wording suggestions. You also need a release note.

rjyounes (changes requested):
See suggestions.

rjyounes (changes requested):
Minor changes requested.

PR #1152, opened 2024-08-30
Differentiate goesToAgent/comesFromAgent and hasRecipient/hasGiver

### LINKED ISSUE #1024: Differentiate goestoAgent/comesFromAgent and hasRecipient/hasGiver
What is the distinction? And if none, combine.

kstudzin:
Modify definitions and add examples

Closes #1024


uscholdm (changes requested):
This is quite tricky.  Sometimes there is no reason to have both a giver and a comes from agent, nor both a recipient and a goes to agent.  In these cases, and it may not matter which to use, hence the original confusion, but we should give some guidance if possible.  I would say if either seems appropriate, then use hasGiver or hasRecipient instead of the to/form agent properties.  Shipment is a good example where there could be multiple parties. Let’s say a purchasing department orders a new printer and it is shipped using a third party.
Company that placed the order and who will receive the goods (recipient)

- The agent the package is addressed to who will pass it along to the recipient (goes to agent)
- The company that ships the package (comes from agent)
- The company that is having the printer shipped (hasGiver)

In this example there might not be a separate goes to agent.

Also, I suggest that all 4 properties have a domainIncludes gist:Event

rjyounes:
> Also, I suggest that all 4 properties have a domainIncludes gist:Event

I don't think of Event as the likeliest domain for `goesTo/comesFrom`.

rjyounes:
This issue has not been triaged yet, and I think as a group we need to come to some understanding of what the differences are meant to be.

uscholdm:
> > Also, I suggest that all 4 properties have a domainIncludes gist:Event
> 
> I don't think of Event as the likeliest domain for `goesTo/comesFrom`.

Its the only one I could think of right away, what are the obvious ones that come to mind for you?

rjyounes:
> Its the only one I could think of right away, what are the obvious ones that come to mind for you?

As the definitions say, I think of shipment, package, letter, etc., which we don't have classes for. I don't see any reason that we must have a `domainIncludes` value, so if nothing fits the bill I'd leave it off.

uscholdm:
> As the definitions say, I think of shipment, package, letter, etc., which we don't have classes for. I don't see any reason that we must have a `domainIncludes` value, so if nothing fits the bill I'd leave it off.

Shipment is an event; if a package or letter goes anywhere, that is an event, which is always a correct way to model a 'going'.  One need not always, model the event explicitly. But I cannot think of a single example of using goesToAgent where there is no event going on (so to speak) - can you?

rjyounes:
Oh, I see, I hadn't been thinking of the events - I was thinking of the objects being shipped and delivered. The locution seems odd in ordinary English - "the event of shipping the package goes to Tom" as opposed to "the package goes to Tom." The distinctions you make above between those who handle goods in passing them from the "ultimate" giver to the "ultimate" recipient are clear, but it's not at all clear to me how we decide which set of predicates goes with which. One pair of agents are "intermediate" givers and recipients, the other are "ultimate" ones. I have no clue why one is "comesTo/goesFrom" and the other is "giver/recipient." Do you?

uscholdm:
> Oh, I see, I hadn't been thinking of the events - I was thinking of the objects being shipped and delivered. The locution seems odd in ordinary English - "the event of shipping the package goes to Tom" as opposed to "the package goes to Tom." The distinctions you make above between those who handle goods in passing them from the "ultimate" giver to the "ultimate" recipient are clear, but it's not at all clear to me how we decide which set of predicates goes with which. One pair of agents are the "intermediate" givers and recipients, the other are the "ultimate" ones. I have no clue why one is "comesFrom/goesTo" and the other is "giver/recipient." Do you? Do we need to rethink our wording choices?

Those are all excellent points, the locution is not good- the event does not go to the agent, the package does.  But changing the wording will only work if we are rock solid on what we are trying to say - we may not be.   Intermediate vs. ultimate may or may not be exactly what we want. There may be too many variations of this.  I wonder if anyone ever used the comesToAgent and goesToAgent properties? I wonder if we should just remove them. Maybe a gist dev group discussion?

ontologies/gistCore.ttl
```diff
@@ -2190,7 +2194,8 @@ gist:comesFromPlace
 			gist:Place
 		) ;
 	] ;
-	skos:definition "Origin"^^xsd:string ;
+	skos:definition "Relates something in motion or something that enables motion to its place of origin."^^xsd:string ;
```
kstudzin, line 2197: I included 'something that enables motion' to cover the case where the subject is `gist:GeoSegment` because the segment itself is not in motion. 

#1165 may remove `gist:GeoSegment`, in which case we could consider also removing the phrase 'something that enables motion.' When we need to connect the pathway between an origin and a destination, we could use `gist:linksFrom` like we do with networks.

ontologies/gistCore.ttl
```diff
@@ -2626,7 +2635,8 @@ gist:goesToPlace
 			gist:Place
 		) ;
 	] ;
-	skos:definition "Destination"^^xsd:string ;
+	skos:definition "Relates something in motion or something that enables motion to its destination."^^xsd:string ;
```
kstudzin, line 2638: See `gist:comesFromPlace` comment

With #1165, we could use `gist:linksTo` to connect a pathway to its destination.

rjyounes: That's a very interesting point, and suggests that `GeoRoute` ought to be a subclass of `Network` as well as `Place`, which seems correct.

ontologies/gistCore.ttl
```diff
@@ -2612,9 +2617,13 @@ gist:goesToAgent
 			gist:Person
 		) ;
 	] ;
-	skos:definition "The party that is the recipient of something (e.g. a message, shipment, etc.)"^^xsd:string ;
+	skos:definition "Relates something in motion to the agent that receives it, concluding the motion."^^xsd:string ;
+	skos:example "A package is sent to a person; a purchase order is recieved by a seller."^^xsd:string ;
 	skos:prefLabel "goes to agent"^^xsd:string ;
-	skos:scopeNote 'This is not the inverse of gist:comesFromAgent. A message can be from someone. If we made it the inverse the person would be "to" the message'^^xsd:string ;
+	skos:scopeNote
+		"In general, 'gist:goesToAgent' is intended for when the subject moves. However, there is some overlap in when 'gist:hasRecipient' and 'gist:goesToAgent' can be used, such as recieving a gift."^^xsd:string ,
```
uscholdm, line 2624: I think this it a good way to handle the overlap when either seems appropriate.

uscholdm (changes requested):
Its a bit unclear what the intended domain of these 4 properties should be.

ontologies/gistCore.ttl
```diff
@@ -2176,9 +2176,13 @@ gist:comesFromAgent
 			gist:Person
 		) ;
 	] ;
-	skos:definition "The party that is the source of something (e.g. a message, shipment, etc.)"^^xsd:string ;
+	skos:definition "Relates something in motion to the agent that originated the motion."^^xsd:string ;
+	skos:example "A package is shipped by a person; a purchase order is sent by an organization."^^xsd:string ;
```
rjyounes, line 2180:
```suggestion
	skos:example "A package is shipped by a person; a purchase order is emailed to a vendor by an organization."^^xsd:string ;
```

Added email here so we have an example of electronic transfer.

ontologies/gistCore.ttl
```diff
@@ -2176,9 +2176,13 @@ gist:comesFromAgent
...
 	skos:prefLabel "comes from agent"^^xsd:string ;
-	skos:scopeNote 'This is not the inverse of gist:goesToAgent. A message can be to someone. If we made it the inverse the person would be "from" the message'^^xsd:string ;
+	skos:scopeNote
+		"In general, 'gist:comesFromAgent' is intended for when the subject is in motion. However, there is some overlap in when 'gist:hasGiver' and 'gist:comesFromAgent' can be used, such as giving a gift."^^xsd:string ,
+		'This is not the inverse of gist:goesToAgent. A message can be to someone. If we made it the inverse the person would be "from" the message'^^xsd:string
```
rjyounes, line 2184:
```suggestion
		'This is not the inverse of gist:goesToAgent, as both properties have the message or shipment as the subject.'^^xsd:string
```

ontologies/gistCore.ttl
rjyounes: See comments on `comesFromAgent`.

rjyounes (changes requested):
So far I've commented in detail only on one of the properties, `comesFromAgent`, and I recommend similar updates for the others. I can re-review once those have been made.

kstudzin:
@uscholdm I made that change - let me know what you think

docs/release_notes/issue1024-differentiation-giver-sender-props.md
```diff
@@ -0,0 +1,4 @@
+
+### Minor Updates
+
+- Update annotations to differentiate `gist:hasRecipient`/`gist:goesToAgent` and `gist:hasGiver`/`gist:comesFromAgent` and align definitions of `gist:goesToPlace`/`gist:goesToAgent` and `gist:comesFromPlace`/`gist:comesFromAgent`. Issue [#1024](https://github.com/semanticarts/gist/issues/1024).
```
rjyounes:
```suggestion
- Updated annotations to differentiate `gist:hasRecipient`/`gist:goesToAgent` and `gist:hasGiver`/`gist:comesFromAgent` and align definitions of `gist:goesToPlace`/`gist:goesToAgent` and `gist:comesFromPlace`/`gist:comesFromAgent`. Issue [#1024](https://github.com/semanticarts/gist/issues/1024).
```

rjyounes (approved):
Nicely done!

Issue #977, opened 2023-09-24
Change type of gist:license values to xsd:anyURI 


rjyounes:

uscholdm:
This seems like a good idea, especially after our recent internal discussion on the matter.

rjyounes:
Note that the suggestion is to change the datatype in our usage of `gist:license`, not to add a range to the property (which is an annotation property anyway).

Issue #682, opened 2022-06-08
Should `gist:license`  be an object property?


pmcb55:
Should `gist:license` be an object property?

Today we have:
```
<https://ontologies.semanticarts.com/o/gistCore> a owl:Ontology;
    owl:versionIRI <https://ontologies.semanticarts.com/o/gistCore11.0.0>;
    skos:definition "gist is a minimalist upper ontology created by Semantic Arts.";
    skos:prefLabel "gist";
    gist:license "https://creativecommons.org/licenses/by-sa/3.0/".
```
...instead, should we have:
```
    gist:license <https://creativecommons.org/licenses/by-sa/3.0/>.
```
...since that license IRI, when de-referenced, does indeed return RDF that is specifically typed as being of `rdf:type` Creative Commons `cc:License`, i.e., if you de-reference that IRI, you get:
```
@prefix cc: <http://creativecommons.org/ns#> .
@prefix dc: <http://purl.org/dc/terms/> .
@prefix ns0: <dct:> .

<http://creativecommons.org/licenses/by/4.0/>
  a cc:License ;
  dc:creator <https://creativecommons.org/> ;
  ns0:creator <https://creativecommons.org/> ;
  dc:title "Attribution-ShareAlike 3.0 Unported" ;
  ns0:title "Attribution-ShareAlike 3.0 Unported" ;
```

As an alternative (e.g., for licenses only described as HTML, and **_not_** as RDF at all), vocab creators might choose to use:
```
    gist:license "http://www.opendatacommons.org/licenses/pddl/"^^xsd:anyURI .
```
...where we are saying that the license is provided by that string literal, which is an IRI, but it isn't expected to return RDF (it just happens to return HTML only, in this case). 
...although allowing both forms of values for `gist:license` should be explicitly expressed by the `gist:license` property itself using `gist:rangeIncludes` to state all the expected/allowable value datatypes.

Personally, I'd suggest that the `gist:rangeIncludes` for `gist:license` be deliberately restricted to **_only_** RDF-returning IRIs (so that we maintain 'Turtles all the way down', as there are already good RDF resources for all the common licenses today).

uscholdm:
@pmcb55 Very interesting observation and analysis.  There is one problem, though...

>     gist:license "http://www.opendatacommons.org/licenses/pddl/"^^xsd:anyURI .

The object for an object property cannot be a literal.  So maybe we have to keep it as a data property?

If you agree, then we can close this issue as not completed.

[...]

PR #1160, opened 2024-09-17
Issue #1155 Annotations conform to style guide

### LINKED ISSUE #1155: Rewrite SKOS definitions and other annotations to follow gist style guide
This is an incremental task; you don't have to do all of them to submit a PR.

kstudzin:
This set of changes focuses on removing excess information from `skos:definition` and moving it to `skos:example` or `skos:scopeNote`. There is also some reformatting of all three annotations. However these changes do not attempt to make `skos:definition` full sentences where they are not. More changes will be added for this.

The definitions most significantly affected by removing examples are `gist:Behavior` and `gist:Content`. `gist:Behavior` maybe falls in the category "Occasionally a definition can hardly be understood at all without an example" of the style guide. My recommendation is that we try to make a more specific, full sentence definition, but we may determine that it is most clear to simply add the example back to the definition.

Our definition of `gist:Content`, in my opinion, was not a definition but rather a list of examples. I attempted to make a new definition, but would appreciate input on if this matches the existing usage. If there is other history on this definition and why we should include examples in the definition, that would also be helpful.

Closes #1155

### DIFF
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -1471,7 +1471,7 @@ gist:Person
 gist:PhysicalActionType
 	a owl:Class ;
 	rdfs:subClassOf gist:Category ;
-	skos:definition "A category indicating the type of an action in terms of its effect in the physical world."^^xsd:string ;
+	skos:definition "A category indicating the type of an action based on its effect in the physical world."^^xsd:string ;
 	skos:example "Lifting a garage door, turning off a valve, dropping cadmium rods."^^xsd:string ;
 	skos:prefLabel "Physical Action Type"^^xsd:string ;
 	.


ontologies/gistCore.ttl
```diff
@@ -2896,8 +2901,8 @@ gist:isAllocatedBy
 
 gist:isBasedOn
 	a owl:ObjectProperty ;
-	skos:definition "The Object is a foundation for, a starting point for, gave rise to or justifies the Subject"^^xsd:string ;
-	skos:example "A document is based on a document template. A metric computing the average income of a population is based on the metric for individual income."^^xsd:string ;
+	skos:definition "The object is a foundation for, a starting point for, gave rise to, or justifies the subject"^^xsd:string ;
```
rjyounes, line 2904:
```suggestion
	skos:definition "The object is a foundation for, a starting point for, gave rise to, or justifies the subject."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -3023,7 +3028,8 @@ gist:isRecognizedBy
 gist:isRecordedAt
 	a owl:DatatypeProperty ;
 	rdfs:subPropertyOf gist:atDateTime ;
-	skos:definition "Date that something was posted, not necessarily the date it occurred. Must be after the date of occurrence, but could be before or after the planned date. (Unusual, but I could record today that I expected to be paid last week.)"^^xsd:string ;
+	skos:definition "Date that something was posted, not necessarily the date it occurred. Must be after the date of occurrence, but could be before or after the planned date."^^xsd:string ;
```
rjyounes, line 3031:
```suggestion
	skos:definition "Date that something was posted (which is not necessarily the date it occurred). Must be after the date of occurrence, but could be before or after the planned date."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -3037,17 +3043,14 @@ gist:isRenderedOn
 gist:isSupersededBy
 	a owl:AnnotationProperty ;
 	skos:definition "Relates a deprecated term to a term that replaces it, which is either an exact (in the case of simple renaming) or approximate (in the case of renaming and some semantic change) semantic match."^^xsd:string ;
-	skos:example "gist:connectedTo gist:isSupersededBy gist:isConnectedTo (renaming with no semantic change)."^^xsd:string ;
+	skos:example "gist:connectedTo gist:isSupersededBy gist:isConnectedTo (renaming with no semantic change)"^^xsd:string ;
```
```suggestion
	skos:example "gist:connectedTo was superseded by gist:isConnectedTo (renaming with no semantic change)"^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -3145,8 +3148,9 @@ gist:occursIn
 gist:offers
 	a owl:ObjectProperty ;
 	skos:definition "Relates something to a thing that is being made available for acceptance or rejection."^^xsd:string ;
-	skos:example "An instance of gist:Offer offers a specific product at a particular price; a company offers an employee benefit; Honda offers vehicles for sale."^^xsd:string ;
+	skos:example "A company offers an employee benefit; Honda offers vehicles for sale."^^xsd:string ;
 	skos:prefLabel "offers"^^xsd:string ;
+	skos:scopeNote "An offer offers a specific product at a particular price."^^xsd:string ;
```
rjyounes, line 3153: This should stay as it is - it's an example, not a note about offers generally. The purpose of the example is to show that the predicate can have a `gist:Offer` as subject, as well as the thing making the offer.

ontologies/gistCore.ttl
```diff
@@ -3145,8 +3148,9 @@ gist:occursIn
 gist:offers
 	a owl:ObjectProperty ;
 	skos:definition "Relates something to a thing that is being made available for acceptance or rejection."^^xsd:string ;
```
rjyounes, line 3150:
```suggestion
	skos:definition "Relates something to a thing it makes available for acceptance or rejection."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -3282,7 +3287,8 @@ gist:prevents
 	a owl:ObjectProperty ;
 	rdfs:domain gist:Intention ;
 	rdfs:range gist:Behavior ;
-	skos:definition "The intention (say a law) is intended to prevent this kind of behavior (say jay-walking)"^^xsd:string ;
+	skos:definition "The intention prevents the behavior"^^xsd:string ;
```
rjyounes, line 3290:
```suggestion
	skos:definition "Relates an intention to behavior that it prevents or prohibits."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -3158,8 +3162,9 @@ gist:owns
 			gist:Person
 		) ;
 	] ;
-	skos:definition "Possessing and controlling.  Ultimate form of ownership is the right to destroy.  Long list of potential Range classes"^^xsd:string ;
+	skos:definition "Possessing and controlling. Ultimate form of ownership is the right to destroy."^^xsd:string ;
```
rjyounes, line 3165:
```suggestion
	skos:definition "Relates something to a thing it possesses and controls."^^xsd:string ;
	skos:scopeNote "The ultimate form of ownership is the right to destroy."^^xsd:string ;
```

docs/gistStyleGuide.md
```diff
@@ -224,7 +224,7 @@ Certain RDFS annotations are recommended where there is no SKOS equivalent.
...
-| `skos:example` | May be either a full sentence or a list. Use a final period only in the former case. E.g., "SSN, driver's license number, employee ID" or "NIH sponsors a research project." Lists with short items, such as the first example, can be delimited by either commas (include Oxford commas) or semi-colons; full-sentence examples should be semi-colon-delimited. |
+| `skos:example` | May be either a full sentence, a list, or sample values. Full sentence or list examples should end with a period, e.g., "SSN, driver's license number, employee ID." or "NIH sponsors a research project." Lists with short items, such as the first example, can be delimited by either commas (include Oxford commas) or semi-colons; full-sentence examples should be semi-colon-delimited. Sample value examples include `"application/sparql-results+xml"` for `gist:MediaType` and datetime values such as `'2021-06-01T00:00:00-6:00'^^xsd:dateTime` for `gist:actualEndDate`. |
```
rjyounes, line 227: I think I meant "single" rather than "sample," since they're all samples.

```suggestion
| `skos:example` | May be either a full sentence, a list, or a single value. Full sentence or list examples should end with a period, e.g., "SSN, driver's license number, employee ID." or "NIH sponsors a research project." Lists with short items, such as the first example, can be delimited by either commas (include Oxford commas) or semi-colons; full-sentence examples should be semi-colon-delimited. Single value examples include `"application/sparql-results+xml"` for `gist:MediaType` and datetime values such as `'2021-06-01T00:00:00-6:00'^^xsd:dateTime` for `gist:actualEndDate`. |
```

ontologies/gistCore.ttl
```diff
@@ -184,15 +185,17 @@ gist:Artifact
...
 	skos:prefLabel "Aspect"^^xsd:string ;
 	skos:scopeNote "Every aspect should be related to a broader aspect or to a unit group. For example, angle of incidence should be related to the broader concept of angle, which in turn is related to a unit group."^^xsd:string ;
 	.
 
 gist:Behavior
 	a owl:Class ;
 	rdfs:subClassOf gist:Category ;
-	skos:definition "A way of categorizing events.  E.g., differentiating drilling versus cutting."^^xsd:string ;
+	skos:definition "A way of categorizing events."^^xsd:string ;
+	skos:example: "A saw behaves such that it cuts while a drill behaves such that it drills."^^xsd:string ;
```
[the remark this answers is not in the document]
kstudzin: I was wondering if we changed the definition, to something along the lines of "The way in which something works or functions," we could improve the definition and leave the example as a `skos:example`. The Oxford definition of behavior is "the way in which a natural phenomenon or a machine works or functions." But perhaps this changes the definition too much?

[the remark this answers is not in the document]
uscholdm: This is very tricky. It's origin was in manufacturing, hence the drilling/cutting examples. I propose

- A category used to indicate the nature of what is happening during an event.

This can then be separated from the examples.

ontologies/gistCore.ttl
```diff
@@ -703,9 +709,9 @@ gist:GeoPoint
 			]
 		) ;
 	] ;
-	skos:definition "An individual point on the Earth's surface, identified by latitude, longitude and altitude. If altitude is missing, it is assumed to be at the Earth's surface.  However, altitude is measured from sea level.  these points are to the WGS-84 coordinate system using the GPS decimal lat/long"^^xsd:string ;
+	skos:definition "An individual point on the Earth's surface, identified by latitude, longitude and altitude. If altitude is missing, it is assumed to be at the Earth's surface. However, altitude is measured from sea level."^^xsd:string ;
```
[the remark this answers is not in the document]
kstudzin: I agree this definition could be improved because right now it says the point is on the Earth's surface and also if it has an altitude it might not be on the Earth's surface. 

Here is one re-writing attempt:
> An individual point in physical space identified by latitude, longitude and altitude. If altitude is missing, the point is assumed to be at the Earth's surface, and the altitude is the distance measured from sea level.

If we want to remove the word 'Earth' we could say something like:
> An individual point in physical space identified by latitude, longitude and altitude within a geographic coordinate system. If altitude is missing, the point is assumed to be at the planet's surface, and the altitude is the distance measured from the reference sphere.

There is a [NASA whitepaper](https://science.nasa.gov/wp-content/uploads/2024/01/luncoordwhitepaper-10-08.pdf) that discusses lunar coordinates and reference sphere. (Scope note: Earth's reference sphere is sea level.)

If we want to allow a geo point to be on any planet, we can no longer assume WGS 84 and may need an additional property to specify the coordinate system.

docs/gistStyleGuide.md
kstudzin: I added sample because I think there is a difference between the following types of examples:
```
gist:actualStartYear skos:example "'2021-01-01T00:00:00-6:00'^^xsd:dateTime"^^xsd:string
gist:allows skos:example  "A grant allows egress."^^xsd:string .
```

The first statement is a sample because you might find a statement with that value in actual data:
```
:_SomeResource gist:actualStartYear '2021-01-01T00:00:00-6:00'^^xsd:dateTime .
```

The second statement is an example but not a sample because you would never find the following in actual data:
```
:_SomeResource gist:allows "A grant allows egress." .
```
 
The difference is that a sample can be used character for character in the data.

I wanted to make this distinction for two reasons:
1. Samples don't need to end in a final period
```
# prefer
gist:actualStartYear skos:example "'2021-01-01T00:00:00-6:00'^^xsd:dateTime"^^xsd:string

# over
gist:actualStartYear skos:example "'2021-01-01T00:00:00-6:00'^^xsd:dateTime."^^xsd:string
```
Because a sample can be used character for character in data, it should not have an addition period that would not be in real data. All other examples should end in a final period. 

2. Term with both examples and samples, should have separate statements for each
```
# prefer
gist:actualStartYear
	skos:example
		"'2021-01-01T00:00:00-6:00'^^xsd:dateTime"^^xsd:string ,
		"The tenure of the current chairman of the board began in 2021."^^xsd:string
		;
# over
gist:actualStartYear
	skos:example
		"'2021-01-01T00:00:00-6:00'^^xsd:dateTime; the tenure of the current chairman of the board began in 2021."^^xsd:string
		;
```
I think lists should maintain parallel structure to make them easy to read and understand. Putting examples and samples together felt hard to read and disorienting.

But perhaps this adds unnecessary complexity

ontologies/gistCore.ttl
```diff
@@ -552,15 +555,15 @@ gist:Determination
 	a owl:Class ;
 	rdfs:subClassOf gist:Event ;
 	skos:definition "An event whose purpose is to establish a specific result, value, or outcome, usually by research, measuring, evaluating, or calculating."^^xsd:string ;
-	skos:example "Measuring the sulphur content of crude oil. Evaluating a loan application for approval. Estimating the price of gas for the next three months. Determining whether and by how much an interest rate should change. Classifying something."^^xsd:string ;
+	skos:example "Measuring the sulphur content of crude oil; evaluating a loan application for approval; estimating the price of gas for the next three months; determining whether and by how much an interest rate should change"^^xsd:string ;
```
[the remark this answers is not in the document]
kstudzin: It is a good example and I wouldn't necessarily think of it right away. I think I removed it because it's very general while the rest of the list is very specific. Could we add a more specific example about classification? Maybe something like "classifying a purchase into a budget category?" If not, I can add back as is

ontologies/gistCore.ttl
kstudzin: Sounds good. The reason I changed it was because we try to avoid ontology terms in annotations and there are other cases of describing what gist terms are connected in scope notes (but I think those are more definitive use cases than this). Should I restore the original wording or move the new wording back to example?

ontologies/gistCore.ttl
```diff
@@ -3062,41 +3069,41 @@ gist:latitude
...
 	a owl:AnnotationProperty ;
-	skos:definition "An annotation for providing the licensing on this or derivative ontologies"^^xsd:string ;
+	skos:definition "An annotation for providing the licensing on this or derivative ontologies."^^xsd:string ;
 	skos:prefLabel "license"^^xsd:string ;
 	.
 
 gist:links
 	a owl:ObjectProperty ;
-	skos:definition "Relates a NetworkLink to a NetworkNode that it connects to another node. Used when the connections are undirected, or the direction is not known."^^xsd:string ;
+	skos:definition "Relates a network link to a network node that it connects to another node. Used when the connections are undirected, or the direction is not known."^^xsd:string ;
```
uscholdm, line 3084:
```suggestion
skos:definition "Relates a network link to one of the two network nodes that it connects. Used when the connections are undirected, or the direction is not known."^^xsd:string ;

File: ontologies/gistCore.ttl [hunk 13]
```diff
@@ -3158,8 +3166,12 @@ gist:owns
 			gist:Person
 		) ;
 	] ;
-	skos:definition "Possessing and controlling.  Ultimate form of ownership is the right to destroy.  Long list of potential Range classes"^^xsd:string ;
+	skos:definition "Relates something to a thing it possesses and controls."^^xsd:string ;
 	skos:prefLabel "owns"^^xsd:string ;
+	skos:scopeNote
+		"The range of this property is the union of many potential classes."^^xsd:string ,
+		"The ultimate form of ownership is the right to destroy."^^xsd:string
+		;
```
uscholdm, line 3174: Presumably only persons or parties can own something, is that the idea? Be clear about whether this is intended to be legal ownership, what other kind is there that a typical enterprise would care about? Perhaps change the definition to:
- `Relates a party to something that it possesses and controls.`

uscholdm (changes requested):
Overall this is excellent, a big job and a painstaking one.  I did make a bunch of suggestions.  

Although this exercise was specifically to align definition style, I often gave in to the temptation to fix a bad definition, which might instead have been a separate issue warranting a group.

I made changes for properties to style starting with 'relates'. But there were too many to attempt a fix for all of them.

ontologies/gistCore.ttl
```diff
@@ -2218,7 +2235,7 @@ gist:containedText
 
 gist:contributesTo
 	a owl:ObjectProperty ;
-	skos:definition "The parts of a system contribute to the goal/ function of the whole system"^^xsd:string ;
+	skos:definition "The parts of a system contribute to the goal/ function of the whole system."^^xsd:string ;
```
dylan-sa, line 2238:
```suggestion
	skos:definition "The parts of a system contribute to the goal/function of the whole system."^^xsd:string ;
```

dylan-sa (changes requested):
Agree with @uscholdm--this is great. A lot of much needed work. 👍

I've added some things that caught my eye.

ontologies/gistCore.ttl
```diff
@@ -535,7 +538,7 @@ gist:CountryGovernment
 			]
 		) ;
 	] ;
-	skos:definition "A Government Organization which asserts both sovereignty (i.e., it is not governed by some other government organization) and governance over an entity generally recognized as a 'country'."^^xsd:string ;
+	skos:definition "A government organization which asserts both sovereignty (i.e., it is not governed by some other government organization) and governance over an entity generally recognized as a country."^^xsd:string ;
```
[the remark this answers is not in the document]
uscholdm: Good idea.

ontologies/gistCore.ttl
```diff
@@ -703,9 +709,9 @@ gist:GeoPoint
 			]
 		) ;
 	] ;
-	skos:definition "An individual point on the Earth's surface, identified by latitude, longitude and altitude. If altitude is missing, it is assumed to be at the Earth's surface.  However, altitude is measured from sea level.  these points are to the WGS-84 coordinate system using the GPS decimal lat/long"^^xsd:string ;
+	skos:definition "An individual point on the Earth's surface, identified by latitude, longitude, and altitude. If altitude is missing, it is assumed to be at the Earth's surface. However, altitude is measured from sea level."^^xsd:string ;
```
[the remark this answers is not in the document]
uscholdm: I agree.

ontologies/gistCore.ttl
```diff
@@ -1016,7 +1029,7 @@ gist:LivingThing
 			]
 		) ;
 	] ;
-	skos:definition "Something that is currently, or at some point in time was, alive."^^xsd:string ;
+	skos:definition "Something that is currently, or at some point in time was or will be, alive."^^xsd:string ;
```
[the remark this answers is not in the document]
kstudzin: The `gist:birthDate` definition says that 'The date some living thing was or will be born' so I thought it only makes sense if there are living things that are not yet born

ontologies/gistCore.ttl
```diff
@@ -570,7 +573,7 @@ gist:ElectronicAddress
 	owl:disjointWith gist:PhysicalAddress ;
 	skos:altLabel "Virtual Address"^^xsd:string ;
 	skos:definition "An address referring to a locatable virtual place that does not physically exist but is made by software or electronics to appear to do so."^^xsd:string ;
-	skos:example "A path to a file in a file system, a website URL, an IP address, an email address, a mobile or landline telephone number."^^xsd:string ;
+	skos:example "A path to a file in a file system; a website URL; an IP address; an email address; a mobile or landline telephone number."^^xsd:string ;
```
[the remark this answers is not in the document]
kstudzin: The style guide says that short examples should use commas and long examples should use semi colons. I judged this to be long examples because 'A path to a file in a file system' and, to a lesser degree, 'a mobile or landline telephone number' are more complex than the short '<article> <optional adjective> <noun>' structure. I can change the first example to 'A filesystem path,' which is a better parallel to the other examples, and use commas

ontologies/gistCore.ttl
```diff
@@ -3062,41 +3069,41 @@ gist:latitude
...
 	rdfs:range xsd:double ;
-	skos:definition "Degrees above or below equator"^^xsd:string ;
+	skos:definition "Degrees above or below equator."^^xsd:string ;
 	skos:prefLabel "latitude"^^xsd:string ;
 	.
 
 gist:license
 	a owl:AnnotationProperty ;
-	skos:definition "An annotation for providing the licensing on this or derivative ontologies"^^xsd:string ;
+	skos:definition "An annotation for providing the licensing on this or derivative ontologies."^^xsd:string ;
```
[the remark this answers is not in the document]
kstudzin: This seems related to #682 and #977 

Either way, we could change the definition to be more generic: "Relates something to its license." And we could add ontologies and software as examples. Are there other examples we should include?

ontologies/gistCore.ttl
```diff
@@ -703,9 +709,9 @@ gist:GeoPoint
 			]
 		) ;
 	] ;
-	skos:definition "An individual point on the Earth's surface, identified by latitude, longitude and altitude. If altitude is missing, it is assumed to be at the Earth's surface.  However, altitude is measured from sea level.  these points are to the WGS-84 coordinate system using the GPS decimal lat/long"^^xsd:string ;
+	skos:definition "An individual point on the Earth's surface, identified by latitude, longitude, and altitude."^^xsd:string ;
 	skos:prefLabel "Geo Point"^^xsd:string ;
-	skos:scopeNote "Assume coordinate system used by Google (WGS 84 Web Mercator)."^^xsd:string ;
+	skos:scopeNote "Assume coordinate system used by Google (WGS 84 Web Mercator using the GPS decimal lat/long)."^^xsd:string ;
```
uscholdm, line 714: I think too much was removed. Feel free to reorder the sentences or put into two separate scope notes. I cannot decide on a preference.

```suggestion
skos:scopeNote "If altitude is missing, it is assumed to be at the Earth's surface. However, altitude is measured from sea level. Assume coordinate system used by Google (WGS 84 Web Mercator using the GPS decimal lat/long)."^^xsd:string ;

If altitude is missing, it is assumed to be at the Earth's surface. However, altitude is measured from sea level.

File: ontologies/gistCore.ttl (hunk 19 above)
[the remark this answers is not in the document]
Comment (uscholdm) [reply ontologies/gistCore.ttl:3078]: I like `Relates something to its license.` .Let's make that change for now and save other changes for the cited issues.

File: ontologies/gistCore.ttl [hunk 21]
```diff
@@ -919,10 +928,13 @@ gist:IntellectualProperty
 		gist:PhysicalSubstance ,
 		gist:UnitOfMeasure
 		;
-	skos:definition "A work, invention or concept, independent of its being expressed in text, audio, video, image, or live performance.  IP can also be tacit knowledge, know-how, or skill. Also includes Brands."^^xsd:string ;
-	skos:example "The Old Man and The Sea; the Page Rank algorithm; Coca Cola"^^xsd:string ;
+	skos:definition "A work, invention or concept existing as tacit knowledge, know-how, or skill or expressed in text, audio, video, image, or live performance."^^xsd:string ;
```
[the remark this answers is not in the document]
kstudzin: I agree the idea that IP doesn't need to be expressed is important. I tried to keep that conceptually in the definition in two ways:

    Emphasizing that IP is often tacit knowledge, and therefore not expressed, by moving that phrase early in the definition
    Providing options, either A or B. In this definition, IP is independent of being expressed because it could only exist tacitly.

What are your thoughts on if this approach?

ontologies/gistCore.ttl
```diff
@@ -1753,9 +1776,9 @@ gist:SubCountryGovernment
 	skos:definition "A government of a governed geo-region other than a country, which is under the direct or indirect control of a country government."^^xsd:string ;
 	skos:prefLabel "Sub-Country Government"^^xsd:string ;
 	skos:scopeNote
-		"Across the world, there are a variety of types of subsections of a country and the governments thereof (as well as different terms, like 'province' and 'state', which refer to essentially the same type of thing). We should not automatically assume 'state', 'county', and 'city'.  It is more future-proof just to mint the instances using the generic SubCountryGovernment and, where greater specificity is needed, define categories or subclasses."^^xsd:string ,
+		"Across the world there are a variety of types of subsections of a country and the governments thereof (as well as different terms, like 'province' and 'state,' which refer to essentially the same type of thing). We should not automatically assume 'state,' 'county,' and 'city.' It is more future-proof just to mint the instances using the generic sub-country government and, where greater specificity is needed, define categories or subclasses."^^xsd:string ,
```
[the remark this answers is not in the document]
kstudzin: MLA Handbook and Chicago Manual of Style both recommend commas inside quotation marks. MLA posted [this](https://style.mla.org/punctuation-and-quotation-marks/) summary of opinions. They note that British style uses single quotes with commas outside.

ontologies/gistCore.ttl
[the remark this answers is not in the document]
uscholdm: Now I get it, good explanation, we can go with this.

ontologies/gistCore.ttl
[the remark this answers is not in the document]
uscholdm: What a shame, let's go to the UK! :-)

ontologies/gistCore.ttl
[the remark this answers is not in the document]
kstudzin: @uscholdm Do you have thoughts on either of these definitions?

ontologies/gistCore.ttl
kstudzin: Yes it does look like too much was removed. There's a larger discussion [here](https://github.com/semanticarts/gist/pull/1160#discussion_r1774079970) that I also just tagged you in

ontologies/gistCore.ttl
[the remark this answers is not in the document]
uscholdm: > An individual point in physical space identified by latitude, longitude and altitude. If altitude is missing, the point is assumed to be at the Earth's surface, and the altitude is the distance measured from sea level.

I like this definition. I don't think we need to worry about coordinate systems outside of earth.

docs/gistStyleGuide.md
uscholdm: The distinction you make is real enough, but is nuanced - it might cause more confusion than clarity.

ontologies/gistCore.ttl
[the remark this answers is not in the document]
rjyounes: @Jamie-SA I don't like it either, it doesn't make sense to me, but it's not a "rule," it's a rule. There is no controversy among grammarians about this.

ontologies/gistCore.ttl
```diff
@@ -1552,24 +1569,25 @@ gist:PhysicalSubstance
 		;
 	owl:disjointWith gist:UnitOfMeasure ;
 	skos:definition "An undifferentiated amount of physical material which, when subdivided, results in each part being indistinguishable in nature from the whole and from every other part."^^xsd:string ;
-	skos:example "An amount of water, penicillin, sand, gold: an actual piece of gold, not the concept of gold."^^xsd:string ;
+	skos:example "An amount of water, penicillin, sand, or gold."^^xsd:string ;
```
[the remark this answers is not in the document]
kstudzin: I'm confused about gold the element as a negative example. Is it a negative example because we can't purify it enough and therefore doesn't exist? In theory, if there were a hunk of elemental gold and it split into two hunks, they would be indistinguishable from one another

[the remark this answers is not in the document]
uscholdm: > I initially removed it because I wasn't really sure what we meant by 'the concept of gold.' I'm also confused about gold the element as a negative example. Is it a negative example because we can't purify it enough and therefore doesn't exist? In theory, if there were a hunk of elemental gold and it split into two hunks, they would be indistinguishable from one another

No, it is a negative example because gold the concept is a type, a category, a kind of thing - it has no mass or volume.  Analogously, my physical iPhone SE2020 does in fact weigh 300g;  you can ask the question: what kind of phone do I have? All phones of that kind should weigh close to 300g, but the model itself is not a physical thing that has mass or volume.  The model has a spec saying each phone of that model should weigh 300g

ontologies/gistCore.ttl
```diff
@@ -641,14 +643,15 @@ gist:Equipment
 			]
 		) ;
 	] ;
-	skos:definition "Tangible property other than land or buildings.  Any kind of equipment, could be machine, router, car etc."^^xsd:string ;
+	skos:definition "Tangible property other than land or buildings."^^xsd:string ;
+	skos:example "A machine, a router, a car."^^xsd:string ;
```
stevenchalem:
```suggestion
	skos:example "A machine, a router, a car, a chair, a necklace."^^xsd:string ;
```
Since the definition is broader that when we usually call "equipment," I think it's helpful to include examples that might not normally be thought of as "equipment."

ontologies/gistCore.ttl
```diff
@@ -794,7 +799,7 @@ gist:GeoRegion
 	skos:definition "A bounded region (or set of regions) on the surface of the Earth."^^xsd:string ;
 	skos:example "The bounded shape that defines the region occupied by Crater Lake; the bounded area known as the contiguous USA."^^xsd:string ;
 	skos:prefLabel "Geo Region"^^xsd:string ;
-	skos:scopeNote "A GeoRegion could be non-contiguous; e.g. the region governed by the USA is the region governed by the lower 48 states plus that of Alaska and Hawaii.  Child classes in lower ontologies can make this distinction."^^xsd:string ;
+	skos:scopeNote "A geographical region could be non-contiguous; e.g., the region governed by the US federal government is the contiguous area of the lower 48 states plus Alaska and Hawaii. Child classes in lower ontologies can make this distinction."^^xsd:string ;
```
stevenchalem:
```suggestion
	skos:scopeNote "A geographical region could be non-contiguous; e.g., the region governed by the US federal government is the contiguous area of the lower 48 states plus Alaska, Hawaii, and overseas territories. Child classes in lower ontologies can make this distinction."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -838,7 +843,7 @@ gist:GeoSegment
 			]
 		) ;
 	] ;
-	skos:definition "A single portion of a GeoRoute consisting of a pair of consecutive GeoPoints which belong to it."^^xsd:string ;
+	skos:definition "A single portion of a geographical route which has been divided (i.e., segmented)."^^xsd:string ;
```
stevenchalem:
```suggestion
	skos:definition "A single portion of a GeoRoute consisting of a pair of consecutive GeoPoints which belong to it."^^xsd:string ;
```
Since this class is being removed in another PR, I think that leaving it unchanged here will avoid a merge conflict.

ontologies/gistCore.ttl
```diff
@@ -2714,39 +2726,38 @@ gist:hasGiver
...
 	.
 
 gist:hasMultiplier
 	a owl:ObjectProperty ;
 	skos:definition "Relates a unit of measure to another unit of measure that is a factor, or relates an aspect to another aspect that is a factor."^^xsd:string ;
-	skos:example
-		"Miles per hour has miles as a multiplier and hour as a divisor."^^xsd:string ,
-		"Speed has distance as a multiplier and duration as a divisor."^^xsd:string
-		;
+	skos:example "Miles per hour has miles as a multiplier and hour as a divisor; speed has distance as a multiplier and duration as a divisor."^^xsd:string ;
```
stevenchalem:
```suggestion
	skos:example "Miles per hour has miles as a multiplier and hours as a divisor; speed has distance as a multiplier and duration as a divisor."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -2779,9 +2790,10 @@ gist:hasParty
 			gist:Person
 		) ;
 	] ;
-	skos:definition "The people or organizations participating in an event, agreement or obligation"^^xsd:string ;
-	skos:example "For loan agreements, one might create hasLender and hasBorrower as subproperties of hasParty."^^xsd:string ;
+	skos:definition "Relates an event, agreement, or obligation to the participating people or organizations."^^xsd:string ;
```
stevenchalem:
```suggestion
	skos:definition "Relates an event, agreement, or obligation to a participating person or organization."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -2929,14 +2936,14 @@ gist:isAssignmentTo
...
-	skos:example "A document is based on a document template. A metric computing the average income of a population is based on the metric for individual income."^^xsd:string ;
+	skos:definition "The object is a foundation for, a starting point for, gave rise to, or justifies the subject."^^xsd:string ;
+	skos:example "A document is based on a document template; a metric computing the average income of a population is based on the metric for individual income."^^xsd:string ;
 	skos:prefLabel "is based on"^^xsd:string ;
 	.
 
 gist:isCategorizedBy
 	a owl:ObjectProperty ;
-	skos:definition "Points to a taxonomy item or other less formally defined class."^^xsd:string ;
+	skos:definition "Relates something to a taxonomy term or other less formally defined class."^^xsd:string ;
```
stevenchalem:
```suggestion
	skos:definition "Relates something to a taxonomy term or other less formally defined thing that classifies or characterizes it."^^xsd:string ;
```
Avoiding the potentially confusing "class." Also saying something about the nature of the relationship.

stevenchalem (dismissed):
On line 586, the skos:prefLabel for gist:DegreeOfCommitment is not in title case

ontologies/gistCore.ttl
```diff
@@ -1461,7 +1471,8 @@ gist:Person
 gist:PhysicalActionType
 	a owl:Class ;
 	rdfs:subClassOf gist:Category ;
-	skos:definition "The effects to be realized in the real world, such as lifting a garage door, turning off a valve, dropping cadmium rods, etc."^^xsd:string ;
+	skos:definition "A category indicating the real-world effects of an action."^^xsd:string ;
```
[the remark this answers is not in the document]
rjyounes: Not crazy about your suggestion because it doesn't indicate the categorization is in terms of its real-world effect. What do you think about "A category indicating the type of an action in terms of its effect in the physical world."?

ontologies/gistCore.ttl
rjyounes: How about just "classification" rather than class?

stevenchalem: I like it.

ontologies/gistCore.ttl
rjyounes: Hm, this is an odd one. I think the definition should be tightened up, rather than extending the examples. It should also say man-made; this used to be a subclass of `Artifact`, which was removed.

Issue #1165, opened 2024-10-07
GeoSegment is unnecessary


rjyounes:
A `GeoSegment` is just the special case of a `GeoRoute` with two points. If it were to subclass `GeoRoute`, it makes the representation of a `GeoRoute` messy because it uses a different mechanism for ordering (`comesFromPlace`, `goesToPlace` rather than an ordered collection). The triples using those predicates will be inferred between every pair of consecutive points on the route.

Either we should keep the class and redefine it as a `GeoRoute` with two members, dropping the predicates from the definition, or deprecate/delete it.

kstudzin:
I think we are correcting `GeoRegion` to `GeoRoute` in #1160. 

I think the current definition, including the formal definition, leave it unclear if `GeoSegment` is a curve or a line. I had the impression it was a line, but looking at it a second time it wasn't 100% clear it can't be a curve. 

For context, this came up due to the gist CCO mapping. CCO has a `GeospatialLineString` class, which is similar to `GeoRoute` and has a `GeospatialLine` subclass.  I was trying to understand if `GeoSegment` and `GeospatialLine` were essentially equivalent

rjyounes:
GeoRoute and GeoSegment cannot contain curves, else the ordered collection of points or the `comesFromPlace` and `goesToPlace` values, respectively, would not be deterministic.

kstudzin:
The way I think about it is that now we assume the function connecting the two points is a line. I was imagining overriding that with a more complex function. I think we should add the word "line" to the definition: "An ordered set of GeoPoints that defines a path of line segments from starting point to ending point."

PR #1175, opened 2024-12-03
fix definition of Geo segment

### LINKED ISSUE #1134: fix definition of GeoSegment
The definition of GeoSegment appears to have a typo. It says a GeoSegment is a single portion of a _GeoRegion_ instead of being a single portion of a _GeoRoute_.

Current definition of GeoSegment: "A single portion of a GeoRegion which has been divided (i.e., segmented)."

Proposed: "A single portion of a GeoRoute consisting of consecutive GeoPoints belonging to the GeoRoute."

Proposed addition to GeoRoute: "Each consecutive pair of GeoPoints in a GeoRoute is a GeoSegment."

philblackwood:
closes #1134

### DIFF
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -778,6 +778,7 @@ gist:GeoRoute
 	] ;
 	skos:definition "An ordered set of GeoPoints that defines a path from starting point to ending point."^^xsd:string ;
 	skos:prefLabel "Geo Route"^^xsd:string ;
+	skos:scopeNote "Each pair of consecutive GeoPoints in a GeoRoute is a GeoSegment."^^xsd:string ;
 	.
 
 gist:GeoSegment
@@ -800,7 +801,7 @@ gist:GeoSegment
 			]
 		) ;
 	] ;
-	skos:definition "A single portion of a GeoRegion which has been divided (i.e., segmented)."^^xsd:string ;
+	skos:definition "A single portion of a GeoRoute consisting of a pair of consecutive GeoPoints which belong to it."^^xsd:string ;
 	skos:prefLabel "Geo Segment"^^xsd:string ;
 	.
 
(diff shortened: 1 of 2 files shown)


ontologies/gistCore.ttl
```diff
@@ -778,6 +778,7 @@ gist:GeoRoute
 	] ;
 	skos:definition "An ordered set of GeoPoints that defines a path from starting point to ending point."^^xsd:string ;
 	skos:prefLabel "Geo Route"^^xsd:string ;
+	skos:scopeNote "Each pair of consecutive GeoPoints in a GeoRoute is a GeoSegment."^^xsd:string ;
```
uscholdm, line 781: Add to scope note tha A single GeoSegment also a GeoRoute. If it's not now, then probably it should be a subclass.

uscholdm (approved):

ontologies/gistCore.ttl
rjyounes: I agree that GeoSegment is a subclass of GeoRoute with two members, but we are using two different mechanisms to define them. In the case of GeoSegment, we are using comesFrom/goesToPlace, whereas we are using the ordered collection mechanism for GeoRoutes. They don't need to be mutually exclusive, but the way they are currently defined they don't harmonize well. This is considered in issue #1165.

rjyounes (approved):

MichaelSullivanArchitect:
seems to me the definition should possibly change to something like "An ordered set of GeoSegments..."

uscholdm:
> seems to me the definition should possibly change to something like "An ordered set of GeoSegments..."

I think the definition is fine as it is. It may be the case that as a consequence, it is also "An ordered set of GeoSegments...". If so, it could be added as a scope note, if we think it would be helpful.

PR #1180, opened 2025-01-24
fix geo route and remove geo segment

### LINKED ISSUE #1165: GeoSegment is unnecessary

philblackwood:
Remove GeoSegment, change GeoRoute class equivalence to say it is an ordered collection whose members are GeoPoints (old class equivalence incorrectly said it was an ordered collection with GeoSegments as parts).
Closes #1165

### DIFF
--- a/docs/release_notes/issue1165-remove-geosegment.md
+++ b/docs/release_notes/issue1165-remove-geosegment.md
@@ -1,0 +1,1 @@
+Change the class equivalence statement for GeoRoute to say it is an ordered collection whose members are geo points. Also, GeoSegment is not necessary because it is just a geo route with two geo points. Therefore the class GeoSegment is being removed from gist.
--- a/docs/release_notes/issue1165-remove-geosegment.md
+++ b/docs/release_notes/issue1165-remove-geosegment.md
@@ -1,1 +1,7 @@
-Change the class equivalence statement for GeoRoute to say it is an ordered collection whose members are geo points. Also, GeoSegment is not necessary because it is just a geo route with two geo points. Therefore the class GeoSegment is being removed from gist.
+### Minor Updates
+
+- Deprecate `gist:GeoSegment`. Issue [#1165](https://github.com/semanticarts/gist/issues/1165).
+
+### Patch Updates
+
+- Correct errors in property restriction on `gist:GeoRoute`. Issue [#1165](https://github.com/semanticarts/gist/issues/1165).
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -802,7 +802,7 @@ gist:GeoSegment
 		) ;
 	] ;
 	skos:definition "A single portion of a GeoRoute consisting of a pair of consecutive GeoPoints which belong to it."^^xsd:string ;
-	skos:editorialNote "See guidance on removing this term in the next major release at #1181."^^xsd:string ;
+	skos:editorialNote "See guidance on removing this term in the next major release at #1182."^^xsd:string ;
 	skos:prefLabel "Geo Segment"^^xsd:string ;
 	skos:scopeNote "A GeoSegment is also a GeoRoute (a short one)."^^xsd:string ;
 	.
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -770,15 +770,14 @@ gist:GeoRoute
 			[
 				a owl:Restriction ;
 				owl:onProperty [
-					owl:inverseOf gist:isDirectPartOf ;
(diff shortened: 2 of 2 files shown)


uscholdm (changes requested):
MIssing release notes.  We also talked about having a note that one could define a segment as a route with two points, but I think its too obvoius to say it. Though you might point it out in the release notes.

dylan-sa:
Are we planning this one for 13.1 or 14.0? If the former, we should deprecate `GeoSegment` rather than delete.

rjyounes:
This should be for `14.0.0.`. Sorry for the oversight. We can't rewrite restrictions in a minor release.

rjyounes:
> Though you might point it out in the release notes.

@uscholdm Our release notes are pretty bare bones, without justification or explanation. We link to the issue for that.

docs/release_notes/issue1165-remove-geosegment.md
```diff
@@ -0,0 +1 @@
+Change the class equivalence statement for GeoRoute to say it is an ordered collection whose members are geo points. Also, GeoSegment is not necessary because it is just a geo route with two geo points. Therefore the class GeoSegment is being removed from gist.
```
rjyounes, line 1: See Guidance on release notes in [Contributing.md](https://github.com/semanticarts/gist/blob/develop/docs/Contributing.md#release-notes). Here's what I propose for this one:

```
### Minor Updates

- Deprecate `gist:GeoSegment`. Issue [#1165](https://github.com/semanticarts/gist/issues/1165).

### Patch Updates

- Correct errors in property restriction on `gist:GeoRoute`. Issue [#1165](https://github.com/semanticarts/gist/issues/1165).
```

ontologies/gistCore.ttl
```diff
@@ -770,40 +770,14 @@ gist:GeoRoute
...
 			]
 		) ;
 	] ;
-	skos:definition "An ordered set of GeoPoints that defines a path from starting point to ending point."^^xsd:string ;
+	skos:definition "An ordered set of GeoPoints that defines a path from a starting point to an ending point."^^xsd:string ;
 	skos:prefLabel "Geo Route"^^xsd:string ;
-	skos:scopeNote "Each pair of consecutive GeoPoints in a GeoRoute is a GeoSegment."^^xsd:string ;
-	.
-
-gist:GeoSegment
```
rjyounes, line 784: This should be deprecated, with an editorial note and addition of a new issue for `14.0.0`. See section "How to Deprecate" in [Deprecation and Deletion Policy](https://github.com/semanticarts/gist/blob/develop/docs/DeprecationAndDeletionPolicy.md).

rjyounes (changes requested):

philblackwood:
Sorry for the confusion. From the guidelines: "If the term slated for removal is referenced by axioms in other gist terms, the term cannot be deprecated in a minor release. "

Since GeoSegment is referenced in the class equivalence definition of GeoRoute, it looks like we should make this change in the next major release, with no deprecation near-term.

Can we just move this to being part of gist 14?

ontologies/gistCore.ttl
philblackwood: OK. Issue #1181 for gist 14 will remove GeoSegment. I deprecated GeoSegment and added a note referring to issue #1181 to remove GeoSegment in gist 14. Also corrected the class equivalence statement for GeoRoute so it uses isMemberOf and GeoPoint.

I think this is ready now. Thanks for the help.

ontologies/gistCore.ttl
```diff
@@ -802,6 +802,7 @@ gist:GeoSegment
 		) ;
 	] ;
 	skos:definition "A single portion of a GeoRoute consisting of a pair of consecutive GeoPoints which belong to it."^^xsd:string ;
+	skos:editorialNote "See guidance on removing this term in the next major release at #1181."^^xsd:string ;
```
rjyounes, line 805:
```suggestion
	skos:editorialNote "See guidance on removing this term in the next major release at #1182."^^xsd:string ;
```

rjyounes (approved):
Made one small change to editorial note: I made #1181 an umbrella issue for deprecated concepts and added a new issue #1182 for `GeoSegment`. (Should have done the other way around.)

PR #1231, opened 2025-06-12
Update definition and scope note of gist:Organization

### LINKED ISSUE #1172: Definition of Organization is circular
"A generic organization that can be formal or informal, legal or non-legal. It can have members, or not" says nothing about what it means to be an organization.  For example it does not tell me whether a Sole Trader (inherently a single person) registered as a LLC counts as an :Organization. 
Nor does it say how organized it has to be. I presume a group of people in line for a ticket would not count. But what about those people attending a single lecture? What if they register and attend the same class for a whole term? How about the group of people who happen to have the same person buy a lottery ticket for them each week? What if they agree with that person that if anyone of them wins the prize will be shared among them all? At any point have these become organizations as opposed to a set of individuals dealing with a common service provider?

rjyounes:
Closes #1172

### DIFF
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -1418,10 +1418,10 @@ gist:Organization
 		gist:SchemaMetaData ,
 		gist:UnitOfMeasure
 		;
-	skos:definition "A generic organization that can be formal or informal, legal or non-legal. It can have members, or not."^^xsd:string ;
+	skos:definition "An organized body of individuals with a particular purpose, often with a structured approach and defined roles."^^xsd:string ;
 	skos:example "Legal entities like companies; non-legal entities like clubs, committees, or departments."^^xsd:string ;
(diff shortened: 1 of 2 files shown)


ontologies/gistCore.ttl
```diff
@@ -1418,10 +1418,10 @@ gist:Organization
 		gist:SchemaMetaData ,
 		gist:UnitOfMeasure
 		;
-	skos:definition "A generic organization that can be formal or informal, legal or non-legal. It can have members, or not."^^xsd:string ;
+	skos:definition "An organized body of individuals with a particular purpose, often with a structured approach and defined roles."^^xsd:string ;
```
uscholdm, line 1421:
```suggestion
skos:definition "A structured entity formed to achieve specific goals, typically involving members with defined roles."^^xsd:string ;
skos:scopeNote "Not all organizations have members, e.g. shell companies."^^xsd:string ;

marksem (approved):

PR #1248, opened 2025-07-25
Replaced produces with isProducedBy

### LINKED ISSUE #1163: Use isProducedBy and delete produces
Based on cardinality and child-to-parent criteria.

coltonglasgow:
Closes #1163

### DIFF
--- a/docs/release_notes/issue1163-produces-to-isProducedBy
+++ b/docs/release_notes/issue1163-produces-to-isProducedBy
@@ -1,3 +1,0 @@
-# Major Updates
-
-- No longer using `gist:produces`. It has been replaced by its logical inverse: `gist:isProducedBy`. Issue [1163](https://github.com/semanticarts/gist/issues/1163).
--- a/docs/release_notes/issue1163-produces-to-isProducedBy
+++ b/docs/release_notes/issue1163-produces-to-isProducedBy
@@ -1,0 +1,3 @@
+# Major Updates
+
+- Replaced `gist:produces` with `gist:isProducedBy`. Issue [1163](https://github.com/semanticarts/gist/issues/1163).
--- a/docs/release_notes/issue1163-produces-to-isProducedBy
+++ b/docs/release_notes/issue1163-produces-to-isProducedBy
@@ -1,3 +1,3 @@
 # Major Updates
 
-- Replaced `gist:produces` with `gist:isProducedBy`. Issue [1163](https://github.com/semanticarts/gist/issues/1163).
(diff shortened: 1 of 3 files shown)


dylan-sa (approved):
Approving w/ optional suggestion.

docs/release_notes/issue1163-produces-to-isProducedBy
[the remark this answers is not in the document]
coltonglasgow: Good catch. I have applied this suggestion.

docs/release_notes/issue1163-produces-to-isProducedBy.md
```diff
@@ -0,0 +1,3 @@
+# Major Updates
+
+- No longer using `gist:produces`. It has been replaced by its logical inverse: `gist:isProducedBy`. Issue [1163](https://github.com/semanticarts/gist/issues/1163).
```
rjyounes, line 3:
```suggestion
- Replaced `gist:produces` with its logical inverse,`gist:isProducedBy`. Issue [1163](https://github.com/semanticarts/gist/issues/1163).
```

ontologies/gistCore.ttl
```diff
@@ -3098,6 +3100,13 @@ gist:isPartOf
 		;
 	.
 
+gist:isProducedBy
+	a owl:ObjectProperty ;
+	skos:definition "The relation from something produced to the thing producing it."^^xsd:string ;
+	skos:example "A measurement is produced by a sensor."^^xsd:string ;
```
rjyounes, line 3106: Can you find another verb so it's not circular? E.g., "create." I would keep the previous example and also maybe add one with a human agent.

rjyounes (changes requested):

docs/release_notes/issue1163-produces-to-isProducedBy.md
coltonglasgow: @rjyounes, I made a slight change to your suggestion (seen below). I added a space after the comma. 

```suggestion
- Replaced `gist:produces` with its logical inverse, `gist:isProducedBy`. Issue [1163](https://github.com/semanticarts/gist/issues/1163).
```

ontologies/gistCore.ttl
marksem: I agree that I'd like the definition to be less circular, where we don't use the word we're defining in the definition.  
Some ideas:  'Relates something to the thing that created, composed, or caused it to have existence.'

coltonglasgow: I have updated the definition to the suggestion from above. I have also added back the original example (but reversed) and an example with a human.

PR #1254, opened 2025-08-12
issue 1234, update scope note on gist:description

### LINKED ISSUE #1234: Add scope note to gist:description
Here is the current definition of gist:description

```
gist:description
	a owl:DatatypeProperty ;
	rdfs:range xsd:string ;
	skos:definition "A statement about someone or something's attributes or characteristics."^^xsd:string ;
	skos:example "The Empire State Building is a 102-story Art Deco skyscraper in Midtown Manhattan in New York City, United States. It was designed by Shreve, Lamb & Harmon and built from 1930 to 1931."^^xsd:string ;
	skos:prefLabel "description"^^xsd:string ;
	skos:scopeNote "This property is used to provide a description of an individual in greater detail than a label."^^xsd:string ;
	.
```

I propose adding  (or replacing existing) a scope note as follows:
```
skos:scopeNote "This property is to be used to describe data individuals which are not part of the ontology subject matter.  Use skos:definition to define the meaning of category instances which are part of the subject matter."^^xsd:string ;
```

pwin:
Closes #1234.

scope note edited taking into account contributions from both Michael and Mark.

### DIFF
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -2353,9 +2353,8 @@ gist:description
 	skos:definition "A statement about someone or something's attributes or characteristics."^^xsd:string ;
 	skos:example "The Empire State Building is a 102-story Art Deco skyscraper in Midtown Manhattan in New York City, United States. It was designed by Shreve, Lamb & Harmon and built from 1930 to 1931."^^xsd:string ;
 	skos:prefLabel "description"^^xsd:string ;
-	skos:scopeNote """This property is to be used to describe data individuals which are not part of the ontology subject matter.  
-	A definition and a description have different semantics. Use skos:definition when it is 'a statement of the exact meaning of a thing', 
-	and gist:description when it is a representation or account of a thing, but not a statement of meaning."""^^xsd:string ;
+	skos:scopeNote "This property is used to describe instance data which is not part of the ontology. A definition and a description have different semantics. Use skos:definition for a statement of the meaning of a thing and gist:description to describe a thing's attributes, characteristics, or features."^^xsd:string ;
+	skos:example "A person does not have a definition, but might be described as being 6 feet tall with brown hair and blue eyes; an ontology class or taxonomy term has a definition."^^xsd:string ;
 	.
 
 gist:domainIncludes


ontologies/gistCore.ttl
sa-cheryld, line 1: New scope note at 2337-2339 is great.

sa-cheryld (commented):
I don't know why there were other things besides the changed scope note for gist:description. I didn't review those - I only reviewed the scope note change.

ontologies/gistCore.ttl
```diff
@@ -2332,7 +2334,9 @@ gist:description
 	skos:definition "A statement about someone or something's attributes or characteristics."^^xsd:string ;
 	skos:example "The Empire State Building is a 102-story Art Deco skyscraper in Midtown Manhattan in New York City, United States. It was designed by Shreve, Lamb & Harmon and built from 1930 to 1931."^^xsd:string ;
 	skos:prefLabel "description"^^xsd:string ;
-	skos:scopeNote "This property is used to provide a description of an individual in greater detail than a label."^^xsd:string ;
+	skos:scopeNote """This property is to be used to describe data individuals which are not part of the ontology subject matter.  
+	A definition and a description have different semantics.Use skos:definition when it is 'a statement of the exact meaning of a thing',
```
Jamie-SA, line 2338:
```suggestion
	A definition and a description have different semantics. Use skos:definition when it is 'a statement of the exact meaning of a thing', 
```

ontologies/gistCore.ttl
```diff
@@ -2332,7 +2334,9 @@ gist:description
 	skos:definition "A statement about someone or something's attributes or characteristics."^^xsd:string ;
 	skos:example "The Empire State Building is a 102-story Art Deco skyscraper in Midtown Manhattan in New York City, United States. It was designed by Shreve, Lamb & Harmon and built from 1930 to 1931."^^xsd:string ;
 	skos:prefLabel "description"^^xsd:string ;
-	skos:scopeNote "This property is used to provide a description of an individual in greater detail than a label."^^xsd:string ;
+	skos:scopeNote """This property is to be used to describe data individuals which are not part of the ontology subject matter.  
+	A definition and a description have different semantics. Use skos:definition when it is 'a statement of the exact meaning of a thing', 
+	and gist:description when it is a representation or account of a thing, but not a statement of meaning."""^^xsd:string ;
```
rjyounes, line 2339: I suggest the following:
```suggestion
	skos:scopeNote "This property is used to describe instance data which is not part of the ontology. A definition and a description have different semantics. Use skos:definition for a statement of the meaning of a thing and gist:description to describe a thing's attributes, characteristics, or features."^^xsd:string ;
	skos:example "A person does not have a definition, but might be described as being 6 feet tall with brown hair and blue eyes; an ontology class or taxonomy term has a definition."^^xsd:string ;
```

rjyounes (changes requested):
You will need a release note for this change as well. Please follow the directions in Contributing.md. This is a patch update.

PR #1262, opened 2025-08-25
Drop disjointness axiom between IntellectualProperty and Intention

dbsemantic:
This is the second time now I've contributed to gist and had the serializer rearrange the order of a lot of the file. I made sure I'm using the serializer in the proper directory, any idea why mine is different?


Jamie-SA:
I looked and didn't see any problem with the serialization of either your starting point or your end result. Maybe the serializer is not fully deterministic in restriction ordering.

Did you edit by hand or with Protege?

dbsemantic:
> Did you edit by hand or with Protege?
Protege

dbsemantic:
I didn't save a diff, but I did check it before committing. Protege's serialization was all over the place, there was hardly a line that wasn't changed in some way. I don't remember seeing any of the specific changes here though.

rjyounes:
We can't accept the PR like this as it's impossible to review. I have put it in draft state until we can get to the bottom of this.

rjyounes (approved):

PR #1271, opened 2025-09-19
Remove accepts property. Closes #1247

### LINKED ISSUE #1247: Fix text definition of gist:accepts or delete the property
The current description of gist:accepts is "The types of input messages that will be allowed."   

This is 
1) way to specific for the general use of the term 'accept' in English usage
2) it is not used anywhere else in gist (e.g. in relation to Messages), so there is no apparent need for it to be so restricted.

We should give it a better definition, or drop the property as too ambiguous.

stevenchalem:
Removed the property `gist:accepts`


Jamie-SA:
When using "Closes #1247" you should put that in the description. Putting it in the title does not link to the issue, nor will it actually close the issue upon merging.

PR #1303, opened 2025-10-22
remove range assertions from certain gist datatype properties

stevenchalem:
removed `rdfs:range xsd:string` from

- `gist:containedText`
- `gist:description`
- `gist:encryptedText`
- `gist:name`
- `gist:uniqueText`
- `gist:symbol`

### DIFF
--- a/docs/release_notes/issue1300-remove-range-assertions-from-certain-datatype-properties.md
+++ b/docs/release_notes/issue1300-remove-range-assertions-from-certain-datatype-properties.md
@@ -1,3 +1,3 @@
-### Minor
+### Major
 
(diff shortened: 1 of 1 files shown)


rjyounes (changes requested):
Just need a release note, everything else good

docs/release_notes/issue1300-remove-range-assertions-from-certain-datatype-properties.md
```diff
@@ -0,0 +1,3 @@
+### Minor
+
+- Removed `rdfs:range xsd:string` from some gist datatype properties. Issue [#1300](https://github.com/semanticarts/gist/issues/1300).
```
rjyounes: OK, but this is a major change as it affects inferencing. Probably in that case we should list them.

rjyounes (approved):

PR #1308, opened 2025-10-24
891 verify pre commit hook

### LINKED ISSUE #891: Refactor pre-commit hook to always use the latest version

Jamie-SA:
Closes #891 

Modified the pre-commit hook to verify that it matches the one in the tools directory. This will tell users to re-run setup.cmd if a new version of the pre-commit hook is released.

This is staked on the 898 branch so that they will not conflict with each other.

This should be tested on Windows to see if the "cmp" command exists. If not, I can change it to use "diff" if that is supported.

### DIFF
--- a/docs/release_notes/891-verify-pre-commit-hook
+++ b/docs/release_notes/891-verify-pre-commit-hook
@@ -1,0 +1,4 @@
+### Patch Updates
+
+- Added a test in the git pre-commit to verify that it is identical to the version in the ./tools directory.
+  If not, it instructs the user to run the setup.cmd command to update the git pre-commit hook.


Jamie-SA:
@rjyounes @stevenchalem 
This was not in the gist v14 project plan but I think it should go into 14 since we are already modifying the pre-commit hook and this change will fix the problem in the future of people not running setup.cmd when needed.

PR #1304, opened 2025-10-23
handle default prefixes

Jamie-SA:
Closes #898
This PR prevents commits if an ontology file has a default prefix declared.
Everyone will need to run setup.cmd again.

PR #1309, opened 2025-10-24
Refactor pre-commit hook

### LINKED ISSUE #891: Refactor pre-commit hook to always use the latest version

Jamie-SA:
Closes #891

This branch is stacked on the #1304 branch.

This moves all of the pre-commit logic into a separate file which is called by the new pre-commit hook.
This means that changes in the pre-commit-code file will automatically be used in the future as we make changes.

I think this is a better solution than the one I originally proposed, which checked if there was an update and told the user to re-run setup.cmd.


stevenchalem (approved):
Tested under Windows 11. 
This is a nice improvement!

rjyounes (approved):

rjyounes:
Note - this has been merged but not to `develop`.

PR #1331, opened 2025-10-30
Fix code block formatting

sa-bpelakh:
By default, `markdown2` doesn't support fenced code blocks, so an indented code block should be used instead to get proper formatting.

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -196,7 +196,6 @@ actions:
   source: "{input}/docs"
   target: "{output}/docs"
   includes:
-    - "Contributing.md"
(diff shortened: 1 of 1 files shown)


bundle.yaml
```diff
@@ -196,6 +196,7 @@ actions:
   source: "{input}/docs"
   target: "{output}/docs"
   includes:
+    - "Contributing.md"
```
rjyounes:
```suggestion

```

We don't include this one

rjyounes (changes requested):

PR #1334, opened 2025-10-31
Fix 'skos:example:' predicate

### LINKED ISSUE #1333: Release v14.0.0 has an non-parsable rdf/xml file
XML Parsing Error: not well-formed
Location: https://ontologies.semanticarts.com/ontology/gistCore
Line Number 2227, Column 16:
		<skos:example: rdf:datatype="&xsd;string">Meet an obligation; adhere to a specification or standard; meet terms and conditions of an offer; comply with a regulation.</skos:example:>
-----------------------------^

Jamie-SA:
Closes #1333 

Fixes typo in `skos:example:`.

It seems to be valid Turtle, but creates an invalid rdf/xml file, which is what is downloaded when you download the ontology IRI.

I hand edited the file on the web server so that it would be valid. But the version in the download zip file is still bad.

### DIFF
--- a/docs/release_notes/issue1333-fix-skos-example
+++ b/docs/release_notes/issue1333-fix-skos-example
@@ -1,4 +1,0 @@
-### Patch Updates
-
-- Fixed typo in `skos:example`.
-


stevenchalem (approved):

stevenchalem:
I created an issue to track this.

PR #1378, opened 2025-12-01
change text def of ProductSpecification and add release note

### LINKED ISSUE #1344: Fix definition of ProductSpecification
Current definition: "Offering something which could be physically warehoused or digitally stored."

* A specification is not necessarily an offering. 
* Definition should conform to our conventions - usually starting with "A..."

stevenchalem:
Closes #1344


rjyounes (approved):
Great, I like this definition.

PR #1388, opened 2025-12-15
Update definition of gist:Text

### LINKED ISSUE #1019: Make definition of Text more precise (was: Make Text a subclass of FormattedContent rather than Content)
I find the `Content` hierarchy confusing:
- ~~Should `Address` be a subclass of `Text`?~~
- Should `Text` be a subclass of `FormattedContent` or at least `ContentExpression`?
- ~~What is the difference between `Content` and `ContentExpression`?~~

willmandel:
Closes #1019

### DIFF
--- a/docs/release_notes/1019_new_text_definition.md
+++ b/docs/release_notes/1019_new_text_definition.md
@@ -1,0 +1,4 @@
+### Patch Updates
+
+- Changed definition of `gist:Text` to include the word "written". Issue [#1019](https://github.com/semanticarts/gist/issues/1019).
+
--- a/docs/release_notes/1019_new_text_definition.md
+++ b/docs/release_notes/1019_new_text_definition.md
@@ -1,4 +1,4 @@
 ### Patch Updates
 
-- Changed definition of `gist:Text` to include the word "written". Issue [#1019](https://github.com/semanticarts/gist/issues/1019).
+- Changed definition of `gist:Text` to specify that text is a written sequence of characters. Changed negative example to clarify that images of text are not text. Added scope note. Issue [#1019](https://github.com/semanticarts/gist/issues/1019).
 
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -1277,16 +1277,6 @@ gist:Offer
 			gist:ContingentObligation
 			[
 				a owl:Restriction ;
-				owl:onProperty gist:offersToProvide ;
-				owl:someValuesFrom owl:Thing ;
-			]
-			[
-				a owl:Restriction ;
-				owl:onProperty gist:offersToReceive ;
-				owl:someValuesFrom owl:Thing ;
-			]
-			[
-				a owl:Restriction ;
 				owl:onProperty gist:hasGiver ;
 				owl:someValuesFrom [
 					a owl:Class ;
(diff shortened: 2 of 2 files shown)


stevenchalem (changes requested):
This correctly addresses the issue as stated, but @uscholdm raised a good point: text can contain symbols that are not words or numbers. This includes puctionation. Maybe the definition should be something more like "Content expressed as a sequence of characters." Then we could give an example like "Natural language writing expressed in Unicode."

uscholdm:
> "Context expressed as a sequence of characters."

I think you meant: "Content expressed as a sequence of characters."

uscholdm:
There is a question of whether a photo of text is text, it is pixels not characters.

uscholdm (changes requested):
There is a question of whether a photo of text is text, it is pixels not characters.  Not sure what I think about this .  Needs a discussion.

willmandel:
> Maybe the definition should be something more like "Content expressed as a sequence of characters." Then we could give an example like "Natural language writing expressed in Unicode."

I agree with this, and have no issue changing to that definition. Might be nice to have a precise definition of "character" vs. "symbol", though I wonder if the definition would also work with the latter. I feel like "character" implies reference to an established meaning; for example, we know any font of the letter A is still an A, since that's what it refers to. Would a sequence of unique symbols - with no reference to a preexisting set - still be considered text?

> There is a question of whether a photo of text is text. 

My initial reaction is to say no, it's an image, which I think are distinct categories. That photo may contain a rendering of text, but the photo itself is not text.

rjyounes:
> There is a question of whether a photo of text is text, it is pixels not characters. Not sure what I think about this . Needs a discussion.

A photo is not text, by any definition, nor are pixels characters.

I think giving the unicode example will introduce some confusion, because text does not have to have an encoding.

Otherwise, I like @stevenchalem's definition except that we need the word "written," since that was the point of our decision.

pppelll:
@uscholdm pointed out that the definition is too narrow.
@stevenchalem suggested: "Content expressed as a sequence of characters."
@rjyounes pointed out that we need to include "written" per the decision reached in the meeting.

I'd suggest: "Content expressed as a written sequence of characters."
With a negative example of "Photographs or scans of text are not text. These instead depict a subject, which happens to be text."
And perhaps a scope note of: "Text may or may not have a particular encoding, such as unicode, but this is not a necessary condition"

Does this warrant a few minutes of revisiting in the next development meeting?

rjyounes:
> And perhaps a scope note of: "Text may or may not have a particular encoding, such as unicode, but this is not a necessary condition"

We should also mention that it has a language. So how about something like:

"Text is expressed in a language (human or computer) and may but need not specify an encoding."

I don't think this needs to go back to the team.

pppelll:
> We should also mention that it has a language. So how about something like:
> 
> "Text is expressed in a language (human or computer) and may but need not specify an encoding."

Good point. I feel good about that as a scope note.

> I don't think this needs to go back to the team.

Got it. @willmandel, let me know if you run into any issues with these, or updating a PR

uscholdm (approved):

stevenchalem (approved):

rjyounes (approved):

PR #1394, opened 2026-01-15
Updated definition and scope note of gist:System

### LINKED ISSUE #1288: Do Systems always have goals?
Textual definition of `gist:System`: "A composite with component parts that contribute to the goal of the system."

Can we add a restriction on `gist:hasGoal` some `owl:Thing`?

willmandel:
Closes #1288 

Removes reference to goals from the definition of `gist:System`. Adds scope note that clarifies that `gist:System` may be used for either man-made or natural systems.

### DIFF
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -1824,7 +1824,7 @@ gist:System
 			]
 		) ;
 	] ;
-	skos:definition "A composite made up of interacting or interdependent component parts that together function as a whole."^^xsd:string ;
+	skos:definition "A composite made up of interacting or interdependent components that together operate as a whole."^^xsd:string ;
 	skos:prefLabel "System"^^xsd:string ;
 	skos:scopeNote "May be used to refer to either man-made or natural systems."^^xsd:string ;
 	.
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -1825,6 +1825,7 @@ gist:System
 		) ;
 	] ;
 	skos:definition "A composite made up of interacting or interdependent components that together operate as a whole."^^xsd:string ;
+	skos:example "A manufacturing system, a storm system."^^xsd:string ;
 	skos:prefLabel "System"^^xsd:string ;
 	skos:scopeNote "May be used to refer to either man-made or natural systems."^^xsd:string ;
 	.
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -1277,16 +1277,6 @@ gist:Offer
 			gist:ContingentObligation
 			[
 				a owl:Restriction ;
-				owl:onProperty gist:offersToProvide ;
-				owl:someValuesFrom owl:Thing ;
-			]
-			[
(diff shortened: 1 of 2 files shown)


stevenchalem (approved):

ontologies/gistCore.ttl
```diff
@@ -1824,8 +1824,9 @@ gist:System
 			]
 		) ;
 	] ;
-	skos:definition "A composite with component parts that contribute to the goal of the system."^^xsd:string ;
+	skos:definition "A composite made up of component parts."^^xsd:string ;
```
dylan-sa: IMO this is too broad. It doesn't distinguish systems from collections, for example. With systems, there's something special about the way the parts interact with one another--to achieve a goal (in the man-made case) or give rise to something (e.g., a storm system giving rise to lightning). See also the scopeNote on `Composite`: "Some composites, like systems, networks, and ordered collections, have internal organization, while others, such as ordered collections, typically do not."

What makes systems different from networks or collections?

dylan-sa (changes requested):

ontologies/gistCore.ttl
willmandel: I agree that there is something to be said about the function/purpose/arrangement of a system.

One potential approach:
A composite made up of component parts that may combine to give rise to a particular action. 

"Action" may not be the best word here (tricky to align with something like a river system). Maybe "situation" would fit? The "may" qualifier is an attempt to include inert/inactive systems, though I'm not 100% certain that's necessary.

willmandel: Update following conversation with @dylan-sa:

"A composite made up of component parts that function as a whole."

This helps clear up any fuzziness around distinguishing between actions/situations/phenomena, and IMO covers the inert/inactive system case well enough. @stevenchalem, what do you think of this definition?

stevenchalem: I like it. I would maybe add something about how the component parts relate to each other. Maybe
"A composite made up of interacting or interdependent component parts that together function as a whole."

ontologies/gistCore.ttl
```diff
@@ -1824,8 +1824,9 @@ gist:System
 			]
 		) ;
 	] ;
-	skos:definition "A composite with component parts that contribute to the goal of the system."^^xsd:string ;
+	skos:definition "A composite made up of interacting or interdependent component parts that together function as a whole."^^xsd:string ;
```
```suggestion
	skos:definition "A composite made up of interacting or interdependent components that together operate as a whole."^^xsd:string ;
```

ontologies/gistCore.ttl
```diff
@@ -1824,8 +1824,9 @@ gist:System
...
 	skos:prefLabel "System"^^xsd:string ;
+	skos:scopeNote "May be used to refer to either man-made or natural systems."^^xsd:string ;
```
rjyounes, line 1830: Maybe add an example of each.

rjyounes (changes requested):

dylan-sa (approved):

rjyounes (approved):

PR #1400, opened 2026-02-23
Revise aspect definition

### LINKED ISSUE #1393: Clarify definition/examples for Aspect
I find the definition/examples for ```Aspect``` to be slightly misleading. If there are no limits on the specificity of an aspect (say, ```VolumeOfOil``` instead of ```Volume```), that should be clarified in the definition. If, instead, we favor users making more general, reusable aspects, I think that would be worth adding to the scope note.

willmandel:
Closes #1393.

### DIFF
--- a/docs/release_notes/1393-revise-aspect-definition
+++ b/docs/release_notes/1393-revise-aspect-definition
@@ -1,3 +1,3 @@
-### Patch Updates
+### Minor Updates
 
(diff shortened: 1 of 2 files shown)


stevenchalem (approved):

rjyounes (approved):
Very nicely done! The notes are well-written.

PR #1401, opened 2026-02-23
Update quotes within quotes

### LINKED ISSUE #1246: Update quotes within quotes
In some gist annotations we have quotes within quotes. For example:
```
skos:scopeNote 'This is not the inverse of gist:goesToAgent. A message can be to someone. If we made it the inverse the person would be "from" the message'^^xsd:string ;
```
In cases like this, I believe it's syntactically OK to use single quotes (') for the outer quotes and double quotes (") for the inner ones. However, I have noticed occasionally that some tools can only handle double quotes on the outside and single on the inside. For this reason, I suggest we tweak the above annotation (and a couple others) accordingly.

This is not a pressing fix, but it could occasionally make life easier.

dylan-sa:
Closes #1246.

Turns out there was only one remaining string to be changed.


rjyounes (changes requested):
Dylan, I'm thinking we should add this to the style guide for posterity. Can you please include that in the PR?

dylan-sa:
@rjyounes Yes, good point. Ready for another look

rjyounes (approved):

PR #1403, opened 2026-02-26
Delete/Deprecate gist:Building

### LINKED ISSUE #1205: Delete gist:Building
Doesn't seem like a core concept.

brett-heffelfinger:
## Major Updates

- Deprecated classes `gist:Landmark` and `gist:Building`. Issue [#1205](https://github.com/semanticarts/gist/issues/1205).

## Minor Updates

- Changed examples of `gist:PhysicalIdentifiableItem` to include a landmark and building. Issue [#1205](https://github.com/semanticarts/gist/issues/1205).
- Modified existing examples of `gist:PhysicalIdentifiableItem` to be more specific. Issue [#1205](https://github.com/semanticarts/gist/issues/1205).

### DIFF
--- a/docs/release_notes/1205-delete-gist-building
+++ b/docs/release_notes/1205-delete-gist-building
@@ -1,0 +1,8 @@
+## Major Updates
+
+- Deprecated classes `gist:Landmark` and `gist:Building`. Issue [#1205](https://github.com/semanticarts/gist/issues/1205).
+
+## Minor Updates
+
+- Changed examples of `gist:PhysicalIdentifiableItem` to include a landmark and building.
+- Modified existing examples of `gist:PhysicalIdentifiableItem` to be more specific.
--- a/docs/release_notes/1205-delete-gist-building
+++ b/docs/release_notes/1205-delete-gist-building
@@ -4,5 +4,4 @@
 
 ## Minor Updates
 
-- Changed examples of `gist:PhysicalIdentifiableItem` to include a landmark and building.
-- Modified existing examples of `gist:PhysicalIdentifiableItem` to be more specific.
+- Updated examples of `gist:PhysicalIdentifiableItem`.
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -239,6 +239,7 @@ gist:Behavior
 gist:Building
 	a owl:Class ;
 	rdfs:subClassOf gist:Landmark ;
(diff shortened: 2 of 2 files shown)


docs/release_notes/1205-delete-gist-building
```diff
@@ -0,0 +1,8 @@
+## Major Updates
+
+- Deprecated classes `gist:Landmark` and `gist:Building`. Issue [#1205](https://github.com/semanticarts/gist/issues/1205).
+
+## Minor Updates
+
+- Changed examples of `gist:PhysicalIdentifiableItem` to include a landmark and building.
+- Modified existing examples of `gist:PhysicalIdentifiableItem` to be more specific.
```
rjyounes:
```suggestion
- Updated examples of `gist:PhysicalIdentifiableItem`.
```

ontologies/gistCore.ttl
```diff
@@ -1563,7 +1565,7 @@ gist:PhysicalIdentifiableItem
 		gist:UnitOfMeasure
 		;
 	skos:definition "A discrete physical object which, if subdivided, will result in parts that are distinguishable in nature from the whole and in general also from the other parts."^^xsd:string ;
-	skos:example "A computer, a book, a car."^^xsd:string ;
+	skos:example "A laptop, a physical book, a car, a building, a landmark (such as a tree stump with an etched marking to denote a campsite)."^^xsd:string ;
```
rjyounes, line 1568: I don't love the repeated indefinite articles (I know they didn't originate with you). What do you think of just "A computer, physical book, car..."?

brett-heffelfinger: There's something odd about it to me. I'm not quite sure what, but here they are side-by-side:

> A laptop, a physical book, a car, a building, a landmark (such as a tree stump with an etched marking to denote a campsite).

> A laptop, physical book, car, building, landmark (such as a tree stump with an etched marking to denote a campsite).

Maybe it's that the second makes me think there should explicitly be an "or" before landmark, and adding that implies the examples are a finite set. The example of the example (in parentheses) is also throwing me for a loop, but I thought something like that was necessary to indicate what a landmark could be.

rjyounes: OK, I'm OK with it as is. There's something odd about it to me either way.

rjyounes (approved):

PR #1405, opened 2026-03-03
Update definition and scope note for gist:Specification

### LINKED ISSUE #1404: Update definition of gist:Specification
Potential `gist:Specification` definition improvement:
"**One or more** characteristics that specify what it means..."
change to
"**The set of** characteristics that specify what it means...".

irinafilitovich:
Closes #1404

### DIFF
--- a/docs/release_notes/1404-update-definition-of-gistspecification.md
+++ b/docs/release_notes/1404-update-definition-of-gistspecification.md
@@ -1,0 +1,3 @@
+### Minor Updates
+
+- Updated definition, example, scope note of `gist:Specification` to explicitly specify common ways of using the class and to disabmiguate with `gist:KnowledgeConcept`. Issue [#1404](https://github.com/semanticarts/gist/issues/1404).
--- a/docs/release_notes/1404-update-definition-of-gistspecification.md
+++ b/docs/release_notes/1404-update-definition-of-gistspecification.md
@@ -1,3 +1,3 @@
 ### Minor Updates
 
-- Updated definition, example, scope note of `gist:Specification` to explicitly specify common ways of using the class and to disabmiguate with `gist:KnowledgeConcept`. Issue [#1404](https://github.com/semanticarts/gist/issues/1404).
+- Updated definition, example, scope note of `gist:Specification` to explicitly specify common ways of using the class and to distinguish from `gist:KnowledgeConcept`. Issue [#1404](https://github.com/semanticarts/gist/issues/1404).
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -1775,10 +1775,10 @@ gist:ServiceSpecification
 gist:Specification
 	a owl:Class ;
 	rdfs:subClassOf gist:Intention ;
-	skos:definition "The set of characteristics that specify what it means to be a particular type of thing, such as a material, product, service or event. A specification is sufficiently precise to allow evaluating conformance to the specification."^^xsd:string ;
-	skos:example "The specification of the iPhone 14; hypothetical events covered by a homeowner's insurance policy."^^xsd:string ;
+	skos:definition "The set of characteristics and constraints on their values that represents what expectations a particular type of thing, such as a material, product, service or event, must satisfy. A specification is sufficiently precise to allow evaluating conformance to the specification."^^xsd:string ;
+	skos:example "The specification of the iPhone 14; hypothetical events covered by a homeowner's insurance policy; product quality specification that a specific can of peas at the supermarket may conform to to be allowed on the shelf."^^xsd:string ;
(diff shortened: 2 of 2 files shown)


ontologies/gistCore.ttl
```diff
@@ -1726,10 +1726,10 @@ gist:ServiceSpecification
 gist:Specification
 	a owl:Class ;
 	rdfs:subClassOf gist:Intention ;
-	skos:definition "One or more characteristics that specify what it means to be a particular type of thing, such as a material, product, service or event. A specification is sufficiently precise to allow evaluating conformance to the specification."^^xsd:string ;
-	skos:example "The specification of the iPhone 14; hypothetical events covered by a homeowner's insurance policy."^^xsd:string ;
+	skos:definition "The set of characteristics and constraints on their values that represents what expectations a particular type of thing, such as a material, product, service or event, must satisfy to conform. A specification is sufficiently precise to allow evaluating conformance to the specification."^^xsd:string ;
+	skos:example "The specification of the iPhone 14; hypothetical events covered by a homeowner's insurance policy; product quality specification that a specific can of peas at the supermarket may conform to to be allowed on the shelf."^^xsd:string ;
```
rjyounes: The peas example is a bit weird. How about one from the pharma domain?

philblackwood:
Looks good.

philblackwood (approved):

docs/release_notes/1404-update-definition-of-gistspecification.md
```diff
@@ -0,0 +1,3 @@
+### Minor Updates
+
+- Updated definition, example, scope note of `gist:Specification` to explicitly specify common ways of using the class and to disabmiguate with `gist:KnowledgeConcept`. Issue [#1404](https://github.com/semanticarts/gist/issues/1404).
```
rjyounes:
```suggestion
- Updated definition, example, scope note of `gist:Specification` to explicitly specify common ways of using the class and to distinguish from `gist:KnowledgeConcept`. Issue [#1404](https://github.com/semanticarts/gist/issues/1404).
```

rjyounes (changes requested):

ontologies/gistCore.ttl
irinafilitovich: Example replaced with a pharma one.

ontologies/gistCore.ttl
```diff
@@ -1727,10 +1726,10 @@ gist:ServiceSpecification
 gist:Specification
 	a owl:Class ;
 	rdfs:subClassOf gist:Intention ;
-	skos:definition "One or more characteristics that specify what it means to be a particular type of thing, such as a material, product, service or event. A specification is sufficiently precise to allow evaluating conformance to the specification."^^xsd:string ;
-	skos:example "The specification of the iPhone 14; hypothetical events covered by a homeowner's insurance policy."^^xsd:string ;
+	skos:definition "The set of characteristics and constraints on their values that represents what expectations a particular type of thing, such as a material, product, service or event, must satisfy to conform. A specification is sufficiently precise to allow evaluating conformance to the specification."^^xsd:string ;
```
```suggestion
	skos:definition "The set of characteristics that specify what it means to be a particular type of thing, such as a material, product, service or event, must satisfy to conform. A specification is sufficiently precise to allow evaluating conformance to the specification."^^xsd:string ;
```

I'd like to go back to the change as you originally made it, for two reasons:

1. "Characteristic" can mean either a plain aspect or an aspect + value; e.g., height or height of 5 ft. So "constraints on their values" is redundant.
2. The definition as it stands is circular. "...satisfy to conform." Conform to what? Conform to the specification. So the circularity is here: "The set of characteristics...that represents what...a particular type of thing...must satisfy to conform to [the set of characteristics...that represents what...a particular type of thing...must satisfy to conform to...etc].

I can live with keeping the constraints, since it's redundant but not wrong, so then we have

	skos:definition "The set of characteristics and constraints on their values that specify what it means to be a particular type of thing, such as a material, product, service or event, must satisfy to conform. A specification is sufficiently precise to allow evaluating conformance to the specification."^^xsd:string ;

rjyounes (changes requested):

ontologies/gistCore.ttl
irinafilitovich: Thank you, @rjyounes, I'd like to keep the word "constraints" to support the pharma example I added, there constraints are expressed as acceptance criteria, e.g. the value for some aspect must be between these 2 values. I landed on the definition below, compared to what you posted, I also removed "must satisfy to conform", which I think you meant to remove. Let me know if more changes are necessary.
`skos:definition "The set of characteristics and constraints on their values that specify what it means to be a particular type of thing, such as a material, product, service or event. A specification is sufficiently precise to allow evaluating conformance to the specification."^^xsd:string ;
`

rjyounes (approved):

PR #1408, opened 2026-03-11
containedText and uniqueText Proposals

### LINKED ISSUE #1043: gist:containedText and gist:uniqueText proposals
Proposals:
* Make `gist:uniqueText` a subproperty of `gist:containedText`
* Add `gist:Content` and `gist:ContentExpression` to the `domainIncludes` assertion. 
  * Or probably better: replace `gist:Text` and `gist:ID` with `gist:Content`: I'm not sure if it's useful to include subproperties of a property that's the object of `domainIncludes`. It is redundant, but perhaps there's a value to being explicit.
* Add a scope note indicating that a content object may have multiple `containedText` assertions, and that whether to do so or not is an implementation-specific choice.
* Definition is poor. Current:

```
skos:definition "A string that is closely associated with an individual."
```
This sounds like any resource could have contained text. E.g., do I contain my name? Obviously not. I propose

```
skos:definition "Text contained by a content object or tag."
```
Actually, I don't like this because it's too narrow, but I can't think of a better option at the moment. Suggestions are welcome.

stevenchalem:
closes #1043

### DIFF
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -2151,10 +2151,13 @@ gist:conformsTo
 
 gist:containedText
 	a owl:DatatypeProperty ;
-	skos:definition "A string that is closely associated with an individual."^^xsd:string ;
+	skos:definition "Relates an entity containing textual information to the literal character sequence constituting that textual information."^^xsd:string ;
 	skos:example "The string associated with a tag or with text content."^^xsd:string ;
 	skos:prefLabel "contained text"^^xsd:string ;
+	skos:scopeNote "An entity may have multiple `gist:containedText` assertions. Whether or not to use multiple assertions is an implementation decision."^^xsd:string ;
 	gist:domainIncludes
+		gist:Content ,
+		gist:ContentExpression ,
 		gist:Tag ,
 		gist:Text
 		;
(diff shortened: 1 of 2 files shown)


ontologies/gistCore.ttl
```diff
@@ -2151,10 +2151,13 @@ gist:conformsTo
 
 gist:containedText
 	a owl:DatatypeProperty ;
-	skos:definition "A string that is closely associated with an individual."^^xsd:string ;
+	skos:definition "Relates an entity containing textual information to the literal character sequence constituting that textual information."^^xsd:string ;
```
rjyounes: Why "literal character sequence" rather than "string"? Seems unnecessarily cumbersome and complex.

ontologies/gistCore.ttl
```diff
@@ -2151,10 +2151,13 @@ gist:conformsTo
...
 	skos:example "The string associated with a tag or with text content."^^xsd:string ;
 	skos:prefLabel "contained text"^^xsd:string ;
+	skos:scopeNote "An entity may have multiple `gist:containedText` assertions. Whether or not to use multiple assertions is an implementation decision."^^xsd:string ;
```
```suggestion
	skos:scopeNote "If there are multiple strings associated with an entity, it is an implementation decision whether to combine them into a single contained text value or to use multiple contained text assertions."^^xsd:string ;
```

dylan-sa (approved):

rjyounes (approved):

PR #1413, opened 2026-03-30
edit annotations on FormattedContent, RenderedContent

### LINKED ISSUE #1025: Clarify distinction between RenderedContent and FormattedContent
Clarify via scopeNotes and examples. I am not clear when you would use one rather than the other.

stevenchalem:
Closes #1025

### DIFF
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -706,7 +706,7 @@ gist:FormattedContent
 	skos:definition "Content encoded in a specific format, existing as data independent of any particular physical medium."^^xsd:string ;
 	skos:example "A PDF file stored on a server; an HTML document in memory; a JPEG file on disk."^^xsd:string ;
 	skos:prefLabel "Formatted Content"^^xsd:string ;
-	skos:scopeNote "Different instances of the same content in the same format are considered distinct."^^xsd:string ;
+	skos:scopeNote "Two copies of the same PDF file on different servers are separate instances of formatted content.z"^^xsd:string ;
 	.
 
 gist:Function
@@ -1633,7 +1633,7 @@ gist:RenderedContent
 			]
 		) ;
 	] ;
-	skos:definition "A specific expression of content in some physical medium."^^xsd:string ;
+	skos:definition "Content expressed in some physical medium."^^xsd:string ;
 	skos:example "Words printed on paper; audio played on speakers; an image displayed on a monitor; an original painting."^^xsd:string ;
 	skos:prefLabel "Rendered Content"^^xsd:string ;
 	skos:scopeNote "Renderings of the same file on two different monitors are separate instances of rendered content."^^xsd:string ;
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -703,7 +703,7 @@ gist:FormattedContent
 			]
 		) ;
 	] ;
-	skos:definition "Content encoded in a specific format, existing as data independent of any particular physical medium."^^xsd:string ;
+	skos:definition "Content encoded in a specific format, but existing as data independent of any particular physical medium."^^xsd:string ;
 	skos:example "A PDF file stored on a server; an HTML document in memory; a JPEG file on disk."^^xsd:string ;
 	skos:prefLabel "Formatted Content"^^xsd:string ;
 	skos:scopeNote "Two copies of the same PDF file on different servers are separate instances of formatted content.z"^^xsd:string ;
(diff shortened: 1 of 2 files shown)


ontologies/gistCore.ttl
```diff
@@ -1632,9 +1633,10 @@ gist:RenderedContent
 			]
 		) ;
 	] ;
-	skos:definition "Content which has been physically expressed."^^xsd:string ;
-	skos:example "Words printed on paper; audio played on speakers; images displayed on monitors."^^xsd:string ;
+	skos:definition "Content expressed in some physical medium."^^xsd:string ;
```
DanCarey404: Perhaps "...expressed via some physical medium." ?

ontologies/gistCore.ttl
```diff
@@ -703,9 +703,10 @@ gist:FormattedContent
 			]
 		) ;
 	] ;
-	skos:definition "Content which is in a particular format."^^xsd:string ;
-	skos:example "An HTML website, a PDF document, a JPG image."^^xsd:string ;
+	skos:definition "Content encoded in a specific format, existing as data independent of any particular physical medium."^^xsd:string ;
+	skos:example "A PDF file stored on a server; an HTML document in memory; a JPEG file on disk."^^xsd:string ;
 	skos:prefLabel "Formatted Content"^^xsd:string ;
+	skos:scopeNote "Two copies of the same PDF file on different servers are separate instances of formatted content.z"^^xsd:string ;
```
DanCarey404: This note leaves me wondering, would the separate instances have different IRIs?  I think they would not.  So, I am unsure if this note is correct or helpful.
I think a better example would be a single piece of content formatted as PDF and HTML.  Each formatted type would have a separate IRI. (But the single PDF IRI could have multiple instantiations on different servers, stores, etc.).
Does that ring true with you?

ontologies/gistCore.ttl
```diff
@@ -703,9 +703,10 @@ gist:FormattedContent
 			]
 		) ;
 	] ;
-	skos:definition "Content which is in a particular format."^^xsd:string ;
-	skos:example "An HTML website, a PDF document, a JPG image."^^xsd:string ;
+	skos:definition "Content encoded in a specific format, existing as data independent of any particular physical medium."^^xsd:string ;
```
DanCarey404: Would a conjunction help? "...format, but existing..."

ontologies/gistCore.ttl
```diff
@@ -1632,9 +1633,10 @@ gist:RenderedContent
 			]
 		) ;
 	] ;
-	skos:definition "Content which has been physically expressed."^^xsd:string ;
-	skos:example "Words printed on paper; audio played on speakers; images displayed on monitors."^^xsd:string ;
+	skos:definition "Content expressed in some physical medium."^^xsd:string ;
+	skos:example "Words printed on paper; audio played on speakers; an image displayed on a monitor; an original painting."^^xsd:string ;
 	skos:prefLabel "Rendered Content"^^xsd:string ;
+	skos:scopeNote "Renderings of the same file on two different monitors are separate instances of rendered content."^^xsd:string ;
```
DanCarey404, line 1639: I am left with a similar question as earlier.  Are you saying that each instance of a music file being played on a speaker has a different IRI?

DanCarey404 (changes requested):

ontologies/gistCore.ttl
stevenchalem: sure. change made.

ontologies/gistCore.ttl
stevenchalem: I like it.

ontologies/gistCore.ttl
stevenchalem: Yes, I am saying that. The decision on this issue was to follow Rebecca's note at https://github.com/semanticarts/gist/issues/1025#issuecomment-1906636441 which states "the same file rendered on my screen vs yours are two different instances of rendered content."

ontologies/gistCore.ttl
stevenchalem: I agree. Is this better?

ontologies/gistCore.ttl
DanCarey404: Okaaay.  As  practical matter, that seems unimplementable to me.  But with luck, I will never confront that issue.

DanCarey404 (approved):

rjyounes (approved):

PR #1414, opened 2026-03-31
restore executable flag in pre-commit hook

### LINKED ISSUE #1345: Track executable bit on all scripts
Comment from @Jamie-SA on PR #1313:

 I do think we should be keeping track of the executable bit on our scripts. I vaguely remember it being added to fix some problem. I would think contributions from users of Windows would be the biggest problem given that Windows does not have the concept of an executable bit.

We could enforce the setting within the github actions by always checking and adding it back in a automated commit. This would be easy to add, however it could also cause issues because for Windows users the commit they just pushed would no longer be the latest commit on the branch.

Maybe we put an warning in the script files to not edit & commit them on Windows?

stevenchalem:
Closes #1345


.gitattributes
```diff
@@ -0,0 +1,6 @@
+# Mark shell scripts as executable.
+# The pre-commit hook restores the executable bit in Git's index
+# for these files, even when edited on Windows.
+*.sh executable
+tools/pre-commit-code executable
+tools/serializer/pre-commit executable
```
Jamie-SA, line 8: Updates, partly in anticipation of #1304 

```suggestion
tools/pre-commit-code executable
tools/pre-commit-hook executable
tools/remove_default_prefix/remove_default_prefix.py
tools/serializer/pre-commit executable
tools/setup.cmd
```

Jamie-SA (changes requested):

.gitattributes
```diff
@@ -0,0 +1,9 @@
+# Mark shell scripts as executable.
+# The pre-commit hook restores the executable bit in Git's index
+# for these files, even when edited on Windows.
+*.sh executable
+tools/pre-commit-code executable
+tools/pre-commit-hook executable
+tools/remove_default_prefix/remove_default_prefix.py executable
```
stevenchalem, line 7: @Jamie-SA in your suggested change the "executable" keyword was missing from this line. I added it.

Jamie-SA (approved):

.gitattributes
```diff
@@ -0,0 +1,9 @@
...
+tools/serializer/pre-commit executable
+tools/setup.cmd
```
Jamie-SA: You caught this, but it seems to have not been updated.
```suggestion
tools/setup.cmd executable
```

Jamie-SA (changes requested):

Jamie-SA (approved):

sa-bpelakh (approved):

PR #1428, opened 2026-04-10
Revise annotations for Content and ContentExpression

### LINKED ISSUE #1312: Add clarifying examples to Content and ContentExpression
Current (gist 14) annotations on the two classes:

```
gist:Content
   skos:definition "Information available in some medium."^^xsd:string ;
   skos:example "A document, program, image."^^xsd:string ;


gist:ContentExpression
	skos:definition "Intellectual property reduced to text, audio, etc.  If it contains text (written or spoken), it will be expressed in a language."^^xsd:string ;
```

`Content` may be at the level of IP but must be expressed in a medium in order to be content - I.e, it cannot exist _solely_ in the abstract. `ContentExpression` is a particular expression (in roughly the FRBR sense). Note that the equivalent of FRBR Work is not `Content` but rather `IntellectualProperty`, since `Content` encompasses both the abstract content _and_ its expression.

Proposed improvements:
* Change "Intellectual property" in definition of `ContentExpression` to `Content`.
* Clarify definitions and examples. 

Specifically:

```
gist:Content
	skos:definition "Information available in some medium."^^xsd:string ;
	skos:prefLabel "Content"^^xsd:string ;
	skos:scopeNote "Categories are not content until they are written down."^^xsd:string , 
            "This class includes both abstract content and its expression, but it must have at least one expression in order to be content. For example, the idea a writer has for a book is not content until it is expressed in some form, but a single content object may have multiple expressions - e.g., the particular print and audio editions."^^xsd:string ;
	.

gist:ContentExpression
	skos:definition "Content reduced to text, audio, etc."^^xsd:string ;
        skos:example "A specific print or audio edition of 'The Adventures of Huckleberry Finn," such as the one identified by  ISBN-13 978-1953649805.^^xsd:string ;
        skos:scopeNote "If it contains text (written or spoken), it will be expressed in a language."^^xsd:string ,
            "While the content expression is less abstract than a 'Work,' it is more abstract than the specific physical items that render the expression. For example, any physical or electronic book with ISBN-13 978-1953649805 is associated with the same content expression."^^xsd:string ;
	skos:prefLabel "Content Expression"^^xsd:string ;
	.
```

stevenchalem:
Closes #1312

### DIFF
--- a/docs/release_notes/1312-revise-annoations-for-content-and-contentexpression.md
+++ b/docs/release_notes/1312-revise-annoations-for-content-and-contentexpression.md
@@ -1,0 +1,3 @@
+### Patch Updates
+
+- Revised annotations for `gist:Content` and `gist:ContentExpression`. Issue [#1312](https://github.com/semanticarts/gist/issues/1312).
--- a/ontologies/gistCore.ttl
+++ b/ontologies/gistCore.ttl
@@ -408,9 +408,12 @@ gist:Content
 		gist:UnitOfMeasure
 		;
 	skos:definition "Information available in some medium."^^xsd:string ;
-	skos:example "A document, program, image."^^xsd:string ;
+	skos:example "The literary work 'The Adventures of Huckleberry Finn' by Mark Twain, independent of any particular edition."^^xsd:string ;
 	skos:prefLabel "Content"^^xsd:string ;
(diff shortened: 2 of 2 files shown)


rjyounes (changes requested):
Looks good, release note needed.

rjyounes (approved):

PR #1446, opened 2026-04-17
Stevenchalem/doc links

stevenchalem:
Fix some links in md files and correct some errors in bundle.yaml. 
Co-authored by Claude Code.

### DIFF
--- a/bundle.yaml
+++ b/bundle.yaml
@@ -228,6 +228,15 @@ actions:
     from: "\\./gist-logo.png"
     to: "../gist-logo.png"
 - action: "move"
+  message: "Replace './migration/' with '../migration/' in README."
+  source: "{output}/docs/"
+  target: "{output}/docs/"
+  includes:
+    - "README.md"
+  replace:
+    from: "\\./migration/"
+    to: "../../migration/"
+- action: "move"
   message: "Replace '../docs/' with './' in ReleaseNotes."
   source: "{output}/docs/"
   target: "{output}/docs/"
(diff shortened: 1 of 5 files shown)


docs/ReleaseNotes.md
```diff
@@ -320,8 +320,8 @@ Issue [#1060](https://github.com/semanticarts/gist/issues/1060).
 
 ### Documentation Updates
 
-- Added section to [gist Style Guide](./gistStyleGuide.html) on use of `gist:nonConformingLabel` annotation.
-- Updated [gist Style Guide](./gistStyleGuide.html) to define best practices for:
+- Added section to [gist Style Guide](https://github.com/semanticarts/gist/blob/v13.0.0/docs/gistStyleGuide.md) on use of `gist:nonConformingLabel` annotation.
```
sa-bpelakh, line 323: The reason we were using relative paths was that each version of the docs was bundled together as part of the release, so accessing the release notes of a specific version would resolve the links to supporting documents of the same version.

We convert the `.md` links in the documents to `.html` links during generation, but only for local links, so an explicit version of the `.md` file would not be remapped and create a broken link, right?

bundle.yaml
```diff
@@ -330,6 +348,15 @@ actions:
   target: "{output}/migration/v12.0/"
   includes:
     - "*.md"
+- action: "move"
```
sa-bpelakh, line 351: It seems unreasonable and unsustainable to create a separate section for each version. Could we possibly use `source: "{output}/migration/"` and then use `- v*/*.html` in the `includes`, I think it would transform all the sub-directories. It's worth testing.

sa-bpelakh (commented):
Had a couple of comments - need to see if we need to improve the tooling to support a better way forward.

PR #1450, opened 2026-04-21
Reconcile histories of main and develop

stevenchalem:
This PR reconciles divergent history between `main` and `develop`. The two branches have identical file trees — only the commit histories differs. (`main` has the canonical `release/14.1.0` merge; `develop` has parallel-applied commits with different SHAs).

It will appear in this PR that 33 files are being changed; this is because GitHub's is looking only at the commits in the source branch.

Prior to merging you can confirm that the file trees are identical by running:
```
git fetch origin
git diff origin/main..origin/develop --stat
``` 
or by previewing the merge first without committing it:
```
git fetch origin
git checkout develop
git merge --no-commit --no-ff origin/main
git diff --cached --stat   # should show nothing
git merge --abort
```
In either case, the empty output from the diff demonstrates that there are no file-level differences.


rjyounes:
Closing - I'm handling the divergent history in a different way.

PR #1508, opened 2026-08-27
class equivalence for electronic address (and ripple effects)

### LINKED ISSUE #1478: Formal definitions for Address and ElectronicAddress that allows inferring an instance into each class
May not be possible in all cases. See parent issue for more discussion.

philblackwood:
Closes #1478 

Created class equivalence for ElectronicAddress and to support it:

Added classes for ElectronicEndpoint and TransmittedContent.

Modified Address, PhysicalAddress, comesFromPlace, goesToPlace, and Message.

This change interacts with the name change from GeoLocation to GeoSpatialObject (i.e. left GeoLocation in some places in this PR that will not be changed by PR 1505).


Jamie-SA:
@philblackwood please put which issue this PR is meant to address/close/fix in the PR description.

ontologies/gistCore.ttl
```diff
@@ -1154,6 +1196,7 @@ gist:Message
 				owl:someValuesFrom [
 					a owl:Class ;
 					owl:unionOf (
+						gist:ElectronicEndpoint
```
Jamie-SA, line 1199: I am wondering if there is a mismatch between `gist:comesFromAgent` and `gist:comesFromPlace`?

You have added `gist:ElectronicEndpoint` to the range unions of `gist:comesFromPlace` and `gist:goesToPlace` which seems reasonable. But here you are adding it to the restriction on `gist:comesFromAgent` and `gist:goesToAgent` and it is not in the range of those properties.

dylan-sa:
I noticed today that if we use the distinctionary pattern for `ElectronicAddress` and `PhysicalAddress`, we can drop the `subClassOf` assertions, since those will be added by the materialize_subclass_inferences script. Ditto for the `TransmittedContent` def.

PR #1509, opened 2026-08-28
add class equivalence to contract term

### LINKED ISSUE #1354: Add restriction to ContractTerm (was Should ContractTerm be a subclass of Specification?)
@philblackwood I think this is what you wanted to discuss.

philblackwood:
Closes #1354

Simplified definition to "A specification that is part of a contract."

Added class equivalence: Specification and (isPartOf some Contract)


ontologies/gistCore.ttl
```diff
@@ -1,3 +1,4 @@
+@prefix : <https://w3id.org/semanticarts/ontology/gistCore#> .
```
Jamie-SA, line 1: @philblackwood do you have the new pre-commit hook installed (run ./tools/setup.cmd)? This default prefix should have been removed by our hook.
```suggestion
```

Jamie-SA (changes requested):

